PR #57, opened 2015-01-04
read_utils.py purge_unmated: add an argument to customize the regex used...

mlin:
add an argument to customize the regex used to parse paired read IDs in the FASTQ files. Helpful for files generated by SRA fastq-dump, which emits read IDs formatted differently from the Illumina base caller.


dpark01:
Awesome, thanks. One thing that I'll note is that process/pipeline-wise, I've been moving towards the new commands in read_utils/taxon_filter that operate on BAM files, so I don't know how different the SRA dump to BAM looks from the FASTQ output. But this purge_unmated step (which follows trimmomatic and all that) still operates on FASTQ files, because it's one of the last steps before Trinity. Anyway, this is helpful to know that with these tweaks, we're able to operate on SRA dumps directly.

PR #61, opened 2015-01-11
Dp trinity

iljungr:
Trinity now builds on linux (but not Mac), and test_tools succeeds on travis. I ran trinity's own test_Trinity_Assembly on my broad account and all tests passed. 

I have NOT added a trinity test to our test suite. Don't know if you have something in mind.

I have not looked over your changes to assembly.py or assembly.rules; I don't know if you are ready to submit them.

### DIFF
--- a/assembly.py
+++ b/assembly.py
@@ -12,6 +12,7 @@ import Bio.AlignIO, Bio.SeqIO, Bio.Data.IUPACData
 import util.cmd, util.file, util.vcf
 import read_utils, taxon_filter
 import tools, tools.picard, tools.samtools, tools.gatk, tools.novoalign, tools.muscle
+import tools.trinity
 
 log = logging.getLogger(__name__)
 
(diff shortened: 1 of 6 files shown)


dpark01:
I'm happy with the changes to assembly.py and assembly.rules, but I'm really puzzled why TestBmtagger.setUp is now failing on python 3 only, even though there were no changes that I can tell to any of the Bmtagger stuff until my latest commit. I also checked the original source URL at NCBI and it doesn't appear that any of those files have recent dates on them. Your last commit on the dp-trinity branch passed in Travis, but the pull-request of the same commit onto master failed, but master hasn't changed since the branch was born....

dpark01:
Hm.. seems to be working today mostly. I have an unconfirmed suspicion that maybe it's an issue of running out of disk space on the Travis VM (BLAST and Trinity tools and the bmtagger test indexes are really huge...). Maybe we should do some sort of post-cleanup after tool installation of unnecessary stuff (source trees, etc). But we can do that in some other branch/pull. Merging this for now.

PR #78, opened 2015-01-25
Fix command-line invocation of `python read_utils.py novoindex foobar.fasta`

mlin:
...by renaming argument in NovoalignTool.index_fasta. The argument name now matches the kwargs passed by the argparser.


dpark01:
Oops, my bad!

In other news, it appears that Travis doesn't give forks the encrypted secret key, even when pull-requesting back to the main branch.  I would've thought that it would at least do it on the PR...

Issue #66, opened 2015-01-12
Replace and generalize align_and_orient steps.


dpark01:
The current align_and_orient steps rely on VFAT and some other custom scripts. They should be rewritten to remove dependency on Mosiak and to work properly on multi-chromosome genomes. A good starting point would be to steal Bellini python code for merge_contigs and order_and_orient.

dpark01:
(also, get rid of the Mosaik tool entirely)

dpark01:
Addressed in PR #216

PR #83, opened 2015-02-09
MosaikTool: fix networkFile path calculation

mlin:
`MosaikTool.get_networkFile` calculates the path incorrectly, causing MosaikAligner invocations to fail with:

```
FANN Error 1: Unable to open configuration file "/home/dnanexus/viral-ngs/tools/build/MOSAIK-2.1.33-source/networkFile/2.1.26.pe.100.0065.ann" for reading.
*** glibc detected *** /home/dnanexus/viral-ngs/tools/build/mosaik-2.1.33/bin/MosaikAligner: malloc(): memory corruption: 0x00007f4edc022550 ***
======= Backtrace: =========
[0x55e6e2]
[0x560d48]
[0x56291d]
...
```

It's unclear if/how this affects the pipeline results. At least for the two samples I use for automatic tests (SRR1553416 and SRR1553554), the final assemblies and BAMs are exactly the same with or without this fix.

A more general concern is that `assembly.py order_and_orient` appears to succeed in spite of the MosaikAligner failures. That's probably because of the V-FAT scripts not checking the return value of `system()` calls which is a common perl hazard. We might want to work [autodie](http://search.cpan.org/~nthykier/autodie-2.26/lib/autodie.pm) into those; perhaps this should be a separate to do issue.

Not even gonna get into the segfault... :P

### DIFF
--- a/test/unit/test_tools_mosaik.py
+++ b/test/unit/test_tools_mosaik.py
@@ -1,0 +1,22 @@
+# Unit tests for Novoalign aligner
+
+__author__ = "mlin@dnanexus.com"
+
+import unittest, os.path, shutil
+import util.file, tools.mosaik
+from test import TestCaseWithTmp
+
+class TestToolMosaik(TestCaseWithTmp) :
+
+    def setUp(self):
+        super(TestToolMosaik, self).setUp()
+        self.mosaik = tools.mosaik.MosaikTool()
+        self.mosaik.install()
+
+    def test_get_networkFile(self):
+        nndir = self.mosaik.get_networkFile()
+        assert os.path.exists(os.path.join(nndir, "2.1.26.pe.100.0065.ann"))
+
+    # TO DO: further testing of Mosaik invocations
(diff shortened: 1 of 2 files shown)


dpark01:
Yeah... there's lots here that I don't like, but I appreciate the quick fix to deal with immediate needs. Definitely don't get inspired to address all the V-FAT bugs. I have some thoughts on the longer term solution (and some new unincorporated code that was part of a project to replace V-FAT from scratch). See #66.

I also am not really sure what providing the reads to V-FAT accomplishes... especially since we have *two* rounds of read-based refinement that follow in our own pipeline.

In the meantime, thanks for the bug fix.  Out of curiosity, can you rebase this PR to the latest master (I just put in today)? I put a few Travis-tweaks in there and I'm curious to see if Travis builds your PR after incorporating that.

dpark01:
Oh I'll also mention, if you rebase/merge the latest master, you'll also need to move your unit test from test/ to test/unit/ (I made a space to put long-running tests in test/integration/).

PR #94, opened 2015-02-15
workaround for blastn memory leaks

mlin:
Start a new blastn process for each 1M reads. This got us through depleting the validation per-sample `.raw.bam`'s in under 10GB memory usage.


mlin:
Oops, looks like a python2/3 issue I'll look into. Travis ftw

util/misc.py
```diff
@@ -76,3 +76,31 @@ def median(l):
                 return l[half]
         else:
             raise Exception("empty list for median")
+
+# from http://biopython.org/wiki/Split_large_file
+def batch_iterator(iterator, batch_size) :
```
dpark01: How about the "grouper" function in https://docs.python.org/2/library/itertools.html#recipes

PR #96, opened 2015-02-18
Irwin vphaser

iljungr:
Created vphaser tool (linux and mac executables) and unit test.


test/unit/test_intrahost.py
```diff
@@ -1,12 +1,34 @@
...
+import util.file, tools.vphaser
+from test import TestCaseWithTmp
 
 class TestCommandHelp(unittest.TestCase):
     def test_help_parser_for_each_command(self):
         for cmd_name, parser_fun in intrahost.__commands__:
             parser = parser_fun(argparse.ArgumentParser())
             helpstring = parser.format_help()
+
+class TestVPhaser(TestCaseWithTmp) :
```
dpark01: Can you move this class to test/unit/test_tools_vphaser2.py (a file that does not yet exist)? That can test just the Phaser2 wrapper itself and test_intrahost can deal more with custom code, filters, and end-to-end stuff.

dpark01:
Looks good to me, this looks like it was a lot of work, but it makes everything else downstream much cleaner!

The test doesn't seem to be taking too long on Travis, as far as I can tell.. we can always tinker with that later if it starts to become an issue.

The P-values aren't actually used by us downstream, as far as I know, since we recalculate strand-bias and other statistics ourselves later on. It probably wouldn't be too hard to write a test that just strips the unstable column out and compares that output.

Also, I'd love to add another method on the Vphaser2Tool... maybe something like iterate(self, inBam, numThreads=None), that creates a temp dir, calls self.execute on it, iterates through the .var.raw.txt file output line by line (the only file we care about), then deletes the temp dir when done.  Something like:

````
outdir = tempfile.mkdtemp('vphaser2')
self.execute(inBam, outdir, numThreads)
prefix = 'not sure how this is computed'
outfile = os.path.join(outdir, prefix + '.var.raw.txt')
with open(outfile, 'rt') as inf:
 for line in inf:
  if not line.startswith('#'):
   yield line.strip().split()
shutil.rmtree(outdir)
````

Thoughts?

test/unit/test_intrahost.py
iljungr: Done. Also renamed all vphaser stuff to vphaser2 for consistency.

dpark01:
Instead of a pickle file, since it's just a bunch of lists of strings (and nothing more complicated than that), you could use a tab delimited file to test the expected output.  This would actually be different than the vphaser output file, which is just spaces, I think.

iljungr:
At this point I have the pickle file working (in python 2 and python 3). Any reason to prefer the Tab-delimited?

iljungr:
The iterate method has now been implemented, iterating through all .var.raw.txt files in the order of the chromosomes in the bam file, and prepending the chromosome name to the list for each line.

The unit test was changed to call the iterate method, rather than execute, and its input bam file was changed to have two chromosomes (and shortened, so the test now runs in about 3 seconds).

I tried to set up the test so it would not include any lines with unstable p-values, but changing the input affects which lines have the instability, so I ended up just excluding these fields from the comparison.

dpark01:
No reason to change it if it works now--looks great!

Issue #99, opened 2015-02-20
implement coordinate mapper


dpark01:
Some generic mapping utility (a class called interhost.CoordMapper) that allows us to map coordinates between related genomes.

Example input genomes:

Genome A:
```
>chr1
ATGCATGC
>chr2
GGTTTTC
```

Genome B:
```
>first_chrom
GCATTTGC
>second_chr
GGTTTCCAC
```

Expected outputs:

 - mapAtoB('chr1', 5) should return ('first_chrom', 3)

 - mapBtoA('first_chrom', 7) should return ('chr1', 7)

 - mapAtoB('chr1',1) should ... maybe raise an IndexError or .. maybe return a negative number or something (we can figure that out). Actually, I think IndexError might be better.

 - mapBtoA('first_chrom', 4) or 5 or 6 should all return the same answer

 - mapAtoB('chr1', 6) should return a deterministic answer (perhaps 4)

I've checked a basic framework into interhost.py and test/unit/test_interhost.py, but it's full of NotImplementedErrors (and @unittest.skip decorators).

iljungr:
Some questions:

Running muscle on chr1/first_chrom (with gapopen=-1) and adding numbering to the result yields:
>chr1
12345  678
ATGCA--TGC
>first_chrom
--GCATTTGC
  12345678

> mapAtoB('chr1', 6) should return a deterministic answer (perhaps 4)
> 
What is meant by deterministic? If it only means "produces the same answer on multiple runs with exactly the same input" then I would take whatever muscle maps it to, which in this case gives the answer 6. (I'm assuming muscle itself is deterministic.) If "deterministic" instead means "can easily be explained", then this wouldn't be sufficient since I don't know why muscle aligned things this way, rather than putting the gap after the T.

If we want to do anything other than take whatever answer muscle gave us, it will be complicated, and in principle would require us to do aligning ourselves. We'd essentially be asking if it could it have aligned it some other way. If all that's there is one base aligned to a run of the same base then we could handle that as a special case, but, what if you have something like this:
1234    5678
GGAT----ATGG
GGATATATATGG
123456789012
We could map 3 to either 3 or 7, since the aligner could have put the gap before or after the initial AT with the same score. The ambiguity exists even though position 3 itself is not next to a gap.

For now, in cases where muscle aligns a base to a non-gap base, I will map to that base.


[...]

PR #100, opened 2015-02-25
Irwin coord mapper

iljungr:
- Implemented CoordMapper with unit test. Space usage is not optimized in current implementation, but should be fine for viruses.
- Made muscle quite by default.


iljungr:
https://github.com/broadinstitute/viral-ngs/issues/99

test/unit/test_interhost.py
```diff
@@ -13,17 +13,37 @@ def test_help_parser_for_each_command(self):
...
+            for chrom, mappedPos in result :
+                self.assertEqual(chrom, toChrom)
+            self.assertEqual(expected,
+                             [mappedPos for chrom, mappedPos in result])
+
+        # Check that IndexError is raised when past ends of other sequence
+        for pos in [-1, 0, 1, 2, 22, 23, 24] :
+            try :
+                cm.mapAtoB('chr1', pos)
+                self.assertTrue(False) # Should never get here
```
dpark01: This block can be replaced with a self.assertRaises

dpark01:
This is awesome. I'll stare at the edge cases more carefully later today when I get a chance, but my guess is that this is spot on. Nice thing is that we can refactor the implementation later on for better space utilization, but having the test coverage in place makes that far easier.

iljungr:
There is no need to refactor in order to get better space utilization. Just define a class for storing the information that implements append, __len__, __getitem__, and __setitem__, that efficiently stores the elements in a list if there are many runs of consecutive integers. Then change the one line "map = []" to "map = MyNewClass()".

I'll put comments in the code to that effect.

test/unit/test_interhost.py
iljungr: Thanks. Done.

PR #103, opened 2015-02-28
Irwin better coord mapper

iljungr:
- Rewrote mapping algorithm in CoordMapper to make it simpler and much more space efficient.
- Extracted the class that does the mapping for two already-aligned sequences into CoordMapper2Seqs. This is useful for unit testing cases that are hard or impossible to get out of the aligner. It also might be useful for situations where we already have an alignment, or where we have to adjust the alignment.
- Added unit tests for CoordMapper2Seqs, including checking that the alignment doesn't have adjacent gaps in the two sequences without an intervening aligned pair of real bases.
- Space usage is now a constant plus 8 bytes per indel. For example, a mapper from ebolavirus to Reston virus takes less than 1KB; a mapper from the 28 megabase fly chromosome 3R to mosquito uses less than 400KB.

### DIFF
--- a/interhost.py
+++ b/interhost.py
@@ -150,18 +150,11 @@ class CoordMapper :
                 prevToPos = toArray[insertInd - 1]
                 nextToPos = toArray[insertInd]
                 assert(prevFromPos <= fromPos < nextFromPos)
-                fromLen = nextFromPos - prevFromPos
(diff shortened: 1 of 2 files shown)


dpark01:
Nice use of bisect.

PR #104, opened 2015-03-01
implement iSNV step 2 (merge to VCF)

### LINKED ISSUE #89: iSNV all-sample merge step (step 2)
Each set of iSNV calls starts off separate for each sample. They are in their own coordinate spaces (having been called against their own consensus sequences) and have their own relative definition of "reference" and "alternate" alleles. We'd like to merge this all to a single master table, using consistent coordinates and alleles, and a standard file format (e.g. VCF). Make a new command in intrahost.py (perhaps called something like "merge_to_vcf").

- input:

 - a bunch of marked up text files per sample with iSNV calls, the output of the command described in #88.

 - each individual sample's consensus assembly (fasta) that describes their reference alleles and coordinate space.

 - a master reference assembly (fasta) that describes our target coordinate space and reference alleles for the output file.

- output: a single VCF file

Steps:

1. For each sample:

 a. Align individual assembly to master reference with MUSCLE. Create custom code that reads that alignment and is able to generally map arbitrary coordinates (chr name, position) between the two coordinate spaces at will.

 b. Map all iSNV calls to target coordinate space. Change the position of deletions off-by-one to match the VCF conventions of how to talk about deletions.

 c. Flip alleles (ref & alt) if necessary to match the target reference genome.

2. Across all samples:

 a. Sort and group all data per genomic position. The VCF format describes one row per position in the genome. This requires us to merge together data from all files into a single output row since our input describes each sample separately. But more than that, V-Phaser often describes a single position multiple times for a single sample, if multiple types of variants (SNPs, insertions, deletions) are seen at the same place. In this step, we must merge all variants together across all samples, and pad out their allele lengths with extra genome sequence if necessary.

 b. If V-Phaser had no output at a given position for a sample, reach back into that sample's consensus assembly and record that consensus allele at 100% (V-Phaser only reports variant positions). If the consensus assembly is an N (insufficient read support at this position), don't do this. If the consensus assembly is gapped at this position, figure out the proper indel to put in this place (at 100%).

The existing "vphaser_to_vcf" command is a good starting point to refer to for the code, but it doesn't do all of the above, and it also does a few steps that belong in #88 instead. In particular, it doesn't do the MUSCLE coordinate remapping and doesn't handle any consensus-level indels, because none of these existed in our Summer 2014 data set, but they definitely exist in our current EBOV data. (they existed in previous Lassa data sets, but we weren't using V-Phaser back then)

Clearly, this step will need some decent unit testing...

dpark01:
closes #89


dpark01:
Hi Irwin,

I think I implemented this right... can you glance over it a bit and merge it in if it looks okay?

Also, note that none of this actually utilizes the AlleleFieldParser yet and is based on the old allele fields out of VPhaser. My suggestion would be to look this over as-is, merge it in if it looks good, then merge the new master back onto your branch to incorporate this stuff, then update it (both the implementation and the unit tests) to use the AlleleFieldParser later.

dpark01:
When the time comes to incorporate the strand bias / library count fields, I have thoughts about how to represent that in the VCF output.

intrahost.py
```diff
@@ -129,157 +169,211 @@ def parser_tabfile_rename(parser=argparse.ArgumentParser()):
...
+                                key=(lambda x:x[1]), reverse=True))
+                            # reposition vphaser deletions minus one to be consistent with
+                            # VCF conventions
+                            if row['s_alt'].startswith('D'):
+                                for a,n in row['allele_counts']:
+                                    assert a[0] in ('D','i')
+                                row['s_pos'] = row['s_pos']-1
+                            # map position back to reference coordinates
+                            row['POS'] = samp_to_cmap[s].mapAtoB(s_chrom, row['s_pos'], side = -1)[1]
+                            row['END'] = samp_to_cmap[s].mapAtoB(s_chrom, row['s_pos'], side = 1)[1]
```
iljungr: This will raise if position is outside either end of the reference sequence. Can we safely assume that can't happen? If not, do we want to throw out any iSNVs beyond the reference ends? Similarly, down below, when we mapBtoA, can we assume that consensus is full length of reference? If not, we can treat it as if it were padded with Ns. Unless we assure in advance that reference and all consensuses have the same extent, I am inclined to change CoordMapper to return None rather than raising when past end so the caller can check result instead of putting a try block; generally, I save raise for cases where the caller has done something wrong, not for cases where the caller has no way to know in advance if the call will be valid or not.

dpark01:
Good catch. In this particular use case, the consensus assemblies will never exceed the reference genome, because of how they are constructed (there's a part during assembly where we lop off anything that exceeds the reference because we simply don't trust what might be there).  But the other direction is not true, so you're right, we should handle that case and simply not populate the consensus (similar to how we currently drop any consensus alleles that have ambiguity bases in them).

If you'd like to switch the return value to None (and handle appropriately) that's fine. You might also consider pushing the logic for the "side" argument further upstream into the mapper class (or maybe just have the mapper class always return a (left,right) pair, even if left==right).

intrahost.py
```diff
@@ -129,157 +169,211 @@ def parser_tabfile_rename(parser=argparse.ArgumentParser()):
...
+                        if 'i' in acounts and 'd' in acounts:
+                            # this sample has both insertions and deletions at the same spot!
+                            # average the two (discordant) reference allele counts and
+                            # drop one of them (so we have only one reference allele
+                            acounts['i'] = int(round((acounts['i'] + acounts['d'])/2.0,0))
+                            del acounts['d']
+                        
+                        if acounts:
+                            # we have iSNV data on this sample
+                            consAllele = consAlleles[s]
```
dpark01: Based on your last comment, I just realized this assumption (that s in consAlleles) is not always right. It is true that iSNV calls can only exist where non-ambiguous consensus bases exist, so we'll never get an iSNV at a position that has an N. The problem is that there's logic upstream that expands the (pos,end) coordinates and make the variant bigger if necessary because of action happening in other samples. As such, an N might exist (or it might exceed the end of the consensus) and consAlleles might not have s.

How about we simply replace "if acounts" with "if acounts and s in consAlleles"?  If s is not in consAlleles, we shouldn't trust our ability to reconstruct the alleles for this sample, and the logic below should properly populate that sample with "no data" marks (the "." symbol).

iljungr:
I will:
- Change CoordMapper to return None instead of raising
- Assert that doesn't happen when mapping to reference
- Handle it when mapping from reference to consensus
- Think about always returning left-right pair

intrahost.py
```diff
@@ -129,157 +169,211 @@ def parser_tabfile_rename(parser=argparse.ArgumentParser()):
...
+                                # end of deletion in sample's coord space
+                                local_end = row['s_pos']+int(a[1:])
+                                # end of deletion in reference coord space
+                                ref_end = samp_to_cmap[row['sample']].mapAtoB(row['s_chrom'], local_end, side = 1)[1]
+                                end = max(end, ref_end)
+                
+                    # find reference allele and consensus alleles
+                    refAllele = str(ref_sequence[pos-1:end].seq)
+                    consAlleles = {} # the full pos-to-end consensus assembly sequence for each sample
+                    samp_offsets = dict((row['sample'], row['s_pos']) for row in rows) # the isnv's index in the consAllele string for each sample
```
iljungr: If consensus for a sample has an insertion relative to reference, then can't several values of s_pos in that sample map to the same POS? For example:
Ref:      A--A
Cons:   AAAA
Allele1: ATAA, POS=1, s_pos=2
Allele2: AATA, POS=1, s_pos=3
In that case, when we form this dictionary we would be adding several s_pos values with the same sample value and the result will depend on the order in which they are added.

dpark01: I *think* I'm handling this scenario but we should add an explicit unit test for it... I don't think the result should depend on the order of the input...

--
Daniel J. Park, PhD
Tel: +1-617-714-8526
dpark@broadinstitute.org

intrahost.py
```diff
@@ -129,157 +169,211 @@ def parser_tabfile_rename(parser=argparse.ArgumentParser()):
...
+                            row['POS'] = samp_to_cmap[s].mapAtoB(s_chrom, row['s_pos'], side = -1)[1]
+                            row['END'] = samp_to_cmap[s].mapAtoB(s_chrom, row['s_pos'], side = 1)[1]
+                            data.append(row)
+            
+                # sort all iSNVs (across all samples) and group by position
+                data = sorted(data, key=(lambda row: row['POS']))
+                data = itertools.groupby(data, lambda row: row['POS'])
+            
+                # process one reference position at a time
+                for pos, rows in data:
```
iljungr: Do we require that in the end alleles at different positions are non-overlapping? If so, I don't think we can handle one POS at a time like this. For example:
Reference:          AAAA
Sample1 Cons:   AAAA
Sample1 Variant: AATA
Sampe2 Cons:    A--A
Sample2 Variant: AC-A
We now have 4 alleles with POS=1: AAA, A--, AC-, 
and 2 alleles with POS=3: A, T
Is it OK that the AAA at POS=1 overlaps the allele at POS=3?

intrahost.py
dpark01: Actually, I just checked: test_snp_within_insertion and test_deletion_within_insertion already tests this scenario (multiple s_pos mapping to a single POS)

intrahost.py
dpark01: So, in practice, very few, if any, callers will correctly handle this, and also, the GATK IndelRealigner (which we call prior to V-Phaser) should try to minimize the real-life scenarios in which we might encounter something like this.  Currently, this code does not attempt to do anything with this, and they are reported on independently.

However, one could conceivably replace the itertools.groupby line with something that does some sort of overlapping intervals function on the POS-END interval (perhaps using an R-Tree for optimal implementation). That would solve everything actually, the rest of the downstream code should merge everything smartly.

I'm not sure how much to worry about this scenario though. It wouldn't be hard to detect the scenario and push a log.warn if it happens.  After the groupby, everything is already sorted on POS.  After the code block that determines "end" based on the longest variant, we could save it in some variable that persists outside of the for loop and then at the top of the loop, if max_previous_end >= pos: log.warn("overlapping variants in VCF!").

intrahost.py
iljungr: Sorry if I was unclear. The situation I am talking about requires more than one varying position in the _same_ sample. (In my example, Allele1 and Allele2 are in the same sample, with consensus sequence Cons.) The tests you mentioned have only one variant line per sample. To demonstrate the problem I'm talking about, in test_snp_within_insertion change merger.add_snp('s3', 's3_1',... to merger.add_snp('s2', 's2_1',...

dpark01:
Yeah so the more I think about it, I think it's a bad idea to attempt to tackle the merge-multiple-variants-in-same-sample scenario, except for the explicitly necessary case of the insertions + deletions at the same position (but vphaser doesn't call it the same position) scenario.

The reason is that you'll never be able to resolve the haplotypes / phasing.  So if an iSNP has a variant at one position, and two bases down, another iSNP in the same sample, and then some big indel in some other sample slurps everybody together, then you're not going to know what haplotypes really exist in the sample with the two iSNPs, without getting into the reads and trying to phase it.  This is the job of the variant caller, not the downstream scripts that attempt to interpret its output.  This is one reason why GATK has moved from UnifiedGenotyper (treat each position independently) to its HaplotypeCaller (treat whole regions together).

Anyway, I think it's best not to worry about the overlapping scenario other than what's already being handled out of necessity.  If we want, we can put in a detect and log.warn bit of code.

PR #107, opened 2015-03-07
Irwin lib bias

iljungr:
- Added library counts to the output of vphaser_one_sample using samtools mpileup. For snps the counts add up to vphaser's count in most cases, with some off by just a little. For indels the counts are off by a bit more, but probably close enough for library-bias calculation.
- Computed library-bias p-values for the case of 2 libraries. I implemented Fisher's exact test in python, using Stirling's formula to make it fast.
- Changed all code to use AlleleFieldParser to set or access the counts for each allele. Current format is allele:fcount:rcount:fcountLib1:rcountLib1:fcountLib2:rcountLib2:...:libBiasPval. If we want a different format the only code that needs to be changed is the AlleleFieldParser class.
- Created a command-line command for vphaser that just calls Vphaser2Tool().iterate and dumps the result to a file.
- I have not yet created any unit tests for the library counts. It will take a little while to create a bam file with multiple libraries that has a good set of indels that survive V-Phaser 2 filtering, and for which V-Phaser 2 runs in an acceptably short amount of time. I figured it is better to get the rest in now so people can start using the counts while I work on that.


iljungr:
I just fixed two bugs in my python fisher_exact and sped it up. At this point, it is accurate enough for us (around 11 digits) and much faster than the scipy version, so there's no need for us to install scipy for the 2x2 case. Since there are other issues with chi-sq besides the install (handling of small values), it is unclear at this point if we need scipy at all.

iljungr:
Added unit test with 2 libraries for test_vphaser_one_sample. For now, only tests snps, not indels.

Issue #119, opened 2015-04-07
Create simple MAFFT command line entry point


dpark01:
In interhost.py, create a new command called "align_mafft".  It should take one or more input fasta files and produce one output fasta alignment file.  You may expose some of MAFFT's optional parameters via argparse.  Some of the ones that we will likely use include: --localpair --maxiterate --reorder --ep --preservecase and --thread.

Depends on #118

tomkinsc:
Addressed in #120

Issue #118, opened 2015-04-07
Toolify MAFFT


dpark01:
Create new Tool subclass for the MAFFT multi-aligner, available here: http://mafft.cbrc.jp/alignment/software/

MAFFT has precompiled MacOSX and Linux binaries, as well as source.  Try the Linux "portable package" version (not the RPM or DEB version) and the Mac "all-in-one" version (basically, don't ever assume you have root privs).  If the Linux binary succeeds on Travis as well as Broad CentOS (copper, tin, nickel) and Broad RHEL (gold, silver), then we're all set and feel free to skip the source-compile installer.  Otherwise, you might need to implement a compile-from-source step.

The MUSCLE tool has an example of a two-installer approach depending on the OS.  If the OS is Linux or Darwin x86, two install methods are appended: the first downloads the precompiled binary, the second downloads and unpacks the source and calls make.  If the OS is not one of those two known ones, then only the source installer is appended.

nosetests -v test.unit.test_tools.TestToolsInstallation will tell you if your installer succeeded.  On your local Mac checkout, it may fail on some other tools (like m-vicuna and v-phaser2) unless you have a special build environment set up, but you can ignore those errors.

tomkinsc:
Addressed in #120

PR #120, opened 2015-04-08
Addition of MAFFT as argparse-accessible tool

tomkinsc:
Per issues #118 and #119: MAFFT has been added as an argparse-accessible tool to interhost.py.

The installation test [passes](https://travis-ci.org/broadinstitute/viral-ngs/builds/57679951), and the alignment has been tested for function manually on Max OSX 10.10, Ubuntu 14.04, and the Broad's RHEL and CentOS servers.

### DIFF
--- a/interhost.py
+++ b/interhost.py
@@ -13,7 +13,7 @@ try :
     from itertools import zip_longest
 except ImportError :
     from itertools import izip_longest as zip_longest
-import tools.muscle, tools.snpeff
+import tools.muscle, tools.snpeff, tools.mafft
 import util.cmd, util.file, util.vcf
 from collections import OrderedDict, Sequence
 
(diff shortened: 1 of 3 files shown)


dpark01:
looks great.  A few comments will follow.

interhost.py
```diff
@@ -204,6 +204,60 @@ def parser_snpEff(parser=argparse.ArgumentParser()):
...
+# =======================
+# ***  align_mafft  ***
+# =======================
+
+def parser_align_mafft(parser=argparse.ArgumentParser()):
+    parser.add_argument('inFastas', nargs='+',
+        help='Input FASTA files.')
+    parser.add_argument('outFile',
+        help='Output file containing alignment result (default format: FASTA)')
+    parser.add_argument('--localpair', default=None, action='store_true',
```
dpark01: --localpair and --globalpair can be a mutually exclusive group:
https://docs.python.org/3/library/argparse.html#mutual-exclusion
(this doesn't negate the need to check within the Tool.execute though, as you are already doing)

tools/mafft.py
```diff
@@ -0,0 +1,157 @@
...
+        inputFiles = []
+        for f in inFastas:
+            inputFiles.append(os.path.abspath(f))
+        outFile = os.path.abspath(outFile)
+
+        # if multiple fasta files are specified for input
+        if len(inputFiles)>1:
+            # combined specified input files into a single temp FASTA file so MAFFT can read them
+            tempFileSuffix = ""
+            for filePath in inputFiles:
```
dpark01: is this just gluing together all the filenames as part of the temp filename?

tools/mafft.py
```diff
@@ -0,0 +1,157 @@
...
+                tempFileSuffix += "__" + os.path.basename(filePath)
+            tempCombinedInputFile = util.file.mkstempfname('__combined.{}'.format(tempFileSuffix))
+            with open(tempCombinedInputFile, "wb") as outfile:
```
dpark01: any reason you're reading/writing binary files? (fasta files are text)

interhost.py
```diff
@@ -204,6 +204,60 @@ def parser_snpEff(parser=argparse.ArgumentParser()):
...
+    parser.add_argument('--gapOpeningPenalty', default=1.53,
+        help='Gap opening penalty (default: %(default)s).')
+    parser.add_argument('--ep',
+        help='Offset (works like gap extension penalty).')
+    parser.add_argument('--verbose', default=False, action='store_true',
+        help='Full output (default: %(default)s).')
+    parser.add_argument('--outputAsClustal', default=None, action='store_true',
+        help='Write output file in Clustal format rather than FASTA')
+    parser.add_argument('--maxiters', default = 0,
+        help='Maximum number of refinement iterations (default: %(default)s).')
```
dpark01: Is zero the actual default for MAFFT?  Maybe in the help string you should also note that activating localpair or globalpair will change the default to 1000, unless overridden here.

tools/mafft.py
```diff
@@ -0,0 +1,157 @@
...
+        # change the pwd, since the shell script that comes with mafft depends on the pwd
+        # being correct
+        pwdBeforeMafft = os.getcwd()
+        os.chdir(os.path.dirname(self.install_and_get_path()))
+
+        # build the MAFFT command
+        toolCmd = [self.install_and_get_path()]
+
+        toolCmd.append("--auto")
+        toolCmd.append("--thread {}".format( max( int(threads), 1 )) )
```
dpark01: according to MAFFT web site, a value of -1 is how to ask MAFFT to auto-detect the available cores, so we should probably allow a value of -1 (perhaps even default to it), but not allow a value of zero or other negative values.

dpark01:
I *think* I just told Coveralls to chill out if coverage decreases less than 1%, so we'll see what it does on the next commit.

interhost.py
tomkinsc: Nice. In the next commit they shall be.

tools/mafft.py
tomkinsc: The first block gets the absolute filepath for each file specified since the subprocess pwd can vary by platform and relative paths can fail. The next bock concatenates multiple FASTA files into one temp file since MAFFT expects a single file as input.

tools/mafft.py
tomkinsc: No reason apart from habit of working with binary files. I'll remove that flag.

interhost.py
tomkinsc: If you run MAFFT directly, it specifies in its help that the default is zero, and in the docs suggests a default of 1000 for localpair or globalpair alignment (hence why that value is used as default in that case). I'll add a note.
--maxiterate # : Maximum number of iterative refinement, default: 0

tools/mafft.py
tomkinsc: Nice find. I'll change the default, and add a check for zero and other negative values.

interhost.py
```diff
@@ -204,6 +204,64 @@ def parser_snpEff(parser=argparse.ArgumentParser()):
...
+    group = parser.add_mutually_exclusive_group()
+    group.add_argument('--localpair', default=None, action='store_true',
+        help='All pairwise alignments are computed with the Smith-Waterman algorithm.')
+    group.add_argument('--globalpair', default=None, action='store_true',
+        help='All pairwise alignments are computed with the Needleman-Wunsch algorithm.')
+    parser.add_argument('--preservecase', default=None, action='store_true',
+        help='Preserve base or aa case, as well as symbols.')
+    parser.add_argument('--reorder', default=None, action='store_true',
+        help='Output is ordered aligned rather than in the order of the input (default: %(default)s).')
+    parser.add_argument('--gapOpeningPenalty', default=1.53,
```
dpark01: --gapOpenPenalty and --ep should be type=float

Issue #121, opened 2015-04-09
Make MAFFT alignment tool work with multiple samples containing multiple chromosomes


tomkinsc:
In the case where MAFFT is being used to align sequences stored in sample-specific FASTA files, MAFFT should be able to do an alignment among the respective chromosomes from the FASTA files, such that the output is a series of chromosome-specific alignments in FASTA files.

In the case of two samples from organisms with two chromosomes, the input and output would be as follows:

```
Input: 
    sample1.fasta would contain chr1, chr2
    sample2.fasta would also contain chr1, chr2
Output:
    chr1.fasta would contain chr1_sample1, chr1_sample2
    chr2.fasta would contain chr2_sample1, chr2_sample2
```

Before MAFFT, an intermediate step would be needed to transpose the chromosomes from their respective FASTA files to chromosome-specific FASTA files prior to alignment. This can be done via ```itertools.izip```(Python2)/```zip```(Python3) and ```Bio.SeqIO.parse()```, which returns an iterable.

tomkinsc:
Added in commit represented in pull request #122

PR #122, opened 2015-04-09
Added multichr_mafft

tomkinsc:
Added argparse function multichr_mafft to perform alignment on multiple
chromosomes stored in sample-specific FASTA files. Added a helper
function used as a pre-processing step to MAFFT that transposes
chromosome info from separate sample-specific FASTA files into FASTA
files each containing respective-records

### DIFF
--- a/interhost.py
+++ b/interhost.py
@@ -10,17 +10,10 @@ import Bio.AlignIO
 from Bio import SeqIO
 import argparse, logging, os, sys, array, bisect
 
-# try to import izip (in the case of Python 2). 
-# No alternate needed as Python3 zip() is iterator-based.
-try:
-    from itertools import izip
-except ImportError:
-    pass
-
 try:
     from itertools import zip_longest
 except ImportError :
-    from itertools import izip_longest as zip_longest
+    from itertools import izip_longest as zip_longest, izip as zip
 import tools.muscle, tools.snpeff, tools.mafft
 import util.cmd, util.file, util.vcf
 from collections import OrderedDict, Sequence
@@ -216,11 +209,7 @@ __commands__.append(('snpEff', parser_snpEff))
 # ***  align_mafft  ***
 # =======================
 
-def parser_align_mafft(parser=argparse.ArgumentParser()):
-    parser.add_argument('inFastas', nargs='+',
-        help='Input FASTA files.')
-    parser.add_argument('outFile',
-        help='Output file containing alignment result (default format: FASTA)')
+def parser_general_mafft(parser=argparse.ArgumentParser()):
     group = parser.add_mutually_exclusive_group()
     group.add_argument('--localpair', default=None, action='store_true',
         help='All pairwise alignments are computed with the Smith-Waterman algorithm.')
@@ -242,6 +231,16 @@ def parser_align_mafft(parser=argparse.ArgumentParser()):
         help='Maximum number of refinement iterations (default: %(default)s). Note: if "--localpair" or "--globalpair" is specified this defaults to 1000.')
     parser.add_argument('--threads', default = -1, type=int,
         help='Number of processing threads (default: %(default)s, where -1 indicates use of all available cores).')
+    return parser
+
+def parser_align_mafft(parser):
+    parser = parser_general_mafft(parser)
+
+    parser.add_argument('inFastas', nargs='+',
+        help='Input FASTA files.')
+    parser.add_argument('outFile',
+        help='Output file containing alignment result (default format: FASTA)')
+
     util.cmd.common_args(parser, (('loglevel', None), ('version', None), ('tmpDir', None)))
     util.cmd.attach_main(parser, main_align_mafft)
     return parser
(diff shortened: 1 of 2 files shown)


tomkinsc:
Committed to address issue #121

dpark01:
The description leads me to believe that this function requires identical sequence IDs in each input file (e.g. "chr1", "chr2"), and then it performs a renaming of each sequence ID (in the FASTA header lines) in the output file (appending the filename or something to each), but that doesn't appear to be the case, looking at the code.

I was thinking about this a bit last night and was realizing that there are some cases where we might want to do some renaming of sequence IDs (because the inputs might be identically named) and other scenarios where we wouldn't want to touch it at all (because everything was handled upstream just fine) and we wouldn't want to impose the renaming by default.. maybe just optionally, but we can deal with that later.

But I think what would be a good idea is to add a sanity check (this can be in the MAFFT Tool.execute itself, after it concatenates its input files) that errors out if the FASTA that serves as input to MAFFT has non-unique sequence IDs.  I don't even know if MAFFT would error anyway, but we certainly wouldn't know how to interpret its output even if it didn't error.

interhost.py
```diff
@@ -262,6 +270,94 @@ def main_align_mafft(args):
...
 __commands__.append(('align_mafft', parser_align_mafft))
 
+# =======================
+# ***  multichr_mafft  ***
+# =======================
+
+def parser_multichr_mafft(parser=argparse.ArgumentParser()):
+    parser.add_argument('inFastas', nargs='+',
+        help='Input FASTA files.')
+    parser.add_argument('--outDirectory', default=os.getcwd(),
```
dpark01: let's make outDirectory explicit/required

interhost.py
```diff
@@ -262,6 +270,94 @@ def main_align_mafft(args):
...
+        help='Location for the output files (default is cwd: %(default)s)')
+    parser.add_argument('--outFilePrefix', default="singlechr",
+        help='Prefix for the output file name (default: %(default)s)')
+    group = parser.add_mutually_exclusive_group()
+    group.add_argument('--localpair', default=None, action='store_true',
```
dpark01: maybe make a python method that takes a parser and adds all these MAFFT-specific arguments to it (from --localpair down to --threads.. not including things like inFastas or outFilePrefix) and then just call it in both of the mafft parsers so as to reduce duplication?

interhost.py
```diff
@@ -262,6 +270,94 @@ def main_align_mafft(args):
...
+
+    # get the absolute path to the output directory in case it has been specified as a relative path,
+    # since MAFFT relies on its CWD for path resolution
+    absoluteOutDirectory = os.path.abspath(args.outDirectory)
+
+    # prefix for output files
+    prefix = "" if args.outFilePrefix == None else args.outFilePrefix
+
+    # reorder the data into new FASTA files, where each FASTA file has only variants of its respective chromosome
+    transposedFiles = transposeChromosomeFiles(args.inFastas, absoluteOutDirectory, prefix)
```
dpark01: I don't like the temp files going into the absoluteOutDirectory.  I also don't particularly see why the temp files need to have human-readable names.  Perhaps have transposeChromosomeFiles just spit out a bunch of files that were created with util.file.mkstempfname and then move the code that creates the final output filename string into the for loop right below here?

Then the post-cleanup code below becomes much simpler, and even optional (since the util.cmd framework makes a point of blowing away the tempfiles after exit anyway).

interhost.py
```diff
@@ -8,8 +8,16 @@
...
+import argparse, logging, os, sys, array, bisect
+
+# try to import izip (in the case of Python 2). 
+# No alternate needed as Python3 zip() is iterator-based.
+try:
+    from itertools import izip
+except ImportError:
+    pass
+
+try:
```
dpark01: How about merging the two try/import blocks?

```
try:
   from itertools import zip_longest
except ImportError :
   from itertools import izip_longest as zip_longest, izip as zip
```

Then farther down below, you can get rid of your sys.version_info thing (what if Python 4.0 comes around?) and just call "zip".  In the case of Python 2, the "from itertools import izip as zip" will overwrite "zip" in the local namespace with the itertools version.

interhost.py
```diff
@@ -304,6 +400,35 @@ def make_vcf(a, ref_idx, chrom):
                             vars.append(m+1)
             yield row+vars
             
+def transposeChromosomeFiles(inputFilesList, outputDirectory, outputFilePrefix, inputFormat="fasta", outputFormat="fasta"):
+    outputFilenames = []
+
+    # get BioPython iterators for each of the FASTA files specified in the input
+    fastaFiles = [SeqIO.parse(open(x, "rU"), inputFormat) for x in inputFilesList]
```
dpark01: Don't forget to close all these files before the end of transposeChromosomeFiles.  I think you'll have to do something like:

```
inFastas = list(open(x, 'rU') for x in inputFilesList)
fastaFiles = [SeqIO.parse(x, inputFormat) for x in inFastas]
```

... then at the end ...

```
for x in inFastas:
  x.close()
```

interhost.py
tomkinsc: Sounds good. I generally try to avoid overwriting built-ins, your point is valid and I'll change it.

interhost.py
tomkinsc: Sure, I'll combine them. I had them separate because I wasn't sure if the defaults might be different for the multichr version.

interhost.py
tomkinsc: OK.

interhost.py
tomkinsc: I'll make the appropriate revisions.

tomkinsc:
The new commit includes changes from the line comments above, and also a check for uniqueness of the sequence IDs seen before the MAFFFT tool

dpark01:
Closes #121

PR #123, opened 2015-04-09
manual closure of input files in transposeChromosomeFiles()

tomkinsc:
input files are now closed in transposeChromosomeFiles(), before the
CPython GC would do so upon returning to the calling function,
multichr_mafft(). Number of open file handles confirmed via lsof

### DIFF
--- a/interhost.py
+++ b/interhost.py
@@ -373,12 +373,14 @@ def make_vcf(a, ref_idx, chrom):
                         if a[k][i] == alt[m]:
                             vars.append(m+1)
             yield row+vars
-            
-def transposeChromosomeFiles(inputFilesList, outputDirectory, outputFilePrefix, inputFormat="fasta", outputFormat="fasta"):
+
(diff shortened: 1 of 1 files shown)


tomkinsc:
My understanding of Python 3 was that generator expressions (and iterators, maps, etc.) were lazily-evaluated, but that list comprehensions were specifically implemented as non-lazy--in effect, consuming the iterator during the list creation operation, like saying list(generator_exp).

If the code is instrumented before/after the file closure step via calls to the function below, the reduction in the number of open handles seems the same with the list comprehension as with using a for loop to close the files (and attempts to write to the files also fail after the list comprehension).

```
def get_num_open_files():
    '''
    return the number of open file handles for this process

    '''
    import subprocess
    import os

    pid = os.getpid()
    procs = subprocess.check_output(
        [ "lsof", '-Ff', '-w', "-p", str( pid ) ] )
    procs = procs.decode('utf-8').split('\n')
    numFiles = len(list(filter(lambda x: x and x[0] == 'f' and x[1:].isdigit(), procs)))

    return numFiles
```

dpark01:
Welcome back @iljungr!
Ok let's run with this

PR #146, opened 2015-08-25
UGER-related fixes

tomkinsc:


pipes/rules/demux.rules
```diff
@@ -117,6 +118,7 @@ rule illumina_basecalls:
     output: config['tmpDir']+'/'+config['subdirs']['demux']+'/bams_per_lane/{flowcell}.{lane}/Unmatched.bam'
     resources: mem=64
     params: LSF=config.get('LSF_queues', {}).get('bigmem', '-q flower'),
+            UGER=config.get('UGER_queues', {}).get('long', '-q long -l m_mem_free=40G'),
```
dpark01: This -l m_mem_free=40G seems unnecessary if the cluster-submitter.py is already translating the mem resource properly into UGER-ese?

tomkinsc: Quite right! It's a vestigial change left from learning how snakemake passes params. `mem` is indeed being passed through. The `-l m_mem_free` flag will be removed in the next commit.

Issue #136, opened 2015-07-06
update location of dynamically-linked library specified in variant_caller?


tomkinsc:
This may be a quirk of my setup (OSX 10.10.4 with Xcode 6.4 installed and gcc 4.9 via [Homebrew](http://brew.sh/)), but I had to change the variant_caller binary to have the correct location for the libgomp.1.dylib library via the `install_name_tool` command, as described [here](http://stackoverflow.com/a/13421926):

    sudo install_name_tool -change /usr/local/opt/gcc49/lib/gcc/x86_64-apple-darwin13.4.0/4.9.1/libgomp.1.dylib /usr/local/opt/gcc/lib/gcc/4.9/libgomp.1.dylib /Users/tomkinsc/git-repositories/viral-ngs/tools/binaries/V-Phaser-2.0/MacOSX/variant_caller

Without this fix, attempts to run commands that would call variant_caller (including unit tests) would result in the following error:

    tools.vphaser2: DEBUG: OMP_NUM_THREADS=4 /Users/tomkinsc/git-repositories/viral-ngs/tools/binaries/V-Phaser-2.0/MacOSX/variant_caller -i /Users/tomkinsc/git-repositories/viral-ngs/test/input/TestPerSample/in.indels.bam -o /var/folders/bx/90px0g1n1v122slzjnyvrsk5gn42vm/T/tmp-TestPerSample-Ykv2o6/tmpvsT9FGvphaser2
    tools.vphaser2: ERROR: dyld: Library not loaded: /usr/local/opt/gcc49/lib/gcc/x86_64-apple-darwin13.4.0/4.9.1/libgomp.1.dylib
    Referenced from: /Users/tomkinsc/git-repositories/viral-ngs/tools/binaries/V-Phaser-2.0/MacOSX/variant_caller
    Reason: image not found

tomkinsc:
This also appears to be an issue with mvicuna, at least for my setup.

tomkinsc:
So it *does* work to copy the .dylib to the same directory as the binary and then change the link to be relative.

tomkinsc:
The command to use to set the binary to use a relative path to the library makes use of the `@executable_path` variable:

`install_name_tool -change /usr/local/opt/gcc49/lib/gcc/x86_64-apple-darwin13.4.0/4.9.1/libgomp.1.dylib @executable_path/libgomp.1.dylib ./variant_caller`

dpark01:
[...]

Issue #135, opened 2015-06-17
Added multi_align_mafft snakemake rule


tomkinsc:
Add variable `number_of_chromosomes` to `config.json` file to represent number of chromosomes/segments in the organism. To the `interhost.rules` file, a "`multi_align_mafft`" rule should be added to call `./interhost.py multichr_mafft`, passing in k assembly files where k is `number_of_chromosomes` specified in the `config.json` file.

tomkinsc:
Included in ef8e045b0f4628cf1a974a356668099f31d348cd

Issue #134, opened 2015-06-02
update ncbi annotation transfer to use pre-built alignments


dpark01:
ncbi.py tbl_transfer currently takes a pair of fasta files and pairwise aligns them.

Create a new command in ncbi.py: tbl_transfer_prealigned, that does the same thing as tbl_transfer, but instead of taking two fasta files as input, it takes pre-made alignments as input and also takes as input the name of the reference sequence and the name of the alternate sequence in the alignment.  The input alignment may contain more than these two sequences, but the other sequences will be ignored.

Also update the Snakemake rule (rule annot_transfer in interhost.rules) accordingly.

This depends on #127 .

tomkinsc:
ncbi.py has been updated in 4aab78ecbdb8ef3a096f87500892987101629b7a to include `tbl_transfer_prealigned`, but it's unclear to me how the snakemake rule will make use of it. The rule can get the reference sequence name from the reference fasta file (as specified in the config), but how will it know the name of the alternate sequence? During the [transposition step](https://github.com/broadinstitute/viral-ngs/blob/master/interhost.py#L82) of a mafft multialignment, the relationship of sample->{seq1, seq2, seq3} is lost, since the mafft alignment creates a separate output fasta file for each chromosome. Should we be including the sample name in the sequence ID as part of the transposition step, or otherwise tracking the relationship?

dpark01:
So, as I look at what NCBI wants as a batch submission, I'm realizing that the current code creates one tbl file per sample, but I think we will need to actually create one tbl file per sample per chromosome...

We had discussed earlier about how the config.json is going to need to have a new variable that indicates the number of chromosomes per genome.  Snakemake rules can then split out multiple files if necessary for certain steps, or pass the number as a parameter to a script that needs to be told that number.

Issue #133, opened 2015-06-02
update iSNV pipeline to use pre-built alignments


dpark01:
intrahost.py merge_to_vcf currently has an --assemblies required input parameter that takes in a ton of fasta files (one per sample).  It then pairwise aligns everything before doing anything useful, and this often takes an hour or so, and is very unoptimal.

Instead, it should take in alignments in an --alignments required input parameter that takes in one aligned fasta file per chromosome/segment.  This must contain all samples as well as the reference genome.

Also update the Snakemake rule (rule isnvs_vcf in intrahost.rules) accordingly. 

This depends on #127 .

tomkinsc:
In updating the Snakemake rule (isnvs_vcf in intrahost.rules), should a multiple alignment including the reference be performed (via mafft?) as part of the isnvs_vcf rule, or should that be assumed to be a prerequisite for the rule that will occur elsewhere beforehand? A call to `./interhost.py multichr_mafft`could be added as the first step in the list of commands to run as part of the rule, accepting the list of assembly files (assuming the assemblies have the same number of chromosomes).

tomkinsc:
Following a live discussion with @dpark01 , multichr_mafft alignment of assemblies will be assumed to be a step preceding isnvs_vcf.

tomkinsc:
Included in ef8e045b0f4628cf1a974a356668099f31d348cd

PR #149, opened 2015-08-31
Ct multialign

tomkinsc:
This PR addresses issues #127, #133, #134, #135, and #136. It also includes the new functionality of being able to download FASTAs, feature tables, and full GenBank records via the Entrez REST API. This functionality is used in the new Snakemake rules to download the files required to make the reference genome, lastal database, and feature tables, all via GenBank accessions specified in the file config.json. It also adds new scripts related to running the Snakemake pipeline on the Broad Institute's Univa Grid Engine cluster ("UGER").


dpark01:
Okay, that's a big PR! I've done a quick skim through. I think you can change the author line at the top of the ncbi.rules snakemake file to yourself (you probably copied it over from interhost.rules when you made that new file). There are a few straggling debug print statements that are commented out that could probably be removed, but no big deal. Main thing I want to sit down and understand better is the snpEff database building stuff (something about a list of accessions becoming a unique hash)--maybe you can explain it tomorrow sometime?

I also have some mixed feelings about unit tests that require a 3rd-party web site to be up and functional (NCBI fetching). This already with some of these download-and-compile-a-tool tests (but those get skipped due to Travis build caching for most unit tests with the exception of the pull requests). But I've also found those tests to be some of the more unstable ones (particularly the ones that download stuff from sourceforge--those occasionally fail). Unit tests are one of those things you'd like people to be able to run repeatedly again and again, so if an NCBI filter blacklists someone who runs our unit tests a few times, that could be annoying for them, but also potentially bad for their institution (NCBI tends to do very coarse blacklisting of subnets).  Maybe we could move the full download-a-genome-from-NCBI stuff to the integration tests, and for the unit tests, can you mock out the network connectivity part and test all your associated code by feeding the results (that are already in your test directories) out as if it came from the network?

tomkinsc:
The concern is warranted about the unit tests (in test_ncbi.py) hitting the NCBI as a third-party. I had a few Travis builds fail there, I suspect because of rate limits. The NCBI fetching functions themselves are pretty simple—mostly just file concatenation and removal—so I'm not sure it is worthwhile to mock out the server requests. I see the main value of the tests as validation that we are interacting with the NCBI API in the correct way for parameters like db name, and rettype. I've moved the tests to `integration/` as you suggest, so we will still have live checks but only in the event of PRs or long-test runs. Do you still think I should mock them out as unit tests? If so, is there a library/pattern you would prefer? The NCBI "EFetch" requests are being handled by Biopython. Something like [HTTPretty](https://github.com/gabrielfalcao/HTTPretty) could make sense for mocking since it intercepts the requests at the Python socket layer, and the tests would then be agnostic of the request library used by Biopython.

We can absolutely look at the snpEff database building [function](https://github.com/broadinstitute/viral-ngs/blob/ct-multialign/tools/snpeff.py#L60-L88) tomorrow, but the gist of it is that if a genome "name" is not in the default snpEff library, it creates a new one from the accessions specified, and names it based on a hash of the list of accessions (that way the containing directory name doesn't hit length limits in the case of many accessions, and it's easy to tell which entries are auto generated). The accessions are still represented as plaintext within snpEff.config, as the chromosome names:

    b1dc827a6915120fbe779eb20ada47cdb3f299f171f3d12a93e08c24cd845077.genome : JQ610675.1, JQ610676.1, JQ610677.1, JQ610678.1, JQ610679.1, JQ610680.1, JQ610681.1, JQ610682.1, JQ610683.1, JQ610684.1
        b1dc827a6915120fbe779eb20ada47cdb3f299f171f3d12a93e08c24cd845077.chromosomes : JQ610675.1, JQ610676.1, JQ610677.1, JQ610678.1, JQ610679.1, JQ610680.1, JQ610681.1, JQ610682.1, JQ610683.1, JQ610684.1

tomkinsc:
My impression was that the {} format can still be used lazily, if the logging formatter is changed to support it. By default the logging substitution expects %s tokens, but maybe pylint expects the logging formatter to be changed to {}. Either is still preferable to "{}".format() which is definitely not lazy.

PR #154, opened 2015-09-28
Ct update mosaik

tomkinsc:
Since Google Code is deprecated, development of MOSAIK seems to have
shifted to [GitHub](https://github.com/wanpinglee/MOSAIK). Old versions of MOSAIK were not building on Ubuntu
15.04, while the new ones from GitHub do build. The MOSAIK tool wrapper
has been updated to download from GitHub. Since the MOSAIK author is
not using tagged releases, the tool download is pinned to a commit hash rather than a version number.

The new version of MOSAIK has a few bug fixes and improvements compared to the version we were using previously (2.1.33).

### DIFF
--- a/tools/mosaik.py
+++ b/tools/mosaik.py
@@ -9,8 +9,8 @@ import subprocess
 import tools
 import util.file
 
-tool_version = '2.1.33'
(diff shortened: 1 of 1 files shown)


dpark01:
Yes! I love when old workarounds can get removed.

PR #155, opened 2015-09-30
Ct easy deploy

tomkinsc:
This adds files to easily create a virtualized environment for running viral-ngs. The environment can be either local (via VirtualBox), or cloud-based (via Amazon EC2). It is tested to work on host machines based on OS X 10.10 (Yosemite) and Ubuntu 15.04 (Vivid Vervet). Currently the pipeline can execute within the VM up to the point where the depletion databases are needed. Per the Vagrantfile, local VM RAM usage is set to 8GB. On EC2 it currently uses an m4.2xlarge instance with 32GB of RAM and 8 vCPUs.

The `run.sh` script sets up everything, and drops the user into a shell on the VM. 

To start, the script `run.sh` installs the necessary dependencies on the user's machine (ansible, vagrant, virtualbox, and virtualbox-aws). The provisioning is handled by Ansible, with Vagrant handling creation of the VMs and EC2 instances. On OSX it depends on Homebrew, and will install it if it is not present.  It depends on having apt on linux, so it's Debian/Ubuntu only for now. Ruby >=2.0 is required for [vagrant-aws](https://github.com/mitchellh/vagrant-aws), so versions of Ubuntu older than 15.04 (notably 14.04 LTS) would need to have ruby >=2.0 installed and made default (this is currently left as an exercise to the user).

Ansible clones the master branch of viral-ngs from GitHub, creates a Python 3 virtual environment, and installs the viral-ngs Python dependencies. The viral-ngs tool unit tests are run to download, install, and build all of the viral-ngs tools. A `Snakefile` for viral-ngs is copied to the home directory of the VM (locally: `/home/vagrant/`, on EC2: `/home/ubuntu/`), along with the associated `config.json` file. Files to contain sample names (`sample-depletion.txt`, etc.) are also created. A directory is created within the VM, `~/data/`, to store data to be processed. This directory on the VM is synced to the `./data/` directory on the host machine, relative to the location of the `east-deploy/Vagrantfile`. On local VMs, syncing of the directory is two-way and fast. On EC2 instances, the syncing is currently one way (local->EC2) due to Vagrant limitations. 

AWS credentials and SSH keypairs are passed in as environment variables, and the `run.sh` will prompt for the values if the environment variables are not set (though the values given interactively are ephemeral).

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -55,3 +55,5 @@ test/input/TestVPhaser2/in.bam.bti
 easy-deploy/data/Snakefile
 
 easy-deploy/data/config.json
+
+easy-deploy/.vagrant/
--- a/.gitignore
+++ b/.gitignore
@@ -51,3 +51,7 @@ nosetests.xml
 coverage.xml
 
 test/input/TestVPhaser2/in.bam.bti
+
+easy-deploy/data/Snakefile
+
+easy-deploy/data/config.json
--- a/docs/install.rst
+++ b/docs/install.rst
@@ -2,8 +2,11 @@ Installation
 ============
 
 
-System dependencies
+Manual Installation
 -------------------
+
+System dependencies
+~~~~~~~~~~~~~~~~~~~
 
 This is known to install cleanly on most modern Linux systems with Python,
 Java, and some basic development libraries.  On Ubuntu 14.04 LTS, the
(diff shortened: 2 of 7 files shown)


dpark01:
Nice ASCII art. This adds a great new feature, though I'm not currently well versed enough to evaluate it thoroughly. Since it doesn't break anything else, I figure that can be done at a later time. I see there was a slight fix made to the RTD docs--should we add a new subsection to the RTD with instructions on using the easy deploy mechanism as an alternative to the more manual approach?

tomkinsc:
Absolutely! I'll add a section to the docs.

PR #156, opened 2015-10-01
Ct landscape improvements

tomkinsc:


dpark01:
Generally looks good, and all tests pass. The one thing I'm not totally sure of is how well tested our wrappers around pysam's Tabix/VCF stuff are. If you've looked into it and are pretty sure all that still works (it probably does--the integration tests would probably catch issues there), then that's the only thing I'd want to double-check.

tomkinsc:
That was my assumption about the integration tests, but I'll manually double check that commands hitting the TabixFile-derived classes are working.

tomkinsc:
The Tabix/VCF wrappers seem to be working consistently with how they were before in master, and inspection of VcfReader instance attributes shows that everything seems to be the same (VcfReader has its parser set correctly to asVCF, for example, rather than the pysam default of "asTuple").

PR #164, opened 2015-11-12
refactor config file, add LSF data_dir check

tomkinsc:
This PR makes a few breaking changes to the template Snakemake config file. 
**Variable names will need to be updated in any derivative config files:** 

* `bmTaggerDbDir` -> `bm_tagger_db_dir`
* `bmTaggerDbs_remove` -> `bm_tagger_dbs_remove`
* `blastDbDir` -> `blast_db_dir`
* `blastDb_remove` -> `blast_db_remove`
* `trim_clipDb` -> `trim_clip_db`
* `spikeinsDb` -> `spikeins_db`
* `dataDir` -> `data_dir`
* `tmpDir` -> `tmp_dir`
* `logDir` -> `log_dir`
* `reportsDir` -> `reports_dir`
* `binDir` -> `bin_dir`
* `venvDir` -> `venv_dir`
* `lastal_ref_db` -> `lastal_ref_db_dir`

The following changes have also been made:

* `number_of_chromosomes` is now set based on the number of accessions specified for the reference genome, rather than an integer given in the config file
* the reference file prefix is now hard-coded to "`reference`" rather than the `ref_genome_file_prefix` value given in the config file. 
* a check has also been added to the LSF cluster submitter to run a pre-exec command to check whether `ls` can be called on the `data_dir`.

### DIFF
--- a/assembly.py
+++ b/assembly.py
@@ -127,7 +127,7 @@ def parser_trim_rmdup_subsamp(parser=argparse.ArgumentParser()):
                         default=100000,
                         type=int,
                         help='Subsample reads to no more than this many pairs. (default %(default)s)')
-    util.cmd.common_args(parser, (('loglevel', None), ('version', None), ('tmpDir', None)))
+    util.cmd.common_args(parser, (('loglevel', None), ('version', None), ('tmp_dir', None)))
     util.cmd.attach_main(parser, trim_rmdup_subsamp_reads, split_args=True)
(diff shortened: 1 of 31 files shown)


dpark01:
Seems good to me, though the one thing I would do is have the config variables be called bmtagger_* instead of bm_tagger_* but otherwise, everything else seems appropriate.

PR #167, opened 2015-11-18
Add some metagenomic tools/install

yesimon:
* Kraken
* MEGAN
* DIAMOND

Unit tests are more like simple integration tests since we we're not
making too many assumptions on the stability of output.


tomkinsc:
Great! 👍
Looking forward to digging in to what you've added.

yesimon:
What's the process for installing external dependencies on the travis container cache?

dpark01:
The container is really just a cache of pip installed packages and installed Tools (and it's actually disabled for tagged releases and certain other scenarios to force a full rebuild).

If you need to introduce system dependencies, use apt-get install--the [Travis environment](http://docs.travis-ci.com/user/ci-environment/) is based on Ubuntu 12.04. The proper place to do it would be in travis/before_install.sh (which doesn't currently have any such commands, but it's where they'd go).

If you're adding some kind of dependency, it should probably also go in the Vagrant/Ansible stuff in easy_deploy. @tomkinsc can talk you through that.

yesimon:
Yeah MEGAN is an out of band install process that is also >100Mb download so even a travis cache would probably be unreasonable. I'll turn it off at least for unit tests.

dpark01:
If part of the reason that it's out of band (other than the size) is the licensing issues, then it's kind of a pain, but you can follow the [example for Novoalign and GATK](https://github.com/broadinstitute/viral-ngs/blob/master/travis/install-tools.sh), where you create an encrypted tarball of an installed version of MEGAN (encrypted with Travis's public key) and use that (and then force users to install it on their own and provide the path in an environment variable). But if it's a large file, that's becomes a pain too.

I think you can skip the MEGAN tests for now since you haven't yet added downstream methods that rely on it.

yesimon:
I've updated it a bit to work with the tests a bit better.

I'm adding the dependency on `python-future` just so we can start using some py3 constructs (in this PR I'm using `super()`. I also changed the signature of mixed `*args` and named kwargs in `kraken.py` just to make it py2 compatible.

I've also added a `@skip_install_test` decorator so we can skip even the MEGAN install process so it would be compatible with the current installation process of running the tool installation test. I've also made MOSAIK skip install on macs because it requires a specialized install process.

yesimon:
I'm not sure why the DIAMOND install is failing. It might be a cmake problem on travis VMs.

tomkinsc:
With @dpark01's branch with [MUMmer](http://mummer.sourceforge.net/), it looks like we may be able to remove MOSAIK soon anyway.

At least the 2.7 test may be failing since subprocess lacks DEVNULL on Py2. May be able to do this:

```
try:
    from subprocess import DEVNULL  # py3
except ImportError:
    DEVNULL = open(os.devnull, 'wb')
```

yesimon:
Finally got the PR fixed by skipping special casing the diamond install for `$TRAVIS` and pulling in chained dependencies when installing kraken, and `$PATH` manipulation when running kraken commands to ensure jellyfish can be found correctly.

The original problem was that the travis test environment is quite old - Ubuntu 12.04, with no plans yet for updating to a newer container OS. The problem was gcc-4.6, the default on Ubuntu 12.04, could not compile diamond and is quite old. Therefore we can install a newer gcc like gcc-4.9, but we cannot set it as default because we have no access to the `update-alternatives` system without root in containerized travis. On the other hand, calling a specific gcc version in our code might not work because users' may not have the specific gcc version, so therefore we just call `gcc` for users and check for travis to set `gcc-4.9`.

It might be ideal to get conda installs up sooner rather than later for some of these trickier situations because the travis platform probably is not representative of a majority of end-users' environments, and binary packages may be ideal providing that they are available for most reasonable environments.

tomkinsc:
It [looks like](https://docs.travis-ci.com/user/trusty-ci-environment/) TravisCI *is* in the process of updating their test environment to 14.04, though it is currently in beta. Testing under 14.04 is a matter of adding the following to the `.travis.yml` file:

```yaml
sudo: required
dist: trusty
```
I'll start a branch to see if there are issues with building and testing our existing tools under 14.04.

yesimon:
Can you also work on a python 3.5.0 build? I tried adding it to the `.travis.yml` but I think it requires some prework on the travis container side. Documenting that would also be great!

dpark01:
Does the trusty beta for Travis allow caching?

On Mon, Nov 23, 2015 at 2:46 PM, Simon Ye <notifications@github.com> wrote:

> Can you also work on a python 3.5.0 build? I tried adding it to the
> .travis.yml but I think it requires some prework on the travis container
> side. Documenting that would also be great!
>
> —
> Reply to this email directly or view it on GitHub
> <https://github.com/broadinstitute/viral-ngs/pull/167#issuecomment-159042216>
> .
>



-- 
Daniel J. Park, PhD
Tel: +1-617-714-8526
dpark@broadinstitute.org

tomkinsc:
The Trusty beta [does **not**](https://docs.travis-ci.com/user/ci-environment/#Virtualization-environments) allow caching on public repos, and boot times are 20-52s vs 1-6s for containerized environments. Guessing that's a deal-breaker? As consolation it does have more RAM than the container-based 12.04 test environments (7.5GB vs. 4GB). We could entertain the idea of switching to a different hosted [CI provider](https://en.wikipedia.org/wiki/Comparison_of_continuous_integration_software).

@dpark01, what are your thoughts on having a python 3.5 build?

dpark01:
The boot time doesn't concern me as much as the overall build time for
small branch commits. We could either switch to a docker-based system
(which sounds like it's supported in the trusty travis environment?) or
bioconda to reduce build times (I actually used miniconda for this prior to
Travis's caching mechanism) and also to reduce network-related random build
errors (ie, intermittent failure to connect to random url to download yet
another Tool). Let's discuss further next week.

Python 3.5? Sure, sounds great. I assume we'd initially just test all three
environments (2.7, 3.4, 3.5) and maybe let go of 3.4 at some future time.
Sounds like a one-liner change? Keep this in a separate branch, it should
be much simpler to implement.


On Mon, Nov 23, 2015 at 3:40 PM, Chris Tomkins-Tinch <
notifications@github.com> wrote:

> The Trusty beta does *not*
> <https://docs.travis-ci.com/user/ci-environment/#Virtualization-environments>
> allow caching on public repos, and boot times are 20-52s vs 1-6s for
> containerized environments. Guessing that's a deal-breaker? As consolation
> it does have more RAM than the container-based 12.04 test environments
> (7.5GB vs. 4GB). We could entertain the idea of switching to a different
> hosted CI provider
> <https://en.wikipedia.org/wiki/Comparison_of_continuous_integration_software>
> .
>
> @dpark01 <https://github.com/dpark01>, what are your thoughts on having a
> python 3.5 build?
>
> —
> Reply to this email directly or view it on GitHub
> <https://github.com/broadinstitute/viral-ngs/pull/167#issuecomment-159058059>
> .
>



-- 
Daniel J. Park, PhD
Tel: +1-617-714-8526
dpark@broadinstitute.org

tomkinsc:
Let's talk through it next week and see if makes more sense to use Docker hub or anaconda.org to store pre-built dependencies. Ideally the system would also work beyond Travis for easier install by users.

For Docker, it might look something like this:
A Dockerfile to create an Ubuntu 15.04 environment (on top of Travis's 15.04 environment, but at least the Docker image would be redistributable). The Dockerfile could reuse the easy-install ansible playbook to install dependencies in the Docker container (passing in variables as needed via `ansible-playbook ansible-playbook.yml --extra-vars "user=ubuntu"`), using GATK and Novoalign from a host directory shared to a mount point within the Docker container (rather than copying in GATK and Novoalign as easy-install is doing now). This would give us a Docker image with viral-ngs, pre-built tools, and a pass-through to GATK, etc. on the host system. For testing on Travis, a shell script can call install-tools.sh to download GATK, Novoalign, etc. as we are doing now via the encrypted blob, storing them on the Travis host OS in a path to be shared with the Docker image. The script could then pull the pre-built viral-ngs image from Dockerhub, git pull or copy in the latest changes to viral-ngs, and run tests. Merges to master could trigger Dockerhub to rebuild the image via post-commit webhook.

Miniconda would be more simple, if we can encapsulate our dependencies. Why did you move away from conda previously?

As an aside, Docker on Travis [uses](https://blog.travis-ci.com/2015-08-19-using-docker-on-travis-ci/) Ubuntu 15.04 as the host OS.

yesimon:
Ok this looks good for merge

tomkinsc:
Are there dependencies that should be added to the [ansible playbook](https://github.com/broadinstitute/viral-ngs/blob/master/easy-deploy/ansible-playbook.yml)?

yesimon:
I've updated the ansible playbook, but also carried out some unrelated ansible changes.

Main difference is to vagrant sync the local git repo of viral-ngs instead of cloning the latest version from github. This is so that we can actually test our playbook, since I expect most uses of viral-ngs easy-deploy to be from within a local git checkout.

Also change the username variable in the playbook to `ansible_ssh_user` so that we don't need to pass in `--extra-vars=user=vagrant` when running the ansible playbook command from the host.

tomkinsc:
Great, thanks.

The reason I had Ansible cloning from github is because the local host copy of viral-ngs may have different OS/architecture/header needs than the VM container OS, and the tools downloaded or built for the host by our wrappers may not run within the container. If/when we switch to a conda-based install, we can probably use a synced local copy of viral-ngs. I'm currently adding more recipes to [bioconda](https://github.com/bioconda/bioconda-recipes), so hopefully the shift can be made in the near future.

The change to `ansible_ssh_user` (and `ansible_user ` in 2.0) is a nice one.

yesimon:
In that case the best approach is probably to do a `git archive` off of the synced git repo and work off the exported copy, cached by the commit hash so we don't do an archive every playbook run. Therefore you have two independent directories working off the same local git checkout, and gets updated every playbook run. As long as we don't add git submodules this will be fine.

@tomkinsc Why are the travis builds failing?

tomkinsc:
That makes sense. Since this PR is already moving to a synced copy, could you add the `git archive` changes as well so everything will work once this merged?

It should be something like this from the easy-deploy path, right?

`[[ -d ./data/viral-ngs/$(git rev-parse --verify HEAD)/ ]] || rm -r ./data/viral-ngs/* && git archive --format=tar --prefix=$(git rev-parse --verify HEAD)/ HEAD | (cd ./data/viral-ngs/ && tar xf -)`

It looks like others have [considered](https://github.com/Kentzo/git-archive-all) the submodule issue before.

Not sure why the builds are failing. Possibly related to @dpark01 [clearing](https://github.com/broadinstitute/viral-ngs/pull/174#discussion_r46236766) the Travis cache? I've manually restarted the tests to see if it was a one-off issue.

tomkinsc:
It [looks like](https://docs.travis-ci.com/user/pull-requests#Security-Restrictions-when-testing-Pull-Requests) encrypted environment variables are **not** available on Travis for pull requests from forks. @yesimon, you have write access to this repo now. Could you resubmit this PR from a branch on this repo, rather than your fork?

yesimon:
Wouldn't it be a problem if future PRs from forks couldn't be travis tested? This was working for me as of last week - what's changed?

dpark01:
I think it hasn't been a common scenario for us in the past. I think we've only had two forks that have ever PR'ed back into the project.

One workaround we could consider is to see if we can/should skip the unit tests that require Novoalign/GATK in the event that the encrypted environment variables aren't set.

But I think the reason it worked last week is that the build cache on master is actually a decent workaround. Since PR's are usually going into master (and Travis uses the target branch's cache if it exists), if master has a Travis cache, then Novoalign/GATK are already there and can be utilized. But last night I wiped the Travis caches just to be sure about some Trinity stuff on a separate PR, so that's why it's failing for you now.

I bet if we rebuild master then rebuild your PR, it'll work again?

tomkinsc:
Yeah, it looks like it's because the cache was cleared. The Travis logs show that there was a successful [test](https://travis-ci.org/broadinstitute/viral-ngs/jobs/94044682#L533) of this PR, yet BUNDLE_SECRET was not set. `tools/install-tools.sh` was called, but succeeded [because](https://github.com/broadinstitute/viral-ngs/blob/master/travis/install-tools.sh#L4) GATK_PATH and NOVOALIGN_PATH both existed in the cache. Once the cache is rebuilt the tests should hopefully pass again.

yesimon:
Okay good to know, we might want to add that to the development docs because this side effect is highly desired (want to test pull requests in general)

The `git archive` method does not make as much sense as I was originally thinking. The main issue is whether to blow away the project directory each time. If we blow away the project directory every time the commit hash changes, that essentially means we need a full tools re-install/rebuild each time we make a change, which is time consuming. If we don't blow away the project directory, then over time as we unzip the latest project tree into the directory we could end up with garbage files that one day will cause strange bugs, but we won't need to worry about tools reinstall.

Probably the best option is to add an `extra_var` that we can pass in on the command line to `ansible-playbook` to toggle between git archive clean builds and directly using the sync directory. Using the synced directory is for rapid iteration whereas the the git archive is for a clean test of local changes.

Finally, the last problem I'm getting with `git archive` is the following:
```
TASK: [Install viral-ngs tools] *********************************************** 
failed: [default] => {"changed": true, "cmd": ["/home/vagrant/venv/exec", "python", "-m", "unittest", "test.unit.test_tools", "-v"], "delta": "0:00:00.051085", "end": "2015-12-01 19:43:09.031517", "rc": 1, "start": "2015-12-01 19:43:08.980432", "warnings": []}
stderr: fatal: Not a git repository (or any of the parent directories): .git
Traceback (most recent call last):
  File "/usr/lib/python3.4/runpy.py", line 170, in _run_module_as_main
    "__main__", mod_spec)
  File "/usr/lib/python3.4/runpy.py", line 85, in _run_code
    exec(code, run_globals)
  File "/usr/lib/python3.4/unittest/__main__.py", line 18, in <module>
    main(module=None)
  File "/usr/lib/python3.4/unittest/main.py", line 92, in __init__
    self.parseArgs(argv)
  File "/usr/lib/python3.4/unittest/main.py", line 139, in parseArgs
    self.createTests()
  File "/usr/lib/python3.4/unittest/main.py", line 146, in createTests
    self.module)
  File "/usr/lib/python3.4/unittest/loader.py", line 146, in loadTestsFromNames
    suites = [self.loadTestsFromName(name, module) for name in names]
  File "/usr/lib/python3.4/unittest/loader.py", line 146, in <listcomp>
    suites = [self.loadTestsFromName(name, module) for name in names]
  File "/usr/lib/python3.4/unittest/loader.py", line 105, in loadTestsFromName
    module = __import__('.'.join(parts_copy))
  File "/home/vagrant/bin/test/__init__.py", line 8, in <module>
    import util.file
  File "/home/vagrant/bin/util/file.py", line 17, in <module>
    import util.cmd
  File "/home/vagrant/bin/util/cmd.py", line 16, in <module>
    __version__ = util.version.get_version()
  File "/home/vagrant/bin/util/version.py", line 71, in get_version
    raise ValueError("Cannot find the version number!")
ValueError: Cannot find the version number!

FATAL: all hosts have already failed -- aborting
```

I'm not sure where the `VERSION` file gets set but this makes it unfunctional when the project is not a git checkout. 

Since we're dealing with a bunch of issues, let's split those out and get the tools related things checked in first.

dpark01:
Hm... what's wrong with "git clone" again?

That's a good point--the dumb hacky mechanism that I put together to
extract a version name for viral-ngs (which I feel is very important for
debugging purposes--every command line call starts off by spitting that out
to stdout) relies on either .git/ existing or on a packaging / install
mechanism making sure that the VERSION file exists (if .git/ exists, then
the first call to any of the python scripts will create the VERSION file
for you).

So "git archive" obviously doesn't work with that... unless your vagrant
code that calls "git archive" also calls "git describe --tags --always
--dirty > VERSION" as well. Then it should be fine.

Danny

On Tue, Dec 1, 2015 at 3:47 PM, Simon Ye <notifications@github.com> wrote:

> Okay good to know, we might want to add that to the development docs
> because this side effect is highly desired (want to test pull requests in
> general)
>
> The git archive method does not make as much sense as I was originally
> thinking. The main issue is whether to blow away the project directory each
> time. If we blow away the project directory every time the commit hash
> changes, that essentially means we need a full tools re-install/rebuild
> each time we make a change, which is time consuming. If we don't blow away
> the project directory, then over time as we unzip the latest project tree
> into the directory we could end up with garbage files that one day will
> cause strange bugs, but we won't need to worry about tools reinstall.
>
> Probably the best option is to add an extra_var that we can pass in on
> the command line to ansible-playbook to toggle between git archive clean
> builds and directly using the sync directory. Using the synced directory is
> for rapid iteration whereas the the git archive is for a clean test of
> local changes.
>
> Finally, the last problem I'm getting with git archive is the following:
>
> TASK: [Install viral-ngs tools] ***********************************************
> failed: [default] => {"changed": true, "cmd": ["/home/vagrant/venv/exec", "python", "-m", "unittest", "test.unit.test_tools", "-v"], "delta": "0:00:00.051085", "end": "2015-12-01 19:43:09.031517", "rc": 1, "start": "2015-12-01 19:43:08.980432", "warnings": []}
> stderr: fatal: Not a git repository (or any of the parent directories): .git
> Traceback (most recent call last):
>   File "/usr/lib/python3.4/runpy.py", line 170, in _run_module_as_main
>     "__main__", mod_spec)
>   File "/usr/lib/python3.4/runpy.py", line 85, in _run_code
>     exec(code, run_globals)
>   File "/usr/lib/python3.4/unittest/__main__.py", line 18, in <module>
>     main(module=None)
>   File "/usr/lib/python3.4/unittest/main.py", line 92, in __init__
>     self.parseArgs(argv)
>   File "/usr/lib/python3.4/unittest/main.py", line 139, in parseArgs
>     self.createTests()
>   File "/usr/lib/python3.4/unittest/main.py", line 146, in createTests
>     self.module)
>   File "/usr/lib/python3.4/unittest/loader.py", line 146, in loadTestsFromNames
>     suites = [self.loadTestsFromName(name, module) for name in names]
>   File "/usr/lib/python3.4/unittest/loader.py", line 146, in <listcomp>
>     suites = [self.loadTestsFromName(name, module) for name in names]
>   File "/usr/lib/python3.4/unittest/loader.py", line 105, in loadTestsFromName
>     module = __import__('.'.join(parts_copy))
>   File "/home/vagrant/bin/test/__init__.py", line 8, in <module>
>     import util.file
>   File "/home/vagrant/bin/util/file.py", line 17, in <module>
>     import util.cmd
>   File "/home/vagrant/bin/util/cmd.py", line 16, in <module>
>     __version__ = util.version.get_version()
>   File "/home/vagrant/bin/util/version.py", line 71, in get_version
>     raise ValueError("Cannot find the version number!")
> ValueError: Cannot find the version number!
>
> FATAL: all hosts have already failed -- aborting
>
> I'm not sure where the VERSION file gets set but this makes it
> unfunctional when the project is not a git checkout.
>
> Since we're dealing with a bunch of issues, let's split those out and get
> the tools related things checked in first.
>
> —
> Reply to this email directly or view it on GitHub
> <https://github.com/broadinstitute/viral-ngs/pull/167#issuecomment-161090536>
> .
>



-- 
Daniel J. Park, PhD
Tel: +1-617-714-8526
dpark@broadinstitute.org

yesimon:
We can merge this now. I'll set `VERSION` in the git archive PR manually since that makes sense.

tomkinsc:
It's nice to see the Ansible changes are now in a separate pull request.

PR #169, opened 2015-11-23
Add tox configuration and corresponding docs.

yesimon:
Tox env set to build docs as treating warnings as errors, so fix some docs warnings as well.

### DIFF
--- a/docs/conf.py
+++ b/docs/conf.py
@@ -23,7 +23,8 @@ sys.path.insert(0, os.path.dirname(os.path.abspath('.')))
 
 # -- Mock out the heavyweight pip packages, esp those that require C ----
 import mock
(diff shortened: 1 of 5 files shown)


docs/conf.py
```diff
@@ -116,7 +117,7 @@ def _git_version():
 
 # The theme to use for HTML and HTML Help pages.  See the documentation for
 # a list of builtin themes.
-html_theme = 'default'
+# html_theme = 'default'
```
dpark01: wait, why change the html_theme? I think ReadTheDocs gets all ugly if it isn't default

yesimon: Sphinx actually gave a warning that default is being renamed to classic, and that alabaster is the new recommended? (Right now it still hasn't converted but there are warnings).

As far as how RTD works, http://read-the-docs.readthedocs.org/en/latest/theme.html, the docs say that the RTD theme will apply when not specified.

dpark01: Ok, sounds great

PR #174, opened 2015-11-30
a few landscape/pylint-inspired changes; replaced separate noBlastHits_v3.py with function in taxon_filter.py

tomkinsc:
a few landscape/pylint-inspired changes; replaced separate noBlastHits_v3.py with function in taxon_filter.py


intrahost.py
```diff
@@ -1109,13 +1109,13 @@ def sampleIDMatch(inputString):
...
     """
-    idRegex = re.compile("(.*?)(?:-\d+|$)+")
+    idRegex = re.compile(r"(.*?)(?:-\d+|$)+")
     m = idRegex.match(inputString)
 
     if m:
         return m.group(1)
     else:
-        raise LookupError("The ID was not of the form (.*?)(?:-\d+|$)+, ex. 5985-0")
+        raise LookupError(r"The ID was not of the form (.*?)(?:-\d+|$)+, ex. 5985-0")
```
dpark01: This seems like an incorrect place to put an r?

pipes/Broad_LSF/cluster-submitter.py
```diff
@@ -45,7 +45,7 @@
 cmdline += " %s/%s.jobfinished" % (sm_tmpdir, sm_jobid)
 
 # the part that strips bsub's output to just the job id
-cmdline += " | tail -1 | cut -f 2 -d \< | cut -f 1 -d \>"
+cmdline += r" | tail -1 | cut -f 2 -d \< | cut -f 1 -d \>"
```
dpark01: this also seems wrong

pipes/Broad_UGER/cluster-submitter.py
```diff
@@ -44,7 +44,7 @@
 cmdline += " %s/%s.jobfinished" % (sm_tmpdir, sm_jobid)
 
 # the part that strips bsub's output to just the job id
-cmdline += " | tail -1 | cut -f 3 -d \ "
+cmdline += r" | tail -1 | cut -f 3 -d \ "
```
dpark01: another

tools/trinity.py
```diff
@@ -68,7 +68,7 @@ def post_download(self):
                 with open(badFilePath + '.orig', 'rt') as inf:
                     for line in inf:
                         if line.startswith('unless ($java_version =~ /java version'):
-                            outf.write('$java_version =~ /java version \"1\.(\d+)\./;\n')
+                            outf.write(r'$java_version =~ /java version \"1\.(\d+)\./;\n')
```
dpark01: incorrect

intrahost.py
tomkinsc: Prefixing with an 'r' to indicate a [raw string literal](https://docs.python.org/2.0/ref/strings.html) seemed better than properly escaping the backslash (this is pylint-inspired).

tools/trinity.py
dpark01: Ok I've convinced myself that the above three raw strings are fine, but this fourth one in trinity.py still worries me a bit. I've manually deleted the cache container in Travis and re-started the tests on this PR just to give it a full test battery, because I'm pretty sure these particular lines of code don't get tested if Trinity is already present in the build dir's cache.

tomkinsc: Sounds good, thank you.

dpark01:
Okay I think I feel more comfortable about Trinity now. Let's see how the tests go and then we can merge it if all looks green.

dpark01:
I'm slightly worried that Trinity is failing because the Chrysalis program takes more RAM than Travis gives it.. some more exploration required...

dpark01:
merge when green!

PR #181, opened 2015-12-01
Make ansible vagrant easier to test changes

yesimon:
Previously easy_deploy relied on git clone of repo from github, which
makes playbook changes hard to test without committing to the master
repo.

Now, we vagrant sync the local git repo into the VM, then perform a git
archive export to the final project directory so tool installation in VM
does not affect the host machine.

This is forked off from #167

### DIFF
--- a/docs/development.rst
+++ b/docs/development.rst
@@ -13,24 +13,73 @@ testing, you first need to perform the following steps:
 
 ::
 
-    pyenv install 2.7.10 3.4.3 3.5.0
+    $ pyenv install 2.7.10 3.4.3 3.5.0
 
 3. Change directory to the `viral-ngs` git checkout.
 4. Set the local pyenv versions:
 
 ::
 
-    pyenv local 3.5.0 3.4.3 2.7.10
+    $ pyenv local 3.5.0 3.4.3 2.7.10
 
 5. Install tox and tox-pyenv
 
 ::
 
-    pip3 install tox tox-pyenv
+    $ pip3 install tox tox-pyenv
 
 6. Run tox to run tests and check building docs.
 
 ::
 
-    tox
+    $ tox
 
+Testing easy_deploy with Vagrant and Ansible
+--------------------------------------------
+
+By default, easy_deploy installs a version of viral-ngs by cloning the github
+repository, which makes it difficult to test local changes to the playbook.
+Testing the easy_deploy setup locally can be done using Vagrant and the Ansible
+playbook with some custom commands. To do so, we need to take the following
+steps:
+
+1. `Install Vagrant <https://docs.vagrantup.com/v2/installation/>`_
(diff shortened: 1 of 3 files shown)


tomkinsc:
It would be project-specific and somewhat ugly, but as an alternative to blowing away the project directory every time the hash changes we could check to see if there have been any changes to `tools/` since all tool wrappers, binaries, etc. are contained within.

```
[[ $(git diff -p HEAD tools/ | wc -l) -gt 0 ]] && rm [...]
```

tomkinsc:
For actual testing of the playbook, maybe the Vagrantfile should add a Docker provider, so we can test the playbook on Travis without the performance penalties imposed by several layers of virtual machines (if Travis even allows VMs). We'd have to install Docker on the Travis Ubuntu host, and could probably skip the test on the OSX test host.

yesimon:
I think the most logical is actually to test the playbook locally, but that would require sudo (ie just provision the travis instance). Almost all realistic travis tests of the playbook would involve moving easy-deploy into a separate repository unless we're willing to give up containerized builds.

Trying to get it to kinda work by testing playbook versus docker would simply be too confusing, especially since we're committed to getting local conda installs to work. Having a docker image - either prebuilt - or built via a Dockerfile on travis would just be another thing to maintain.

tomkinsc:
I still like that having `git clone` in the playbook means that the host copy of viral-ngs running doesn't have to be a git repo checkout. The playbook should perhaps be two though, one to install Ansible in the instance and clone viral-ngs, and then a second from the cloned viral-ngs to set up the rest of the environment locally within the instance.

With conda install of packages, moving away from containerized builds is not such a problem since anaconda.org would cache the binaries. Network throughput to Travis could be a big bottleneck, but that remains to be seen.

Testing the ansible playbook locally on Travis by provisioning the test instance doesn't really make sense to me, since the Travis test environment is not necessarily reflective of the environment we want to target for running viral-ngs (we should make Travis more like the target, not vice-versa). Docker would give us more flexibility there, and the implementation wouldn't be so bad; the approach seems to be working well for bioconda. We wouldn't need to maintain a Docker image at all, though we could if we care about caching built tools. It would also give the project a Docker image, which might be nice for those who like to run things that way.

It's seductive to have automated testing for the vagrantfile and the playbook, but I'm not sure it's the best use of time right now. The playbook has mostly system dependencies, and it's unlikely to see many changes, especially when things installed via conda can install their own binary dependencies.

yesimon:
I'll change it so that `git clone` is the default method. Then we'll have two options - one for `git archive` from local checkout, and the other for direct synced viral-ngs repo: both of these to facilitate local testing of the playbook which is the scope of this commit.

We can punt on travis testing of easy_deploy until we have a better idea of our requirements.

yesimon:
I have also added docs to make these options clearer

PR #189, opened 2015-12-15
condafy tools

tomkinsc:
This PR adds an `InstallMethod()` class to `tools/__init__.py` to install tools sourced from the bioconda channel. It also prepends an instance of this method to each tool for which there is currently a conda package available.

The following tools **do not** have conda packages available:
- Tbl2AsnTool (source not published, binaries link against version of GLIBC too new to build via the bioconda test container)

Kraken (in)/Diamond (pending)/Yaggo(in)/Jellyfish(pending) are currently in the process of being merged in to bioconda, but their wrappers are ready.

Novoalign uses a single-threaded conda install if the NOVOALIGN_PATH is not set, otherwise it uses the multithreaded licensed install.

Some of the series of Travis debug commits are fixes, others are futile attempts at working around what turned out to be a Travis cache issue.

Not all conda packages are currently supported on OSX (notably bmTagger).


tomkinsc:
Cache issue may still be present...

yesimon:
Is there a standard formatting to use? Maybe like yapf so we can keep the sources relatively consistent?

tomkinsc:
I've been using yapf with these arguments: 

`yapf --verify --in-place --style='{based_on_style: google, column_limit=120 }' ./some_file.py`

yesimon:
Cool, we should add the style config to the project.

tomkinsc:
Sure, in #191.

dpark01:
If it's the Travis cache, it's weird that it would be just for Python 2.7 and not the others.... if I take the error message at face value, it implies that there's some Tool out there that does not have a InstallMethod (or that it's set to None for some reason).

I actually really wish that testAllToolInstallers was not a single unit test. I think both unittest and nose have a way of defining a whole bunch of unit tests in code. This ought to actually iterate over the leaf subclasses and define a dozen or so separate test cases, one per subclass, so that they can fail and succeed independently, and so that it's much easier to interpret the failures. Anyone familiar enough with nose and know how to implement that quickly? =)

tomkinsc:
The error may be true, but it's not always true. I had several tests passing yesterday under 2.7, such as b9ff292. As a sanity check, I restarted them and they all failed (sanity: questioned). Since the code is the same, that led me to believe it was the cache, but perhaps it's not. I'm still working to trace it.

I'll start on breaking up `testAllToolInstallers()` into separate programatically-generated test cases. That should definitely help determine what's causing the problem. One approach would be to have the initializer for `TestToolsInstallation()` iterate over the leaf tool classes add test methods via `setattr ` to `TestToolsInstallation`, each with a "test_"-prefixed `__name__`.

dpark01:
I lean towards test generators. I'm willing to drop compatibility with unittest and just support nose.

tomkinsc:
How would you feel about switching to `nose2`? It's [mostly the same](https://nose2.readthedocs.org/en/latest/differences.html), with some expanded functionality via plugins. If we opt for `nose2` over `nose` and use the [`load_tests protocol` ](https://docs.python.org/2/library/unittest.html#load-tests-protocol) we can have parameterized tests and preserve the ability to test via `unittest`.

dpark01:
Sure... if an environment already has nose, can nose2 coexist with it?

Maybe we should add a requirements-tests.txt and change travis/install-pip.sh to just use that instead of naming packages by hand (coveralls, nose-cov).

tomkinsc:
Nose2 can coexist with nose, though nose2 does not yet have all of the same plugins as the original nose. It does appear to have what we are currently using, notably coverage. Nose2 does not come with a `nosetests` script, so the invokation would be different. What is our current use for the xunit-formatted XML?

Having a `requirements-tests.txt` makes sense. I'll migrate to that. If we are looking at changing test runners/frameworks there are others to consider. From what I've been reading, the python community seems to be leaning toward [pytest](http://pytest.org/latest/contents.html). Pytest can run both nose and unittest-style tests, can be parallelized, has [extra detail](https://fedoraproject.org/wiki/User:Tflink/AutoQA_nose_pytest_comparison#Output_For_Test_Failures) in its failure output, more flexible [output capturing](http://pytest.org/latest/capture.html#captures), supports simple assertions (`assert x == y`, rather than `assertEqual()`), and has decent docs. 

@dpark01, what do you think? @yesimon, care to weigh in?

yesimon:
Sure - I think that switching to pytest is a good idea, especially for more complex cases as in viral-ngs.

dpark01:
Skimming through the conda changes here, hopefully the samtools 0.1.19 -> 1.2 goes smoothly. (ie, I hope our test coverage would actually catch any problems that might arise... I feel like it would though).  I'm a bit more concerned about the Trinity version upgrade though--there were some specific algorithmic changes that happened between the older one and the newer one that kept us using the old version (this would not get caught by a unit test). Is it possible to push the older version into bioconda or else stick with the download-modify-and-compile tool installer for Trinity?

tomkinsc:
We should probably note the reason in `trinity.py` for pinning to an older version. I'll add bioconda recipes for older versions of trinity and samtools.

dpark01:
I don't know that an older version of samtools is necessary--if integration tests are passing, it's probably fine.

tomkinsc:
As of c2834bd, all tools are installing via Conda on Travis under Python 2 and 3, except for a few tools: 
* Diamond (recipe not yet in bioconda channel)
* Jellyfish (recipe not yet in bioconda channel)
* Tbl2Asn (links against newer glibc than bioconda test environment can handle, no source, no mac binary)

It looks like the Novoalign wrapper is not installing the tool via conda on Travis because the install methods give higher precedence to the licensed binary defined in env vars, since the unlicensed conda version is single-threaded. If the env is set (as it is on Travis), it will use the licensed version.

tools/__init__.py
```diff
@@ -19,14 +21,19 @@
...
+        if filename.endswith(
+            '.py') and filename != '__init__.py' and filename not in [    # Add any files to exclude here:
+    # e.g. 'sometool.py',
+            ]
+    ]
+)
 installed_tools = {}
 
-LOG = logging.getLogger(__name__)
+log = logging.getLogger(__name__)
```
dpark01: I'm a little curious / confused: at some point in the recent past, there was a sweeping change across the codebase to convert log -> LOG (not sure if this was you, @yesimon, or pylint), and here I see some, but not all, of them reverting back. I don't particularly care either way, so long as it's consistent (but if there's a compelling argument one way or another, let's do that one).

tomkinsc: The all-caps version was due to pylint asserting that anything in the global scope should be in all caps. Since the logger object is not a constant, the `LOG` should probably be `_log` per the PEP8 [guideline](https://www.python.org/dev/peps/pep-0008/#global-variable-names) for non-public module globals. I'll make that change in a separate PR, and revert the names in this PR to be all caps for consistency.

tomkinsc:
OK to merge, or should we wait for the Diamond and Jellyfish conda recipes?

dpark01:
I looked at it this morning and that comment was the only question I had. I'll merge now, let's not wait for upstream!

PR #195, opened 2016-01-14
improve conda install logging

tomkinsc:
in the event that conda installs result in a json decoding error,
installs now fall back to our wrappers, and log messages are emitted.


dpark01:
nice, I like this implementation

PR #203, opened 2016-02-16
fix to snpeff availabl_databases() for conda-based installs

tomkinsc:
the version of snpeff installed by conda relies on a shell script wrapper, and must be called differently than a standard jar file. this difference in invocation was present in snpeff.execute() but not snpeff.available_databases(). The latter now relies on snpeff.execute(). The execute() function itself has been changed to use
util.misc.run() rather than subprocess.check_call() so we can capture the output (it is read to get the list of available databases)

### DIFF
--- a/tools/snpeff.py
+++ b/tools/snpeff.py
@@ -59,7 +59,7 @@ class SnpEff(tools.Tool):
             ] + args
 
         _log.debug(' '.join(tool_cmd))
-        return util.misc.run(tool_cmd)
+        return util.misc.run(tool_cmd, stdin=stdin)
 
     def has_genome(self, genome):
         if not self.known_dbs:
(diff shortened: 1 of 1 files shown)


tools/snpeff.py
```diff
@@ -43,22 +44,22 @@ def version(self):
...
             ] + args
         else:
             tool_cmd = [
-                self.install_and_get_path(), '-Xmx' + JVMmemory, '-Djava.io.tmpdir=' + tempfile.tempdir, command
+                self.install_and_get_path(), '-Xmx' + JVMmemory, '-Djava.io.tmpdir=' + tempfile.gettempdir(), command
             ] + args
 
         _log.debug(' '.join(tool_cmd))
-        subprocess.check_call(tool_cmd, stdin=stdin, stdout=stdout)
+        return util.misc.run(tool_cmd)
```
dpark01: The end of annotate_vcf relies on being able to pass a file handle as the stdout named parameter, and it looks like this change may not accomodate that?

tomkinsc: Thanks, nice catch. I'll change `annotate_vcf()`.

dpark01:
any idea what's going on with landscape.io these days? anyway, this PR looks good

tomkinsc:
Not sure why landscape is failing. I pinged their support email a while ago back when we had a syntax error in our pylint dotfile, and there doesn't seem to be a way to get more verbose error reporting.

tomkinsc:
Looks like landscape may be having issues because it cannot install pysam.

PR #206, opened 2016-02-18
add conda install method to MUMmer tool

tomkinsc:


dpark01:
I was just looking around in the MUMmer INSTALL documentation and it unfortunately says this:

>    To make all of the scripts and executables accessible from different
directories, simply add the full MUMmer directory path to your system PATH, or
link the desired MUMmer programs to your favorite bin directory. Please note
that the 'make' command dynamically generates the MUMmer scripts with the
appropriate paths, therefore if the MUMmer directory is moved after the 'make'
command is issued, the scripts will fail to run. If the MUMmer executables
are needed in a directory other than the install directory, it is recommended
that the install directory be left untouched and its files linked to the
desired destination. An alternative would be to move the install directory and
reissue the 'make' command at the new location.

Hmm... sounds like MUMmer+Conda might be a little more work?

tomkinsc:
The conda recipe may need some adjustment to create symlinks, but at least part of the [problem](https://travis-ci.org/broadinstitute/viral-ngs/jobs/110181901#L930) appears to be in constructing the paths to the various mummer tools:

https://github.com/broadinstitute/viral-ngs/blob/dp-scaffold/tools/mummer.py#L48

For the conda tools, install_and_get_path() returns the path to the main mummer binary, whereas to get the path to promer, etc., we'd want the dirname of that full path. The existing tool wrappers like the one for [blast](https://github.com/broadinstitute/viral-ngs/blob/dp-scaffold/tools/blast.py), have separate tool classes for the blast subtools, each with a different subtool_name that then becomes the executable for the conda install method. Moving `promer`, `nucmer`, etc. to their own tool classes would mean some refactoring, but it would better validate their executability at install time.

tomkinsc:
There is a [PR with changes](https://github.com/bioconda/bioconda-recipes/pull/829) to the conda recipe for MUMmer that preserves the location of the MUMmer binaries within their build directory, and places symlinks to the binaries in the conda `bin/`. Hopefully conda will take care of fixing any incorrect paths as it moves things around from build/ to test/.

dpark01:
Interesting. Current problem might be a bug in the MUMmer bioconda package: while nucmer, mummer, promer, show-coords, show-diff, show-snps, show-tiling, and delta-filter all exist in the conda-tools/bin/ directory, it appears to be missing "show-aligns", which is needed by the scaffolding code I wrote.

tomkinsc:
That's a simple omission to fix, via [PR #832](https://github.com/bioconda/bioconda-recipes/pull/832). Not sure how I missed `show-aligns`...

PR #209, opened 2016-02-18
move dotkit sourcing in Broad_LSF/run-pipe.sh

tomkinsc:
pyyaml is imported by sourcing the anaconda dotkit. It's python3, so hopefully compatible?


dpark01:
wow, sneaky... so instead of using any of the Broad python dotkits, we're using the python that comes with the conda dotkit. Nice.

PR #217, opened 2016-02-26
Escape sample names

tomkinsc:
sample names in the sample sheet can contain characters that are
illegal in filenames, causing Picard to fail when trying to write bam
files. a sanitization function (util.file.string_to_file_name()) is now
applied to sample names in illumina.py, and a unit test has been added
in test_illumina.py


test/unit/test_illumina.py
```diff
@@ -150,6 +150,29 @@ def test_tarball_fail_missing_data(self):
...
+        illumina.miseq_fastq_to_bam(outBam, sampleSheet, fastq[0], inFastq2=fastq[1], runInfo=runInfo)
+        rgs = list(tools.samtools.SamtoolsTool().getReadGroups(outBam).values())
+        self.assertEqual(len(rgs), 1)
+        rgs = rgs[0]
+        self.assertEqual(rgs.get('ID'), 'AEF96')
+        self.assertEqual(rgs.get('PL'), 'illumina')
+        self.assertEqual(rgs.get('PU'), 'AEF96.1.CGTACTAG-CTAAGCCT')
+        # we care more about Picard not failing, but if we care about how filenames are processed:
+        #self.assertEqual(rgs.get('LB'), 'difficult-value+for_-sample_name0.1')
+        #self.assertEqual(rgs.get('SM'), 'difficult-value+for_-sample_name0.1')
```
dpark01: Any reason not to include these tests? At a minimum, it'd be nice to test that the fields exist and are non-empty.

util/file.py
```diff
@@ -339,3 +339,51 @@ def temp_catted_files(input_files, prefix=None, suffix=None):
...
+        r"?": "_", # could be mistaken for a wildcard
+        r"*": "_", # could be mistaken for a wildcard
+        r"`": "_", # no subshells
+        r" -": "_-", # could be mistaken for an argument
+        r" --": "_--", # could be mistaken for an argument
+        r">": "]", # no redirect chars
+        r"<": "[", # no redirect chars
+        r"\\x": "_", # hex char
+        r"\\o": "_", # octal char
+        r"\\u": "", # unicode char
```
dpark01: does this trigger on unicode, or just encoded-unicode? More specifically, would normal unicode (e.g. Sénégal) be accepted as a sample name? I don't see a particular reason to disallow it, especially if DNAnexus is using py3 environments.

test/unit/test_illumina.py
tomkinsc: I'll add them back in.

util/file.py
tomkinsc: Just encoded unicode. We could omit line 365; I'll comment it out for now. Under py3 unicode passes through the function unchanged, but now that I've added "Sénégal" and some other unicode to the test case I see there are failures elsewhere. I took a look at the bam file created by the unit test, and seems looks like samtools writes latin-1 but strips utf-8? I've specified the encoding in samtools.py getHeader() as "latin-1" and we can at least load "Sénégal" now. We should, in general, talk about our encoding expectations for viral-ngs...

PR #221, opened 2016-03-01
Escape sample names

tomkinsc:
previously only sample IDs (and the sample part of library) were
sanitized to be filename-compatible, now sample, library, and run are
all sanitized


dpark01:
looks good!

PR #230, opened 2016-03-09
Add lastal params

tomkinsc:
add parameters for lastal to specify
max_gapless_alignments_per_position, min_length_for_initial_matches,
max_length_for_initial_matches, and max_initial_matches_per_position.
The default values were set based on values determined by @haydenm
during experimentation with influenza data


dpark01:
The changes look good to me. Looking at the diffs, the fact that none of the test files changed, does that mean we don't have test coverage on the taxon filter / lastal step?

tomkinsc:
There is [some existing coverage](https://github.com/broadinstitute/viral-ngs/blob/master/test/unit/test_taxon_filter.py#L50). This PR really only provides @haydenm's new defaults to lastal, and a way to override them via argparse params. Should we have additional coverage, or expose more of the lastal binary's params?

dpark01:
Oh I guess it makes sense for a simple test scenario the change in default params shouldn't affect whether reads match or don't match the taxon of interest... We're not testing the edge cases. Ok, let's merge this when it passes.

PR #236, opened 2016-03-16
travis log capture and cleanup

tomkinsc:
Replace subprocess calls with custom calls that capture subprocess output and pipe to stdout (or files where appropriate), so nose can capture all output and only show stdout for failures. In order for nose to capture log messages, it is necessary to clear existing logging handlers via “—logging-clear-handlers” in the nosetests call. This should clean up the Travis logs

### DIFF
--- a/assembly.py
+++ b/assembly.py
@@ -25,6 +25,7 @@ except ImportError:
 # intra-module
 import util.cmd
 import util.file
+import util.misc
 import util.vcf
 import read_utils
 import taxon_filter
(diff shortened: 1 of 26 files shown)


dpark01:
I have a question about our util.misc.run_and_print method. Does it capture all output and buffer in memory? Previously, we were O(1) memory on the parsing of subprocess output because we iteratored through it. This is mostly potentially an issue for samtools / picard type stuff (stuff that parses very large outputs).

Can we make that method O(1) mem?

tomkinsc:
We should be able to replace Simon's py-3.5-like `run()` in util.misc with one based on `Popen()` that will direct `stdout` to `subprocess.PIPE`, which we can poll to write to `sys.stdout` in pipe-size chunks (4096 bytes). How does that sound? We could prevent that chunking by running Python with `-u` flag (unbuffered), but that's not ideal.

dpark01:
Sounds fine to me, right now it's buffering in chunks of infinity, so 4k sounds great!

PR #237, opened 2016-03-17
jobscript.sh fixes for UGER

dpark01:


pipes/Broad_UGER/jobscript.sh
```diff
@@ -2,11 +2,14 @@
 # properties = {properties}
 # this is identical to the default jobscript with the exception of the exit code
 
+source /broad/software/scripts/useuse
+use Python-3.4
```
tomkinsc: Can this be omitted if the dotkit setup script is sourced above the pyyaml-dependent python calls?

dpark01: except, how would we know where to find the dotkit setup script, since we need $BINDIR first?

tomkinsc: Right! Though could $BINDIR be relative, as it is in `run-pipe.sh`?

PR #239, opened 2016-03-17
add "-b y" flag to docs about UGER use

tomkinsc:


yesimon:
I just had an error where `-b y` doesn't work for scripts piped in via STDIN to `qsub`, despite what it says in the docs. Therefore one needs to actively set `-b y | -b n`. The inconsistency is remarkable.

Issue #241, opened 2016-03-23
VERSION file missing from v1.3.4.tar.gz?


sbecuwe:
Hello,

Trying to run e.g. intrahost.py, I get following error:

```
fatal: Not a git repository (or any of the parent directories): .git
Traceback (most recent call last):
  File "./intrahost.py", line 25, in <module>
    import util.cmd
  File "/dev/shm/viral-ngs-1.3.4/util/cmd.py", line 16, in <module>
    __version__ = util.version.get_version()
  File "/dev/shm/viral-ngs-1.3.4/util/version.py", line 71, in get_version
    raise ValueError("Cannot find the version number!")
ValueError: Cannot find the version number!
```

If I create a file "VERSION" with 1.3.4 in it in the root directory, it seems to work. Is this the preferred way, or should I grab the latest master.zip from https://github.com/broadinstitute/viral-ngs instead of getting the latest official release from https://github.com/broadinstitute/viral-ngs/releases?

Thanks in advance!
Stefan

tomkinsc:
Hi Stefan,

Creating a `VERSION` file is fine; the latest master.zip will have the same behavior as an official tagged release. Ordinarily, if the copy of viral-ngs is a git checkout we use git to infer the version number from the `.git/` metadata. For tagged releases created by GitHub, no such directory is present but `git` will still output an artificial value to fill in the `VERSION` file. I'm guessing in your case git may not be installed? In that case, creating a `VERSION` file containing the version number is fine. 

The only real operation that will be impacted by the value in the `VERSION` file is `./ncbi.py prep_genbank_files`, where the version number is included in files destined for NCBI submission as part of the assembly method description. We also use it in auto-generating the [documentation](https://viral-ngs.readthedocs.org/en/latest/), and as the value returned by `./*.py --version`, but those cases are unrelated to usage of the pipeline.

I'll think about how we can change this behavior in future versions to make it more intuitive.

tomkinsc:
This should be resolved in #242. Closing.

PR #242, opened 2016-03-23
guess version number when git is unavailable or source is not git checkout

tomkinsc:
add a function to util/version.py, approx_version_number(), to output a
version number in the cases where git is unavailable, the copy of
viral-ngs is not a git checkout, and a VERSION file is not present. It
first tries to parse a version number from the viral-ngs root path
(likely for copies extracted from tagged releases), then falls back to
the mtime of the version.py file (since this will reflect either the
commit or tag date when the archive was created). In cases where mtime
cannot be accessed, the current time is used as the version number (the first time--it is saved to VERSION)

### DIFF
--- a/util/version.py
+++ b/util/version.py
@@ -1,3 +1,4 @@
+#!/usr/bin/python
 ''' This gets the git version into python-land
 '''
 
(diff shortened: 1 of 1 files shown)


tomkinsc:
This is response to #241

dpark01:
thanks--this is great. There's no reason to fail hard if we can't get the version number, so best guess is the way to go anyway.

PR #246, opened 2016-04-06
Rebuilt mvicuna binary on OS X without OpenMP.

yesimon:
Therefore we don't have to rely on the user having specific versioned
brew gcc runtime libraries because OS X clang does not support OpenMP.
This should work on all OS X machines. M-Vicuna is not a bottleneck
anyways.


tomkinsc:
Thanks, @yesimon. I don't see why we shouldn't merge this. 
Could you please open a second PR to the [separate repository](https://github.com/broadinstitute/mvicuna) we use as a source for the conda-based install of M-Vicuna?

PR #248, opened 2016-04-06
Add check and warning for bmtagger on OS X since it requires gnu-getopt.

yesimon:


tomkinsc:
If we switch to a fully conda-based install, this will not be needed since the bioconda package for [bmtagger](https://github.com/bioconda/bioconda-recipes/blob/master/recipes/bmtagger/meta.yaml) includes `gnu-getopt` as an OS X dependency.

PR #251, opened 2016-04-11
lastal arguments ordering, bmtagger conf

yesimon:
Move all options before positional arguments in lastal. Use customizable bmtagger conf.

Some `check=True` adds for subprocesses.

### DIFF
--- a/taxon_filter.py
+++ b/taxon_filter.py
@@ -3,6 +3,7 @@
 on membership or non-membership in a species / genus / taxonomic grouping.
 '''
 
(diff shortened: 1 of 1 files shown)


dpark01:
What does this change in the behavior of the pipeline as a whole?

yesimon:
This shouldn't change anything. I was getting some errors with lastal argument ordering on OS X so I just rearranged so that parsing should work. Bmtagger conf is the exact default arguments.

PR #259, opened 2016-04-12
Add autopep8 pre-commit hook.

yesimon:
Automatically fixes some trailing whitespace in .py files.


tomkinsc:
Thoughts on having the pre-commit hook also apply [`yapf`](https://github.com/google/yapf) formatting, in addition to the whitespace removal? We [already have](https://raw.githubusercontent.com/broadinstitute/viral-ngs/master/.style.yapf) a stye dotfile for yapf.

yesimon:
I think that might be a little too aggressive in a pre commit hook?
Obviously It can be optional. We should add it to the Travis report though.

tomkinsc:
Do we have `autopep8` as a dependency of any other package, because I don't see it in any of our requirements.txt files.

yesimon:
I don't think so, but commit hooks have to be manually installed as well?

PR #260, opened 2016-04-12
Upgrade util.misc.run more like py3.5 run.

yesimon:
Change the run function to be more feature complete like py3.5's run
function. Relies on temporary files for stdout/stderr since
subprocess.PIPE is not allowed when we're trying to capture them.
Otherwise, use check_output like before which is more efficient if we
are redirecting stderr to stdout or not capturing.

Therefore allow kraken to separate stdout as datastream writing to file
and stderr in log messages.

### DIFF
--- a/tools/__init__.py
+++ b/tools/__init__.py
@@ -206,6 +206,7 @@ class CondaPackage(InstallMethod):
         self.conda_env["CONDA_ENVS_PATH"] = conda_cache_path
 
         self.installed = False
+        self._is_attempted = False
 
         super(CondaPackage, self).__init__()
 
@@ -228,6 +229,12 @@ class CondaPackage(InstallMethod):
             ver_str = self.package
         return ver_str
 
+    def is_attempted(self):
+        return self._is_attempted
+
+    def is_installed(self):
+        return self.installed
+
     @property
     def bin_path(self):
         return os.path.join(self.env_path, "bin")
@@ -237,16 +244,13 @@ class CondaPackage(InstallMethod):
 
     @property
     def _package_installed(self):
-        result = util.misc.run_and_print(["conda", "list", "-f", "-c", "-p", self.env_path, "--json", self.package], silent=True, env=self.conda_env, buffered=True)
+        result = util.misc.run_and_print(["conda", "list", "-f", "-c", "-p", self.env_path, "--json", self.package], silent=True, env=self.conda_env)
         if result.returncode == 0:
             command_output = result.stdout.decode("UTF-8")
             data = json.loads(self._string_from_start_of_json(command_output))
             if len(data) > 0:
                 return True
         return False
-
-    def is_installed(self):
-        return self.installed
 
     def verify_install(self):
         # if conda does not the package as installed, no need to check further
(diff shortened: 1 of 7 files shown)


yesimon:
This change got a bit gnarly but should be fine now. Everything should be better once we move to `python>=3.5`!

tools/__init__.py
```diff
@@ -228,6 +229,12 @@ def _package_str(self):
             ver_str = self.package
         return ver_str
 
+    def is_attempted(self):
+        return self._is_attempted
+
+    def is_installed(self):
```
dpark01: by changing is_installed here, won't we lose the memoization of this return value that we were manually implementing before? so now it has to invoke "conda list" each time? @tomkinsc thoughts

yesimon: Actually the issue is that `CondaPackage` didn't have an `is_installed` even though it should be implemented. We can always add memoization later.

tools/__init__.py
```diff
@@ -237,17 +244,14 @@ def executable_path(self):
...
+        result = util.misc.run_and_print(["conda", "list", "-f", "-c", "-p", self.env_path, "--json", self.package], silent=True, env=self.conda_env)
         if result.returncode == 0:
             command_output = result.stdout.decode("UTF-8")
             data = json.loads(self._string_from_start_of_json(command_output))
             if len(data) > 0:
                 return True
         return False
 
-    def is_installed(self):
-        return self.installed
```
dpark01: CondaPackage had this version of is_installed previous to this PR, which returns self.installed, which is set by the verify_install() which manually memoizes its return value.

dpark01:
I just reviewed the diff in this PR and it looks good overall. It's nice to introduce the separation of stdout/stderr and allow the pipeability while retaining the ability to capture if necessary.

I think the one thing I would ask is to revert the CondaPackage.is_installed method to its previous implementation assuming that it doesn't cause any problems. The previous implementation tested for more than just _package_installed-ness, it also tested for executability of the binary sometimes.

tools/__init__.py
yesimon: Whoops that is a mistake on my behalf. I think I was running into problems before where I wasn't calling `install_and_get_path()`, which resulted in problems. I still might have to make a change on `verify_install` to allow optionally skipping executability check because krona doesn't have a `krona` executable.

dpark01:
All you have to do is instantiate the CondaPackage with require_executability=False and you should be set in that case.

On another note, how is it that your PRs show up as a single commit? Are you rebasing each time? I can't see the diff between what I reviewed last night vs what you introduced just recently.

yesimon:
I typically do `git commit --amend` followed by a force push to my branch. This works well for keeping commit history clean, but not great for github line tracking (gerrit does this way better). Perhaps for larger commits I'll just do a squash at the end of the review process. For small commits this adds the overhead of waiting for travis twice.

dpark01:
Keeping commit history to a minimal # of commits isn't really that important, and especially in the middle of a review process, you want to keep them separate. Doing a squash prior to the initial PR makes sense though. Leaving all the revisions in during PR is actually helpful in reviewing history.

Can you manually describe the diff between last night and now?

yesimon:
The only change was to revert the is_installed change for CondaPackage.

PR #261, opened 2016-04-12
Separate trinity conda env from default conda env.

yesimon:
Change the CondaPackage param from env_path to env to make separate
environments easier. Trinity is going to be installed in a non-default
env because it relies on perl-threaded, which conflicts with the default
perl package that krona uses (to be safe).

### DIFF
--- a/tools/__init__.py
+++ b/tools/__init__.py
@@ -181,8 +181,10 @@ class CondaPackage(InstallMethod):
         verifycmd=None,
         verifycode=0,
         require_executability=True,
-        env_path=None,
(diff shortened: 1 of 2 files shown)


tomkinsc:
Nice, having alternate envs as children of the env_root_path seems like a reasonable default. It would still be nice to be able to specify a different env_root_path in case we want `conda-tools/` stored elsewhere. Can you add another parameter for the env root, with the default set as it is currently?

yesimon:
Readded the root_path arg.

Issue #257, opened 2016-04-12
implement krona entry point and Tool


dpark01:
Create python wrapper command "metagenomics.py krona" and create Tool wrapper.

- input: kraken-formatted summary report (TXT)
- output: krona D3 report (single standalone HTML file)

dpark01:
depends on #255

Issue #255, opened 2016-04-12
implement kraken entry point


dpark01:
Create python wrapper command "metagenomics.py kraken" that utilizes existing Tool.

- input: unaligned reads (BAM), database (path to unpacked directory)
- output: kraken summary report (TXT)
- output [optional]: kraken raw report per-read (TXT.GZ and/or TXT.LZ4)

For now, don't worry about slurping down the DB from the outside, or serializing operations across multiple inputs--let the dnanexus wrapper handle that in the first iteration.

PR #264, opened 2016-04-14
Metagenomics snakemake rules.

yesimon:
For now runs kraken+krona. Only run if kraken_db is defined. Kraken db
setup process is an out of band process.


metagenomics.py
```diff
@@ -0,0 +1,86 @@
...
+import gzip
+import shutil
+import util.file
+import tools.kraken
+import tools.krona
+import tools.diamond
+__commands__ = []
+
+
+def kraken(inFastq1, inFastq2, db, outReport=None, outReads=None,
```
dpark01: As a general UI rule, we've been using exclusively BAM files as inputs and outputs for viral-ngs top-level scripts. Fastqs are always hidden, ephemeral / never persist. Can you modify this to take a single BAM as input and start off the method with a conversion as done [elsewhere](https://github.com/broadinstitute/viral-ngs/blob/master/taxon_filter.py#L271)? Same for the krona method.

illumina.py
```diff
@@ -507,6 +507,19 @@ def fetch_by_index(self, idx):
 # ***  miseq_fastq_to_bam   ***
 # =============================
 
+def sam_to_fastq(outFastq1, outFastq2, inBam, clippingAction=None, clippingAttribute=None,
```
dpark01: You can delete these additions to illumina.py (which may have been better housed at read_utils.py anyway) as they are unnecessary if you make the requested changes to the kraken and krona methods.

metagenomics.py
```diff
@@ -0,0 +1,86 @@
...
+           filterThreshold=None, threads=1):
```
dpark01: is the threads=1 argument extraneous? It doesn't get exposed in argparse, and it doesn't appear to be used later in this method.

metagenomics.py
```diff
@@ -0,0 +1,86 @@
...
+    }
+
+    kraken_tool.execute('kraken-filter', db, tmp_filtered_reads, args=[tmp_reads],
+                        options=opts)
+
+    if outReads:
+        with open(tmp_filtered_reads, 'rb') as f_in:
+            with gzip.open(outReads, 'wb') as f_out:
+                shutil.copyfileobj(f_in, f_out)
+
```
dpark01: just to be safe, can we explicitly os.unlink(tmp_filtered_reads) at the end and any other large things? It shouldn't matter in a full invocation of the script as the tmpdir gets blown away at the end--unless something calls the function directly within python (e.g. nosetests)

metagenomics.py
```diff
@@ -0,0 +1,86 @@
...
+    util.cmd.common_args(parser, (('loglevel', None), ('version', None),
+                                  ('tmp_dir', None)))
+    util.cmd.attach_main(parser, kraken, split_args=True)
+
+
+def parser_krona(parser=argparse.ArgumentParser()):
+    parser.add_argument('inTsv', help='Input tab delimited file.')
+    parser.add_argument('outHtml', help='Output html report.')
+    parser.add_argument('--queryColumn', help='Column of query id.',
+                        type=int)
```
dpark01: Since we seem to be leaning towards using the kraken summary report as the defacto standard format for how we will represent this kind of information, can you go ahead and set the default argparse values for --queryColumn, taxidColumn, and scoreColumn to be whatever kraken emits?

metagenomics.py
```diff
@@ -0,0 +1,86 @@
...
+            with gzip.open(outReads, 'wb') as f_out:
+                shutil.copyfileobj(f_in, f_out)
+
+    if outReport:
+        kraken_tool.execute('kraken-report', db, outReport, args=[outReads],
+                            options=opts)
+
+def krona(inTsv, outHtml, queryColumn=None, taxidColumn=None,
+          scoreColumn=None, noHits=None, noRank=None):
+    krona_tool = tools.krona.Krona()
```
dpark01: if(inTsv.endswith('.gz')), decompress to a temp file and use that instead. It'd be nice if our krona wrapper can take input directly from our kraken wrapper w/o having to externally process it first.

pipes/Snakefile
```diff
@@ -25,15 +25,18 @@ include: os.path.join(pipesDir, 'ncbi.rules')
 include: os.path.join(pipesDir, 'intrahost.rules')
 include: os.path.join(pipesDir, 'reports.rules')
```
dpark01: looks great, except don't forget to include metagenomics.rules at the top

pipes/config.yaml
```diff
@@ -1,19 +1,19 @@
 # ==================== [ Values that must checked ] =========================
```
dpark01: looks great

pipes/config.yaml
```diff
@@ -126,15 +136,15 @@ trinity_n_reads: 250000
...
 
 # Sub-directories within data_dir to store the input and output
 # files for various stages of the pipeline.
-# For raw sequencer output, the files should go in demux or source, 
+# For raw sequencer output, the files should go in demux or source,
 # depending on whether or not they have been demultiplexed
-subdirs: 
+subdirs:
   demux: "00_demux"
   source: "00_raw"
```
dpark01: don't forget to add a metagenomics subdir

pipes/rules/metagenomics.rules
```diff
@@ -0,0 +1,66 @@
...
+    input: join(config["data_dir"], config["subdirs"]["metagenomics"], "{sample}.kraken.reads.gz")
+    output: join(config["data_dir"], config["subdirs"]["metagenomics"], "{sample}.kraken.krona.html")
+    shell:
+        """
+        zcat {input} | {config[bin_dir]}/metagenomics.py krona - {output} --queryColumn 2 --taxidColumn 3 --noRank
+        """
+
+
+rule pair_fastq:
+    input: expand(join(config["tmp_dir"], config["subdirs"]["source"], "{{sample}}.{pair}.fastq"), pair=(1,2))
```
dpark01: I know diamond isn't implemented at this point, but eventually when it is, this pair_fastq code can move into that wrapper as well.

dpark01: Actually, for both this part (pair_fastq) and above (rule diamond), just leave as-is for now (at most, just comment it out if it's causing snakemake any confusion), since we can work on that wrapper later. The bam_to_fastq_paired rule can be removed or commented, just remember to use those two clipping rules when invoking the picard Tool in the kraken wrapper script.

dpark01:
Please don't rebase/flatten/squash the next commit--it's nice to see the diffs between each revision

dpark01:
closes #257 and #255

metagenomics.py
```diff
@@ -78,8 +99,42 @@ def parser_krona(parser=argparse.ArgumentParser()):
...
+    }
+    picard.execute(inBam, [tmp_fastq, tmp_fastq2] , unpaired=True,
+                   picardOptions=picard.dict_to_picard_opts(picard_opts),
+                   JVMmemory=picard.jvmMemDefault)
+
+    diamond_tool = tools.diamond.Diamond()
+    diamond_tool.install()
+    tmp_alignment = util.file.mkstempfname('.daa')
+    tmp_m8 = util.file.mkstempfname('.diamond.m8')
+    diamond_tool.blastx(db, [tmp_fastq], tmp_alignment,
```
dpark01: Instead of modifying tools/picard.py, why not just do a normal SamToFastq and then cat the two outputs together to a temp file for diamond to take as input?

dpark01:
Looking pretty good--can you revert all changes to illumina.py and tools/picard.py to keep the PR clean (so the github diff shows only 7 changed files in total)? Also I just merged in the latest origin/master.

metagenomics.py
```diff
@@ -0,0 +1,142 @@
...
+import util.file
+import tools.kraken
+import tools.krona
+import tools.diamond
+import tools.picard
+__commands__ = []
+
+
+def kraken(inBam, db, outReport=None, outReads=None,
+           filterThreshold=None, threads=1):
```
dpark01: the kraken "threads" parameters still looks to be unused / extraneous, remove?

yesimon: Actually using it now. It's pretty important because kraken actually has good boost with threads.

yesimon:
I've left minimal changes to picard.py to keep the `illumina_clipping_attribute` because it's a consistent constant.

dpark01:
maybe I'm just missing it, but I don't see where the threads (or numThreads, there appears to be an impedance mismatch between argparse and the method itself) is being passed on to the kraken tool itself... I see it being passed to diamond, but not kraken.

dpark01:
Can you add a test/unit/test_metagenomics.py that, for now, just contains the TestCommandHelp class that most of the other things have? it just tests that the argparse --help works for everything in metagenomics.py.

Separately, I might also ask for a krona unit test too, since the input/output files seems small, and I imagine the runtime is short, but only if that's quick to implement on a tiny dataset.

yesimon:
Thanks for catching the kraken threads. I need to do a few more things to have a proper krona unit test, but other things are good.

dpark01:
It still looks to me as if both the kraken and diamond methods take a "threads" optional parameter in the method definition, but argparse is giving both of them a "numThreads". Does it run? Otherwise, looks pretty good to me.

dpark01:
Great, please do add a simple krona unit test, I'll merge this in for now

Issue #269, opened 2016-04-21
kraken gzipped reads output file errors


dpark01:
Bug: if --outReport specified, [metagenomics.kraken](https://github.com/broadinstitute/viral-ngs/blob/master/metagenomics.py#L51) runs kraken-report on "outReads" as input. This may not exist (if --outReads is None) and even if it does, it is likely gzipped, which kraken-report won't like. Perhaps it should take "tmp_filtered_reads" as input?

Desire: can the [opening of outReport](https://github.com/broadinstitute/viral-ngs/blob/master/metagenomics.py#L47) be changed to use util.file.open_or_gzopen instead? Might be nice to allow people to write out uncompressed .txt or compressed .txt.gz files depending on the output filename.

Issue #268, opened 2016-04-21
kraken-report doesn't like --threshold


dpark01:
In [metagenomics.kraken](https://github.com/broadinstitute/viral-ngs/blob/master/metagenomics.py#L52), the same opts map is used for kraken-filter and kraken-report, but kraken-report fails with perl errors:

Unknown option: threshold
Can't open 0.05: No such file or directory at _(long path here)_/tools/build/conda-tools/default/bin/kraken-report line 80.

PR #272, opened 2016-04-21
Fix some metagenomics bugs and add unit tests

yesimon:
Much more comprehensive unit testing with mocking out actual subprocess calls just so we can make sure arguments are being sent to the right places.


yesimon:
To help with #269 #268

dpark01:
Looks like a quick & easy fix?

PR #273, opened 2016-04-22
Add snakemake integration runner and kraken test

yesimon:
To test the entire pipes flow.

### DIFF
--- a/test/integration/test_kraken.py
+++ b/test/integration/test_kraken.py
@@ -1,0 +1,71 @@
+# Integration tests for kraken
+
+from builtins import super
+import os.path
+from os.path import join
+import tempfile
+import unittest
+import util.file
+import tools.kraken
+import tools.picard
+from test import TestCaseWithTmp
+from test.integration.snake import SnakemakeRunner
+
+
+class TestKrakenTiny(TestCaseWithTmp):
+
+    def setUp(self):
+        super().setUp()
+        self.kraken = tools.kraken.Kraken()
+        self.kraken.install()
+
+    @property
+    def data_dir(self):
+        return join(util.file.get_test_input_path(), 'TestToolKraken')
+
+    @property
+    def db_dir(self):
+        return os.path.join(self.data_dir, 'db')
+
+    def build_kraken_db(self):
+        db = os.path.join(tempfile.tempdir, 'db')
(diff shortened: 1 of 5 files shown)


dpark01:
Hm, I don't think this is going to work for various reasons. I think the better approach would be to:

1. Write basic integration tests that run kraken and krona on tiny test-case input (100 pairs of reads) without using snakemake. See test.unit.assembly for a few examples of things that arguably ought to be moved to test.integration (but we'll leave that for another time).  I really do want an execution of the actual binaries to be tested in a Py2.7 environment, which means it can't use snakemake.

2. See test.unit.test_snakemake.TestSimpleDryRuns for an example of something that tests the snakemake rules themselves, but skips the tests if Python < 3.2.

yesimon:
I think we can test at a high level both the pipes and the python bins run via the command line, such that we make sure that argparse also works. This is the same test that was the original kraken "unit" test.

dpark01:
That would be great so long as an execution test exists for 2.7 that does not rely on snakemake

test/integration/test_kraken.py
```diff
@@ -0,0 +1,71 @@
...
+            os.symlink(realpath, name)
+
+        # Minimizer len corresponds to memory/disk usage of index.
+        self.assertEqual(0, self.kraken.build(db, options={
+            '--minimizer-len': 10,
+            '--build': None,
+        }).returncode)
+        return db
+
+    def test_pipes(self):
```
dpark01: Please add the decorator:
@unittest.skipIf(sys.version_info < (3, 2), "python version is too old for snakemake")

yesimon:
In this I've modified TestCaseWithTmp to add a class level tmpdir for the purpose of caching a kraken db across multiple integration tests such that it can be reused.

dpark01:
Nice... that's a useful add to that class.

dpark01:
wtf--how long has github had like buttons and emoji's?? @tomkinsc

dpark01:
from future import print command?

tomkinsc:
About a month? It was in response to a [letter](https://github.com/dear-github/dear-github) complaining about `+1`s cluttering comment strings. Sadly the available "reactions" on GitHub don't include the full unicode emoji set...

PR #275, opened 2016-04-23
Add integration test for kraken viral mix

yesimon:


dpark01:
Looks like the kraken-build is failing, but luckily the other tests appear to be succeeding. Unfortunately there isn't much debug output describing why. I don't think it should be a memory issue, but if it were, you could always reduce the --minimizer-len. Mem/index use is exponential against that parameter, and my understanding is that it only affects performance (which doesn't matter for tiny test data), not behavior/output.

yesimon:
The error was a missing mock patch for jellyfish on the previous PR. It seems like the travis build is passing for the merge but still has the failing result from before the merge?

dpark01:
Yeah so I just merged your other branch into master that fixed the jellyfish stuff, then I merged master back into this branch. So those tests are now passing on Travis on this PR. What's failing is the integration test for kraken-build.  Other integration tests for kraken seem okay.  Note that integration tests do not execute on the travis-ci/push cycles, only on the /pr cycles.

yesimon:
That explains why it was failing only for the push cycle.

PR #277, opened 2016-04-25
Add diamond integration test

yesimon:


dpark01:
In my mind, there's no point having a diamond integration test until the diamond python script (metagenomics.diamond) gives a kraken-formatted output file. If you have an end-to-end test, I'd rather test that end.

dpark01:
I'm going to close the PR, but let's keep the branch alive since there's good work in here. I'd rather wait until the LCA code ported from Julia is in so that metagenomics.diamond outputs a kraken-formatted text file identically formatted to the --outReads from metagenomics.kraken. (and maybe it can also have an --outReport that actually is produced by kraken-report). I don't want to bake in any nosetests in the master branch that enforce a user interface (blast m8 files) we don't ultimately want.

PR #278, opened 2016-04-25
Add krona integration test running after kraken

yesimon:


dpark01:
can you keep this PR topic clean (ie, separate the diamond-related stuff)

yesimon:
Fun times with git rebase! Still deps on viral mix integration test.

yesimon:
One other thing - this will not pass tests until krona gets updated to support local db builds a bit better. I might need to have to apply a patch.

dpark01:
github.com says you'll need to hand-merge the changes in master for this one (probably because of the typos in the original version of the kraken integration test). can you handle that?

Also, are you saying that metagenomics.krona generally won't work on most machines (because it auto-builds a local db on the first execution of it)?

yesimon:
Sorry, what I mean is marbl/Krona#10 needing to be in so I don't have to zip up the test files just to unzip them right after.

dpark01:
I don't know how frequent krona's release cycle is, but in the worst case, can we just add a gzip/gunzip cycle and get around it?

yesimon:
I just noticed that the krona taxonomy db for the viral mix integration test is 1GB, inflating our cache size of 2GB by 50%.

dpark01:
How long does it take to build the taxonomy db, and does that process involve network hits on NCBI? If so, it's probably worth caching something that big anyway... or... possibly even figuring out how to create a reduced taxonomy db that's small enough to fit in the git repo for the purposes of krona tests. Tests should definitely not require NCBI-connectivity to succeed (or they'll certainly fail if you run them too often).

dpark01:
Are the current failures due to problems with the tests, or problems with the tool? I can't tell from this output whether metagenomics.krona works.

yesimon:
I'm actually unsure about the current failures. It passes locally on 3.5 with nosetests. Fails with tox and travis as if updateTaxonomy.sh were called twice - (the first call will delete the *.dmp files), the second looks for it and panics, although I can't find a code path where they are called twice, even with logging and specifying a single test. The size probably has to do with the internal binary format that krona uses to represent the taxonomy tree, which my guess is probably an array of size `max(taxid)`, so it's not the size but the contents that matter more.

yesimon:
With more extensive patching by taking out the `rm` commands on the original *.dmp files, it passes. But that's not addressing probably a greater underlying problem.

dpark01:
Can you mock to debug the tests (and figure out where the 2nd invocation is happening)?

yesimon:
Now it works on tox/nosetests locally but fails travis for different reasons :(

dpark01:
I think these tests are structured in a way that I can't really tell what the failure is. The travis log looks the same to me from build 1940 and 1942. You said it's failing for different reasons now?

test/integration/test_kraken.py
```diff
@@ -117,5 +154,34 @@ def setUpClass(cls):
...
+        bin = join(util.file.get_project_path(), 'metagenomics.py')
+        out_report = util.file.mkstempfname('.report')
+        out_reads = util.file.mkstempfname('.reads.gz')
+        cmd = [bin, 'kraken', self.bam, self.db, '--outReport', out_report, '--outReads', out_reads]
+        util.misc.run_and_print(cmd, check=True)
+        out_html = util.file.mkstempfname('.krona.html')
+        cmd = [bin, 'krona', out_reads, out_html, '--db', self.krona_db]
+        print(cmd)
+        try:
+            util.misc.run_and_print(cmd, check=True)
```
dpark01: I've been realizing one big drawback to the way you're testing commands (by actually spawning a whole python subprocess in order to test argparse and everything) is that it means none of it shows up in coveralls at all.

The way most of our other unit tests do this would look like this:
`
 parser = metagenomics.parser_krona(argparse.ArgumentParser())
 args = parser.parse_args([out_reads, out_html, '--db', self.krona_db])
 args.func_main(args)
`

Can you switch these invocations of run_and_print over?

test/integration/test_kraken.py
```diff
@@ -92,20 +128,21 @@ def setUpClass(cls):
         cls.bam = cls.input_bam('zaire_ebola')
 
     def test_kraken_tool(self):
-        output = os.path.join(tempfile.tempdir, 'zaire_ebola.kraken')
-        output_filtered = os.path.join(tempfile.tempdir, 'zaire_ebola.filtered-kraken')
-        output_report = os.path.join(tempfile.tempdir, 'zaire_ebola.kraken-report')
-        self.assertEqual(0, self.kraken.classify(self.db, self.fastqs, output).returncode)
+        out = join(tempfile.tempdir, 'zaire_ebola.kraken')
```
dpark01: I think you need to create a mkdtemp instead of just dropping things straight in /tmp

dpark01:
(I'm not even totally sure which python the subprocess call of python would invoke--the system python or the one that comes with conda...)

dpark01:
A few comments:
- this is an odd hack to fix a previously unnecessary hack
- while some aspects of the recent tweaks are potentially useful, this PR is getting entirely out of scope of the original topic and any changes to the travis/coveralls setup should be dealt with separately
- something is wrong if proper execution tests of the kraken and krona scripts actually requires any of this
- your failures are harder to interpret because you're swallowing the output with run_and_print instead of letting nose do the suppress-unless-failure work that it normally does. run_and_print's whole reason for existence was really only for direct calls to external binaries that nose didn't have control over--not for python code.

I'll reiterate that the subprocess/run_and_print calls of the top-level python scripts should all be removed from test/* as it is an unnecessary departure from the existing test patterns and is creating problems without adding value. And the other reason to do this is because I said so. Given more time, I'd be happy to discuss alternatives and learn new ideas, but not for this particular PR / feature set, which has already been delivered downstream w/o test coverage.

yesimon:
Sorry for the travis spam and the ridiculous amounts of debug noise - trying to debug an issue with tests failing on travis but succeeding locally. The root cause is actually conda improperly finding the wrong krona package - this is something that due to a conflux of changes in bioconda and pinning of big dependencies like perl, as well as not having the right amount of logging on our side.

The coverage attachment is not a hack and is actually fully documented and supported: see http://coverage.readthedocs.io/en/coverage-4.0.3/subprocess.html , and is the only option for coverage if we ever decide to test our snakemake rules. If you don't like writing a sitecustomize.py we can just ignore it and ignore coverage for snakemake rules, as coverage should be covered by unit tests anyways. I'm fine with switching the run_and_print calls over as the additional coverage by invoking the entire python script is negligible.

dpark01:
Feel free to do induce noise on this branch in order to get travis to trigger for you--that's pretty common practice in these scenarios, I do it myself all the time.

I saw the diff on the coveralls/travis stuff and was quite interested in much of it--for a future PR.

Is krona using a different perl than the rest of our tools? I thought we recently reconciled the threaded/unthreaded disparity upstream in the bioconda channel.

tools/krona.py
```diff
@@ -35,6 +35,8 @@ def import_taxonomy(self, input_tsvs, output, query_column=None, taxid_column=No
                         score_column=None, no_hits=None, no_rank=None, db=None):
         self.install()
         bin_path = os.path.dirname(self.executable_path())
+        env = os.environ.copy()
+        env['PATH'] = '{}:{}'.format(bin_path, env['PATH'])
```
dpark01: Wouldn't it make more sense to add the opt/krona/lib/ directory to env['PERL5LIB']?  This seems to be the only way to make metagenomics.krona not hit perl errors on my mac.

dpark01:
In taking a closer look at this, something else is wrong with the metagenomics.krona command. It fails unless someone calls tools.krona.Krona.create_db before running it, which is not an ability that's currently exposed to the user, so it will always fail on a fresh install. Here are some options:

1. Add a metagenomics.create_krona_db command and require the user to call it the first time -- probably my least favorite option.
2. Have metagenomics.krona auto-detect if a db exists in default location and, if not, create_db one.
3. Have tools.krona.Krona.import_taxonomy auto-detect if a db exists in default location and, if not, create_db one.
4. Add some optional post-install hook to CondaPackage(InstallMethod) and, in tools.krona.Krona's case, use it to create_db immediately after install.

What do you think?

dpark01:
I guess there's also this option:
5. package and compress and distribute our own pre-built krona db at a public url, similar to what we do for kraken and diamond. I suppose this might be the most consistent with the other metagenomics.py commands? Then we should change the --db argument to a positional required parameter (maybe the middle of three).

yesimon:
I think we might want to add the conda bin path to our path generally as some programs assume that it's installed to your `$PREFIX`.

I don't know what the right answer is, mainly because creating a database is a huge side effect if we're doing it lazily/automatically, and might be very surprising to see large disk space usage or running a sample taking a very long time. Krona might be alright to do automatically because it's just the taxonomy database, and not a reference database.

dpark01:
The more I think about it the more I lean towards #5. It's most like kraken and diamond and I'm weary of side effects that hit an ftp site on NCBI (certainly will cause intermittent failures of tests if we allow it to).

Can you add the positional required arg to the krona command?

dpark01:
Ok great. This looks pretty good, but I want to merge in PR #283 before merging in this one.

In the meantime, can you also create a krona database tarball and drop it somewhere in sabeti-scratch or hptmp for me to push out?

yesimon:
Yeah there are a few bugfixes for `run_and_print` in this PR, maybe that will help #283?

dpark01:
Ok, can you take them out of this PR and offer a patch to @tomkinsc for that piece of it?

dpark01:
Can you hand merge origin/master back into this branch? The octocat appears to be failing on this one without a little help.

metagenomics.py
```diff
@@ -107,6 +110,7 @@ def parser_krona(parser=argparse.ArgumentParser()):
                         action='store_true')
     parser.add_argument('--noRank', help='Include no rank assignments.',
                         action='store_true')
+    parser.add_argument('--db', help='Directory of taxonomy db.')
```
dpark01: can you remove the --db optional argument from parser_krona?

test/integration/test_kraken.py
```diff
@@ -1,13 +1,17 @@
 # Integration tests for kraken
 
 from builtins import super
+import argparse
 import os.path
 from os.path import join
 import sys
+import subprocess
```
dpark01: unused import subprocess

dpark01:
Sorry about the last two comments after all that, but we're very very close to final merge. The second comment isn't a show-stopper, but the first one might cause problems maybe possibly..

dpark01:
looks like the test is still calling the --db arg

yesimon:
Hopefully we're good now

PR #280, opened 2016-04-28
add optional argument to specify thread count for blastn

tomkinsc:


dpark01:
Good catch. One thing though: what is the default behavior of blastn if this parameter is omitted? Instead of setting the default to 4, can we just set the default to None (and don't specify anything to blastn if threads is None)?

tomkinsc:
The [default](https://www.ncbi.nlm.nih.gov/books/NBK279675/) for `blastn` is to run with one thread. Four threads seemed like a reasonably default since most systems these days are likely to have four cores or two hyper threaded cores. Even in the case where it would be running on a single core, the penalty of running of running four threads shouldn't be huge since the search threads are reasonably independent with minimal sync overhead, so it'll mostly look like single-threaded sequential operation.

We can certainly change the default value.

PR #283, opened 2016-05-02
add check=True to run_and_print() calls

tomkinsc:
the ‘check’ kwarg (default False) was added to misc.py:run_and_print(),
which checks the return code of the called command and raises a
CalledProcessError if returncode != 0. This commit adds check=True to
instances of run_and_print where run_and_print previously replaced
invocations of subprocess.check_call() (but not invocations of
subprocess.call())


dpark01:
Thanks for this. I sure hope all lights turn green on this build, but if they don't, it's quite possible that the problem is that we've already introduced buggy code into the master branch w/o knowing it.

PR #285, opened 2016-05-02
Fix for run_and_print running twice

yesimon:
Also adds a logging facility where subprocess commands are printed with
a specified loglevel instead of to stdout.

### DIFF
--- a/util/misc.py
+++ b/util/misc.py
@@ -2,6 +2,7 @@
 from __future__ import print_function, division  # Division of integers with / should never round!
 import collections
 import itertools
+import logging
 import os
 import subprocess
 import sys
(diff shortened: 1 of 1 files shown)


util/misc.py
```diff
@@ -175,19 +180,38 @@ def run(args, stdin=None, stdout=None, stderr=None, shell=False,
...
-        if not silent:
-            print(result.stdout.decode('utf-8'))
+        if check and not silent:
+            try:
+                result = run(args, stdin=stdin, stdout=subprocess.PIPE,
+                           stderr=subprocess.STDOUT, env=env, cwd=cwd,
+                           timeout=timeout, check=check)
+            except subprocess.CalledProcessError as e:
+                if loglevel:
+                    logging.log(loglevel, result.stdout.decode('utf-8'))
```
dpark01: Interesting... instead of using the root level logger (logging.log), can you do what the other toplevel scripts do and put a
```
log = logging.getLogger(__name__)
```
global variable at the top of the file and use that "log" variable here?

Also separately, you'll want to merge origin/master back in here, as Chris has independently fixed the run-twice issue there.

yesimon: Actually updated on krona branch because it is already merged, so closing this. The purpose of this will become clear shortly.

yesimon:
Fixed in #283

Issue #256, opened 2016-04-12
implement diamond/lca entry point


dpark01:
Create python wrapper command "metagenomics.py diamond" that utilizes existing Tool.

- input: unaligned reads (BAM), database (path to unpacked directory)
- output: summary report in kraken format (TXT)

For now, don't worry about slurping down the DB from the outside, or serializing operations across multiple inputs--let the dnanexus wrapper handle that in the first iteration. Also, the raw per-read output in kraken format will eventually be desired, but we can implement that later.

PR #286, opened 2016-05-02
Switch to py.test

yesimon:
Does subprocess capturing properly so we may convert many run_and_print into run_and_log invocations.


yesimon:
I don't have an answer for `TestToolsInstallation`. Theoretically tools should be installed automatically before first use - otherwise we should probably just make a tiny change to be able to install tools manually.

dpark01:
Intriguing, but I'm definitely tabling this one for further consideration after we get past more pressing things, such as Issue #256

tomkinsc:
I looked at switching to pytest last fall, based on its support for multiprocess test load-balancing, but deferred it based on the `TestToolsInstallation` generative test issue (not yet resolved in this PR?).

Ultimately, what does this buy us that we don't have with `nose`? Can it capture subprocess output out of the box? 

What do you mean by replacing `run_and_print()` with `run_and_log()`? Capturing output at different log levels? How do you propose we stratify the levels? 

If, for example, DNAnexus has hooks for a fancier logging service, I could see where it might be nice to have different log levels. ex. "pipeline-user" (see where your job fails, with a simple reason), "pipeline-debug" (see much more detailed info).

yesimon:
The new pytest passes just fine! It can capture subprocess out of the box AND do coverage of subprocess out of the box! Easier parallelization and `xdist` allows automatically rerunning of tests whenever code changes in a dev environment. There are lots of other features like DI-like fixtures and tempdir integration, but it supports almost all of unittest constructs so both can be used at the same time.

The `run_and_log` is a separate idea. The problem is in some places we want programs to be quiet except when we are debugging. In this case, we use `silent=True`, but then we have to manually change for debugging. By setting a low loglevel like `debug`, it will be silent under normal program execution but under tests we can set logger handler's level to `debug`, which is fine because it will still be captured unless there's an error.

tomkinsc:
That all sounds great! Are there any downsides to the change?

yesimon:
The only downside is it doesn't support the nose-specific test tools installation, which just took a tiny tweak. On the other hand, I had a fairly good time with it for the blast lca changes because it handily prints func args and a good chunk of code where it errors instead of just the exception stacktrace.

dpark01:
Been thinking about this one more. If we go this way, can we eliminate util.misc.run* throughout the code base and revert to call/check_call?

Nose was its only reason for existence originally, and py.test would obviate that. It'd be real nice to no longer buffer the output of log running programs and be able to see their output in realtime on platforms that support it (UGER's stdout stderr logs and DNAnexus's web logs of running instances). Plus I always thought it was weird to squash an entire program's output to a single loglevel. Maybe that makes sense for "make" but for lots of other things, I'd rather pass a verbosity or loglevel parameter down to the program itself.

That would definitely sell me on this switch!

yesimon:
Yes the main reason I was going for py.test for better subprocess capture. 

I do think we need a `run_and_log` wherever we're doing `run_and_print(silent=True...` just to aid debugging some failures.

dpark01:
I did a quick grep... the conda installers seem to be the main place where
run_and_print silent=True is happening, is that right? I see it in GATK and
snpEff but I'd actually prefer to remove those cases if py.test can swallow
the output just fine.

Also, if we do transition to primarily conda-based installs, that
definitely reduces make/gcc-based noise we used to experience.

yesimon:
That is right. I think... we can use `call` or `check_call` as long as we don't need the stdout or stderr captured in python anywhere.

dpark01:
Can you strip out the following code from all the test/unit/ and test/integration py files?
'''
if __name__ == '__main__':
    unittest.main()
'''
That is a bit vestigal from when we were using unittest.

dpark01:
Do you know what all the "Coverage.py warning: Module <xxx> was never imported" lines in the build log are about? I'm assuming it's incorrect, but do you know why it's happening?

yesimon:
It's coming from the coverage package, which I'm not sure on how to control its verbosity because it's being called from pytest-cov. I think it's actually correct because some modules are not imported for integration tests.

dpark01:
Yeah but I see it naming every single top-level module in tests-unit.sh, which can't possibly be right.

tomkinsc:
The test speedup by running on Travis's 2 cores via xdist is nice. I see the thread count specified in the `pytest.ini` file. 

Is the option precedence such that we can override it locally by calling `py.test -n NUM [...]`?

tomkinsc:
Also, has tool installation been removed from `easy-deploy/ansible-playbook.yml`?

Shouldn't the nosetests call be replaced with `py.test -v test/unit/test_tools.py`?

yesimon:
The speedup isn't that great because it's 1.5 cores? What does that even mean... Better than nothing I guess

Command line opts override pytest.ini opts. I re-added the ansible playbook tools installation. Ideally it shouldn't require pytest.

dpark01:
I agree that ideally, we'd have some sort of install-all-tools script that
didn't rely on any testing framework. That is, I'd like folks to be able to
install the whole thing without pip installing requirements-tests.txt at
all. Maybe that's a separate PR.

Would either of you like to undertake a very careful replacement of almost
all of the util.misc.run_* invocations with subprocess call/check_call?  By
"careful", I mean if you get it wrong, you either break the build or
introduce opportunities for silent failures.

tomkinsc:
Separating tool installs from tests fits in with a general move to use a monolithic conda environment file for all dependencies, so we could wait for that—if we go the conda route.

From the [Travis docs](https://docs.travis-ci.com/user/ci-environment/#Virtualization-environments) it looks like we do get two cores now for container-based environments. Their hypervisor/manager probably does something like fractional time sharing, so 1.5 cores kinda makes sense...it would be something like 2 cores for 75% of a time window.

tomkinsc:
I'd argue it could be worth preserving `util.misc.run_and*`, renaming it, and changing its subprocess calls to pass stdout normally. This would give us a convenience function along the lines of `run_and_log`, as @yesimon originally suggested. It could be nice in case we ever want to change how all of our subprocess calls are made (instrumentation?). The `util.misc.run_and_save` function is arguably more concise than a lengthy Popen expression for the instances where we want to save stdout (though perhaps the `run_and_save` could be merged with `run_and_log`).

dpark01:
Okay well how about this, let's just merge in the latest origin/master, figure out what's causing Travis timeouts (maybe py.test needs to be more verbose?) get the green lights and merge it in?

Seems like a good win. In retrospect, while this project started off with nose on Travis, we never really did anything fully nose-specific (and were supporting unittest for quite some time) up until relatively recently. The subprocess output capture is the biggest reason to go for this.

As for refactoring of the run_and_* stuff, agreed that we should leave that for a future PR.

yesimon:
Travis timeouts are because the krona tests are inherently bad. They require writing multiple gigs of basically zeros to probably travis underlying HDDs, while I am testing locally on fast SSDs so that time is negligible. I think we should skip them for now - krona master branch has a new version that obviates my patching and doesn't store a large array, but requires accessions so I'm subsetting the taxonomy for the accessions right now. We need to skip because we need to wait for versioned krona release.

tomkinsc:
@yesimon, could we make use of [`travis_wait`](https://docs.travis-ci.com/user/common-build-problems/#Build-times-out-because-no-output-was-received) to keep the test from timing out? 
I guess it wouldn't solve the problem builds taking a long time...

dpark01:
That sounds fine, go ahead and add a skip decorator to krona, it'd be nice to merge this one in soon

Issue #279, opened 2016-04-28
Update Diamond version?


tomkinsc:
The viral-ngs wrapper for Diamond [currently uses](https://github.com/broadinstitute/viral-ngs/blob/master/tools/diamond.py#L17) v0.7.9, while the Diamond package [on bioconda](https://github.com/bioconda/bioconda-recipes/tree/master/recipes/diamond) is v0.7.10. Can we use version v0.7.10 and add a `CondaPackage()` install method to our wrapper?

dpark01:
Done

PR #287, opened 2016-05-02
Update diamond conda version to 0.7.10

yesimon:
To address #279

### DIFF
--- a/tools/diamond.py
+++ b/tools/diamond.py
@@ -15,6 +15,7 @@ import util.misc
 
 URL = 'https://github.com/bbuchfink/diamond/archive/b576a5c03177603554f4627ae367f7bcbc6b8dcb.zip'
 TOOL_VERSION = '0.7.9'
+CONDA_VERSION = tools.CondaPackageVersion('0.7.10', 'boost1.60_1')
 DIAMOND_COMMIT_DIR = 'diamond-b576a5c03177603554f4627ae367f7bcbc6b8dcb'
 DIAMOND_DIR = 'diamond-{}'.format(TOOL_VERSION)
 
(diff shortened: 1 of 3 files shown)


tools/diamond.py
```diff
@@ -27,7 +28,9 @@ class Diamond(tools.Tool):
 
     def __init__(self, install_methods=None):
         if not install_methods:
-            install_methods = [DownloadAndBuildDiamond(URL, os.path.join(DIAMOND_DIR, 'bin', 'diamond'))]
+            install_methods = [
+                DownloadAndBuildDiamond(URL, os.path.join(DIAMOND_DIR, 'bin', 'diamond')),
+                tools.CondaPackage("diamond", version=CONDA_VERSION)]
```
dpark01: Is there a reason that Conda should only be tried if download-and-build fails? Most tools specify these installers in the opposite order for the sake of speed-optimization.

yesimon: You are right. This is not ready yet because the `CondaPackageVersion` spec is in another PR.

dpark01:
Just merged in the other PR. Rebase on master and flip the tool installer order?

yesimon:
Lots of conda commands return nonzero exit codes which were already handled previously so we shouldn't use `check=True`.

dpark01:
Ok, go ahead and merge this asap once it passes. Then tag v1.4.1.

dpark01:
This is a bit disconcerting. Not sure why the shell script returns exit code 1 when invoked with subprocess.run on Py3.5 but appears to exit 0 with subprocess.check_output on 3.4 and 2.7. Not sure who's right / which one is the bug.

I'm inclined to harmonize it by removing the try: from subprocess import run statement and always use your custom run method even if subprocess.run exists. While that feels backwards, the real goal is not to re-implement something from 3.5 accurately, but to create a method that works in our own use cases consistently.

PR #289, opened 2016-05-03
oops: dont check=True if we are checking return codes anyway

dpark01:


dpark01:
defer to #287 for expediency

PR #291, opened 2016-05-03
Test deplete_human

tomkinsc:
add a test for deplete_human, along with a few helper functions in the
blast, bmtagger, and last tool wrappers for building databases. The db
build helper functions have not yet replaced one-off builds elsewhere
in the codebase, nor been exposed as argparse commands (where should
they live if they are exposed?)


dpark01:
Great. The db build helper commands can go in taxon_filter.py. Utilizing those helper functions elsewhere would be nice at some point. I'm putting #287 ahead of this one since it contains fixes for broken master.

tomkinsc:
Ok, taxon_filter it is. The db build functions have some duplication that can be removed, but I'll get to that.

dpark01:
Not sure if you're getting caught up on assert_equal_contents on BAM files, but that can be a bit of a dangerous / non-deterministic game to play. I might make a new test function called assert_equal_bam_reads or something that basically dumps BAM -> SAM and removes the headers (the most non-deterministic part) and asserts equal reads. Or you could skip all that and assertEqual on samtools.count(bamfile) just to keep it simple.

tomkinsc:
Yeah, I feared that might be an issue, but was hoping the input would be small enough that it would be deterministic. I'll change the comparisons.

dpark01:
This looks good to merge for me. Is it good on your end?

tomkinsc:
Yeah, should be fine to merge.

PR #292, opened 2016-05-03
Cache pip on travis

yesimon:


dpark01:
woah... hold on... let me dig through the git history, but pip was always cached from the very beginning, not sure how it got removed.

dpark01:
Ok, pip is being cached, in that we're caching the virtualenv directory, and as far as I can tell, that means pip isn't re-installing anything when the cache exists. Do you know what this travis directive adds on top of that?

yesimon:
Hmm well the Travis build logs previous to this seem to have segments of
around 200s where pip packages were being downloaded, but only once per
build?

yesimon:
Actually the cache seems to be working most of the time, except once or twice like https://travis-ci.org/broadinstitute/viral-ngs/jobs/127405945. Nevertheless the pip cache dir seems to be `/home/travis/.cache/pip` so it might be an orthogonal cache?

dpark01:
If you look at that same build log you referenced, you'll find that this was a build where an appropriate cache for this build did not exist (see line 608 of the build log). Travis has an interesting branch-specific caching mechanism based on what a particular commit is a child of, and if it isn't resolvable, it defaults to no cache, which means doing everything from scratch. Sometimes if we merge in something to master with weird rebasing in its history, this will happen. Sometimes we manually clear the cache on purpose. Tagged releases will intentionally nuke the tools/build directory from the cache but not pip. Generally speaking, I don't mind it when the cache gets blown away on master--it exists mostly to speed up the build cycle on branch commits and pull requests to master (though the latter requires a cache on master).

At one point in history, I did cache the pip-cache, but then [stopped doing it](https://github.com/broadinstitute/viral-ngs/commit/9380c6ea935a8b236c9ff34b515259aebb17c015#diff-354f30a63fb0907d4ad57269548329e3). As far as I can tell, if we're caching the virtualenv, adding the pip cache just adds unnecessary payload to the travis cache which slows down the retrieval of the cache. I *think* that the pip cache only contains downloaded eggs, and downloading eggs isn't the slow part. The slow part is compiling C code during the pip install process of each egg, and I think that's not in the pip cache but I might be wrong.

yesimon:
That makes sense on doubling the cache size! I'll close this.

PR #293, opened 2016-05-03
Blast LCA implementation

yesimon:
* Pushing up taxonomy counts to hit a certain support level
* Producing kraken style reports from tsv files.


tomkinsc:
Can the py.test change be split out into a separate PR, or is it needed for this one?

yesimon:
It is needed because I am using pytest fixtures.

dpark01:
Ok let's hold on this one for now

dpark01:
At some point, after we deal with the py.test PR, you can rebase/merge that back into here, so no need to mess with this branch too much at the moment. But one thing: can you move the contents of taxonomy.py to metagenomics.py? Unless those methods are used elsewhere, in which case you can move it to util/ but in a cursory look, it only gets used in metagenomics.py.

dpark01:
Move contents of taxonomy.py to metagenomics.py?

PR #296, opened 2016-05-04
change argparse help text

tomkinsc:
define a custom argparse help action to be called at the top parser
level to print a column of the various subcommands for a file along
with their description strings, then print the standard *.py usage info

screen shot 2016-05-04 at 2 01 05 pm

### DIFF
--- a/reports.py
+++ b/reports.py
@@ -1,5 +1,5 @@
 #!/usr/bin/env python
-''' Reports
+''' Functions to create reports from genomics pipeline data.
 '''
 
 __author__ = "dpark@broadinstitute.org"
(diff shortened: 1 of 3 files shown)


dpark01:
Looks cool. I just triggered [a manual build of this branch ](http://viral-ngs.readthedocs.io/en/change_argparse_help_format/)on readthedocs.org.  Double check that it looks okay?

tomkinsc:
The build for this branch on readthedocs.org looks fine. There shouldn't be any changes there apart from description strings that did not previously exist, since this PR only impacts the output of `./*.py --help` (where `--help` is still implied if the file is called without arguments).

PR #298, opened 2016-05-04
Add missing threads param to blastn call

tomkinsc:


dpark01:
Hm, I don't really like mixing things in PRs. #296 looks like it's almost done building, let's merge that first, then #298 after that, and then #286 once that's fixed.

PR #300, opened 2016-05-05
Convert to kraken integration tests to fixtures 

yesimon:
Cleaning up the spaghetti


dpark01:
Looks alright.. one point of confusion for me though: why are the tests utilizing fastq inputs? We don't have any methods that utilize fastq inputs...

yesimon:
Those tests for for the original "integration" tests which interfaced with the tool at a lower level (no metagenomics.py), but had the real input files/db.

PR #301, opened 2016-05-12
Diamond integration test with pytest

yesimon:
Also rename test input folder to reflect multiple usages.


tools/diamond.py
```diff
@@ -99,7 +99,10 @@ def execute(self, command, options=None, option_string=None, return_stdout=False
         if option_string:
             cmd.extend(shlex.split(option_string))
         log.debug("Calling {}: {}".format(command, " ".join(cmd)))
-        return util.misc.run(cmd, stdout=subprocess.PIPE, stderr=subprocess.STDOUT)
+        res = util.misc.run_and_print(cmd)
```
dpark01: two thoughts:

1. why run_and_print and not subprocess.check_call?
2. if run_and_print, how about log.error(res.stdout.decode('utf-8')) instead of print()?

pipes/rules/metagenomics.rules
```diff
@@ -29,16 +30,29 @@ rule kraken:
...
-        """
-        {config[bin_dir]}/metagenomics.py krona {input} {config[krona_db]} {output} --noRank
-        """
+method_reads_ext = {
+    'diamond': 'diamond.lca.gz',
+    'kraken': 'kraken.reads.gz'
+}
 
+method_krona_columns = {
+    'diamond':  (1, 2),
```
dpark01: I hate to say this, but I think I'd prefer the diamond LCA output to throw in a blank column in front of each row, so that the format can _exactly_ match the kraken text output format (at least as far as krona is concerned). That should only add one byte per read to the uncompressed text file, but should make the output files far more interchangeable.

pipes/rules/metagenomics.rules
```diff
@@ -29,16 +30,29 @@ rule kraken:
...
+    input: lambda wildcards: join(config["data_dir"], config["subdirs"]["metagenomics"], "{sample}.%s" % method_reads_ext[wildcards.method])
+    output: join(config["data_dir"], config["subdirs"]["metagenomics"], "{sample,\w+}.{method,\w+}.krona.html")
+    run:
+        query_col, taxid_col = method_krona_columns[method]
+        shell("""
+        {config[bin_dir]}/metagenomics.py krona {input} {config[krona_db]} {output} --noRank \
+          --queryColumn {query_col}  --taxidColumn {taxid_col}
+        """)
 
 rule pair_fastq:
```
dpark01: probably delete snakemake rule pair_fastq?

pipes/rules/metagenomics.rules
```diff
@@ -29,16 +30,29 @@ rule kraken:
     resources: cores=int(config.get("number_of_threads", 1))
     params: numThreads=str(config.get("number_of_threads", 1))
     run:
+        shell("pyenv which python")
```
dpark01: are these debugging lines? I'd propose removing them, unless we want to introduce them to more of the rules. But actually, I think util.cmd starts off each new execution of a script with the python version. We could probably have it also report the path-to-python as well since that would be nice debugging info.

dpark01:
Excellent, I like this. Just a few small comments already made in-line.

pipes/rules/metagenomics.rules
yesimon: I changed this slightly but keeping it because the output formats are not really compatible - the kraken output has extra columns whose meaning cannot be translated for diamond.

PR #322, opened 2016-05-23
removed custom install methods for kraken, added decorator to skip install test for Kraken on OSX. added package_available() to CondaPackage() to search for conda package

tomkinsc:
removed custom install methods for kraken (+jellyfish, yaggo), relying instead on “kraken-all” from bioconda; added decorator to skip install test for Kraken on OSX. added package_available() to CondaPackage() to assess availability of package on channel


tools/__init__.py
```diff
@@ -423,6 +441,9 @@ def uninstall_package(self):
         self.verify_install()
 
     def install_package(self):
+        if not self.package_available():
```
yesimon: log here?

dpark01:
@yesimon, can you code review, request changes, and merge this PR when it looks good to you?

tomkinsc:
There will be more revisions before this is ready.

dpark01:
Not sure if that "thread count exceeds number of processors" is a warning or an error/cause-of-failure...

yesimon:
I think it's because the default 'config.yaml' has `threads: 8`, and the snakemake integration test executor copies that config. Either we can override it in the snakemake integration test runner or provide a separate `config.yaml` for integration tests.

tools/__init__.py
```diff
@@ -400,6 +400,26 @@ def get_installed_version(self):
...
+        # If we ever use conda to install pip packages as tools, "-c" needs to be removed
+        run_cmd = ["conda", "search", "--json", "-c", self.channel, self.package]
+
+        result = util.misc.run_and_print(run_cmd, silent=True, env=self.conda_env)
+        if result.returncode == 0:
+            try:
+                command_output = result.stdout.decode("UTF-8")
+                data = json.loads(self._string_from_start_of_json(command_output))
+            except:
+                _log.warning("failed to decode JSON output from conda create: %s", result.stdout.decode("UTF-8"))
```
yesimon: This isn't conda create

tools/__init__.py
```diff
@@ -400,6 +400,26 @@ def get_installed_version(self):
                     return CondaPackageVersion(installed_version, installed_build_type)
         return None
 
+    def package_available(self):
+        # If we ever use conda to install pip packages as tools, "-c" needs to be removed
+        run_cmd = ["conda", "search", "--json", "-c", self.channel, self.package]
+
+        result = util.misc.run_and_print(run_cmd, silent=True, env=self.conda_env)
```
yesimon: `loglevel=logging.DEBUG` instead of `silent=True`?

tools/kraken.py
```diff
@@ -174,23 +106,7 @@ def execute(self, command, db, output, args=None, options=None,
...
-        util.file.mkdir_p(bin_dir)
-        for bin_name in Kraken.BINS:
-            libexec_bin = os.path.join(libexec_dir, bin_name)
-            bin = os.path.join(bin_dir, bin_name)
-            if not os.path.islink(bin):
-                os.symlink(libexec_bin, bin)
+@tools.skip_install_test(condition=tools.is_osx())
+class Jellyfish(Kraken):
+    """ Tool wrapper for Jellyfish (installed by kraken-all metapackage) """
+    subtool_name = 'jellyfish'
```
yesimon: Add newline

dpark01:
my take would be to override it for the integration test only.

Actually, given this hard failure mode, my real preference might be to simply have the default for this be unspecified at the command line, with argparse defaulting to None, which would pass nothing to the subprocess call, or alternatively autodetect the core count.

tools/__init__.py
```diff
@@ -400,6 +400,26 @@ def get_installed_version(self):
...
+            except:
+                _log.warning("failed to decode JSON output from conda create: %s", result.stdout.decode("UTF-8"))
+                return # return rather than raise so we can fall back to the next install method
+
+            if data and len(data):
+                if self.package in data and "error" not in data:
+                    for sub_pkg in data[self.package]:
+                        if sub_pkg.get("version", "") == self.version.version_spec:
+                            return True
+        return False
```
yesimon: Log data if we're returning False

tools/__init__.py
dpark01: how about logging.WARNING instead?

util/misc.py
```diff
@@ -353,3 +354,109 @@ def get_intervals(self, c=None):
                 right = right - 1
                 features = list(self.get_features(c, left, right))
                 yield (c, left, right, len(features), features)
+
+
+def available_cpu_count():
```
yesimon: I don't think we support most of these platforms, plus this requires installing psutil. The snakemake version should be okay and is a bit conservative about different ways cpu counts are revealed to the process under different VM schemes: https://bitbucket.org/snakemake/snakemake/src/dbfd288e71e494f8964e4f1302f2e66f3b6b1eb6/snakemake/utils.py?at=master&fileviewer=file-view-default#utils.py-265

tomkinsc: It's a heuristic. psutil is not *required*; it's one option attempted. I can simplify it to use the Snakemake version; it would be less to maintain.

tools/__init__.py
```diff
@@ -337,9 +337,9 @@ def verify_install(self):
     def _attempt_install(self):
         try:
             # check for presence of conda command
-            util.misc.run_and_print(["conda", "-V"], silent=True, check=True, env=self.conda_env)
+            util.misc.run_and_print(["conda", "-V"], loglevel=logging.WARNING, check=True, env=self.conda_env)
```
yesimon: `run_and_print` logs at that level regardless of success or failure, so `logging.WARNING` might be a too verbose. `logging.INFO` could be good to differentiate between that and `DEBUG` level logging.

tomkinsc:
OK to merge?

yesimon:
Should be good!

Issue #314, opened 2016-05-18
create metagenomic report merger


dpark01:
Create a script that combines metagenomic reports (classified read .txt.gz files) from multiple Illumina lanes. It should also regenerate the summary report file and produce output that is easily consumed by our krona wrapper. End goal is a DNAnexus applet.

tomkinsc:
#342 has been merged in, so closing this for now

PR #342, opened 2016-06-14
add metagenomic_report_merge()

tomkinsc:
This addresses #314, and adds a function, `metagenomic_report_merge()`,  to `metagenomics.py` to combine multiple metagenomic reports into a single file usable as Krona input, with an option to recreate the Kraken summary (db required).

### DIFF
--- a/metagenomics.py
+++ b/metagenomics.py
@@ -25,6 +25,7 @@ import tools.kraken
 import tools.krona
 import tools.diamond
 import tools.picard
+import csv
 
 __commands__ = []
 
@@ -410,6 +411,10 @@ def kraken_dfs(db, lines, taxa_hits, total_hits, taxid, level):
 
 def kraken(inBam, db, outReport=None, outReads=None,
            filterThreshold=None, numThreads=1):
+    '''
+        Classify reads by taxon using Kraken
+    '''
+
     assert outReads or outReport, (
         'Either --outReads or --outReport must be specified.')
     kraken_tool = tools.kraken.Kraken()
@@ -437,24 +442,6 @@ def kraken(inBam, db, outReport=None, outReads=None,
         kraken_tool.report(tmp_filtered_reads, db, outReport)
 
 
-def krona(inTsv, db, outHtml, queryColumn=None, taxidColumn=None,
-          scoreColumn=None, noHits=None, noRank=None):
-
-    krona_tool = tools.krona.Krona()
-    if inTsv.endswith('.gz'):
-        tmp_tsv = util.file.mkstempfname('.tsv')
-        with gzip.open(inTsv, 'rb') as f_in:
-            with open(tmp_tsv, 'wb') as f_out:
-                shutil.copyfileobj(f_in, f_out)
-                to_import = [tmp_tsv]
-    else:
-        to_import = [inTsv]
-
-    krona_tool.import_taxonomy(
-        db, to_import, outHtml, query_column=queryColumn, taxid_column=taxidColumn,
-        score_column=scoreColumn, no_hits=noHits, no_rank=noRank)
-
-
 def parser_kraken(parser=argparse.ArgumentParser()):
     parser.add_argument('inBam', help='Input unaligned reads, BAM format.')
     parser.add_argument('db', help='Kraken database directory.')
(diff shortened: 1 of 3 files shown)


tomkinsc:
(This is a WIP)

dpark01:
Hm, since you're starting to deal with the diamond output (I haven't touched any of this yet), I was going to say that one thing I'd really like to do is modify the current diamond entry point to throw a blank/empty column in front of the whole output file, so that the format actually matches a format that we want (not the format the tool gives us), similar to how much muscle we put into forcing vphaser to conform to our desired input/output file formats.  Then we can drop all this stuff about specifying column numbers and the user having to know whether a report was generated with one tool or another (which really kills the usability/portability if they have to keep track of that kind of thing).

tomkinsc:
We could modify the Diamond output; it would certainly simplify our code and calls. One argument against is that we'd be increasing the report size (due to extra delimiters or column info) without adding useful information. For a MiSeq run it wouldn't be much, but for HiSeq runs, it could add up. Compression would help, of course. I'm also not sure (yet) how the krona tool handles empty columns—whether it will honor empty columns as it should, or whether it will treat multiple delimiters as single delimiters. Another concern is whether we should worry about users calling Diamond from our `metagenomics.py` wrapper and getting an output format that's different from what they may expect from Diamond. Our modified format could be optional. I guess I'll add a flag for it to the parser, and see how Krona deals with empty columns. If it all checks out we can modify our calls to include the flag and strip out our column specification.

I can say that Krona works fine on two-column files, and the Krona input part of the merger is functional.

dpark01:
I'm much less concerned about simplifying our code and more concerned about simplifying usability from the user side. Our use cases will always compress the raw read-based output anyway, and an extra byte per read (the tab character) should compress nicely. Do need to double check whether kraken-report and krona handle a blank column okay or whether we have to put a dummy value there.

dpark01:
Oh also, our diamond wrapper already emits something that is very different (both format and semantically) from what comes out of raw diamond... there's a whole bunch of custom LCA code after the raw diamond call to reinterpret the output. These command line utilities are less about replicating functionality of some baseline tool (no one needs us for that.. just use bioconda). It's about performing a functional task of some kind, using some tool underneath, but hiding away the ugly interactions with that tool so people don't have to think about all the file conversions and peculiarities and foibles of each program.

The current output file format from metagenomics.kraken happens to match kraken's native output format, but it's not on purpose, it's by coincidence. I asked Simon to pick a generic metagenomic output format that should be portable across all tools, and he picked the kraken formats because they were pretty good and conceptually universal.

tomkinsc:
Ok, the merger interface has been simplified to accept metagenomic reports in Kraken format without separate arguments for kraken/diamond reports, and the Diamond output now prepends an extra column (`\t`) so the columns align with Kraken format.

metagenomics.py
```diff
@@ -547,10 +560,65 @@ def parser_diamond(parser=argparse.ArgumentParser()):
...
+        
+        util.file.cat(tmp_metag_combined_txt, [metag_file.name for metag_file in metagenomic_reports])
+
+        kraken_tool = tools.kraken.Kraken()
+        kraken_tool.report(tmp_metag_combined_txt, kraken_db.name, out_kraken_summary)
+  
+
+def parser_metagenomic_report_merge(parser=argparse.ArgumentParser()):
+    parser.add_argument("metagenomic_reports", help="Input metagenomic reports created by Kraken", nargs='+', type=argparse.FileType('r'))
+    parser.add_argument("--outKrakenSummary", dest="out_kraken_summary", help="Input metagenomic reports created by Diamond") #, type=argparse.FileType('w'))
```
dpark01: A few questions about the argparse signature on metagenomic_report_merge:
 - outKrakenSummary has a help signature that says "Input metagenomic reports created by Diamond"
 - krakenDB has nargs='+' which is weird
 - can we call the outputs outSummaryReport and outByReads or something like that? remove the vendor-specific implications of what these are good for (for example, I'd hope that metagenomic_report_merge could take as input files that were created by metagenomic_report_merge --outKronaInput)

tomkinsc: My fault. I'll fix these.

PR #343, opened 2016-06-15
Fix metagenomics input and provide more resources

haydenm:
The metagenomics tools took as input the file "{sample}.bam" in the
source subdir, but this file is not created when demultiplexing raw
data and the snakemake pipeline is not setup to create this file.
Thus, snakemake is unable to find input for the metagenomics tools.
This fix changes the input to be "{sample}.raw.bam" in the per_sample
subdir. The snakemake pipeline is able to create this file even
when demultiplexing raw reads, so it is able to create input for
the metagenomics tasks.

This also gives the metagenomics tools more resources; it requests
more memory and a queue with longer runtime.


dpark01:
Thanks @haydenm ! It makes sense you'd have to update the test case to match your changes, just ask @yesimon if you need assistance.

tomkinsc:
Could use some help here, @yesimon.

dpark01:
Hm, looks like it's still not putting the test bam in the 01_per_sample directory for the snakemake test. I can take a look after the weekend (I don't think it's so urgent that we have to get this sorted before then).

dpark01:
perfect!

PR #350, opened 2016-06-17
update krona to 2.6.1

tomkinsc:


yesimon:
2.6.1 is still going to require the patch from before

tomkinsc:
I'll let you update it, since you know more about this that I do.

PR #351, opened 2016-06-17
Add qsub -R flag for mem >= 15G

yesimon:
Adds advance reservation flag for larger jobs.


yesimon:
I should double check the node map to make sure this is appropriate

dpark01:
@yesimon have you double checked what you need to on the UGER environment?

PR #354, opened 2016-06-18
Tools simplification

dpark01:
- Revert samtools.faidx to call samtools instead of pysam (fix transient errors on travis py3.5).
- Ease debuggability of diamond (partly to help debug current failures on dnanexus CI) and a simplify a bunch of code by reverting all simple run_and_print calls to subprocess.check_call (any scenario where silent=False, check=True, and no output is captured, no env is specified, etc).
- Remove a few non-conda installers for some tools.

### DIFF
--- a/requirements-tests.txt
+++ b/requirements-tests.txt
@@ -1,5 +1,6 @@
 coveralls==1.1
 flake8<=3
+pycodestyle
 mock==2.0.0
 six<2
 pytest-xdist==1.14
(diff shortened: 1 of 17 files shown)


dpark01:
This test failure should be an easy fix (just change the mock target), I'll get to it once I'm back in front of a computer. I'm hoping to get this in to help debug some issues Yifei is having on their CI.

dpark01:
To do: add pycodestyle to requirements-test.txt before mock. Mock depends on it, but pypi fails to say so, and Travis happens to have it so we never notice.

PR #356, opened 2016-06-19
Add read depth to merged intrahost variation output

haydenm:
This adds the 'DP' field to the VCF file created at the end of
the intrahost component of the pipeline. The field gives, at
each position, the read depth for each sample. Together with
the allele frequency ('AF'), this can provide allele counts at
each position in each sample. This can be useful, e.g., for
statistical tests that compare the abundance of an iSNV between
samples.

### DIFF
--- a/intrahost.py
+++ b/intrahost.py
@@ -474,7 +474,7 @@ def merge_to_vcf(
         outf.write('##fileformat=VCFv4.1\n')
         outf.write('##FORMAT=<ID=GT,Number=1,Type=String,Description="Genotype">\n')
         outf.write('##FORMAT=<ID=AF,Number=A,Type=Float,Description="Allele Frequency">\n')
-        outf.write('##FORMAT=<ID=DP,Number=1,Type=Integer,Description="Read depth at position">\n')
+        outf.write('##FORMAT=<ID=DP,Number=1,Type=Integer,Description="Read Depth">\n')
         outf.write('##FORMAT=<ID=NL,Number=R,Type=Integer,Description="Number of libraries observed per allele">\n')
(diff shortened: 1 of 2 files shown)


intrahost.py
```diff
@@ -794,6 +797,8 @@ def merge_to_vcf(
                     # AF col emitted below, everything excluding the ref allele (one float per non-ref allele)
                     freqs = [(s in iSNVs) and ','.join(map(str, [iSNVs[s].get(a, 0.0) for a in alleles[1:]])) or '.'
                              for s in samplesToUse]
+                    # DP col emitted below
+                    depths = [str(iSNVs_read_depth.get(s, '.')) for s in samples]
```
dpark01: Should be:
```for s in samplesToUse```
(like freqs and nlibs). Some samples get tossed in the filtering process above.

File: intrahost.py [hunk 2]
```diff
@@ -474,6 +474,7 @@ def merge_to_vcf(
         outf.write('##fileformat=VCFv4.1\n')
         outf.write('##FORMAT=<ID=GT,Number=1,Type=String,Description="Genotype">\n')
         outf.write('##FORMAT=<ID=AF,Number=A,Type=Float,Description="Allele Frequency">\n')
+        outf.write('##FORMAT=<ID=DP,Number=1,Type=Integer,Description="Read depth at position">\n')
```
dpark01: Description should be "Read Depth" in keeping with [VCF fixed vocabulary](http://samtools.github.io/hts-specs/VCFv4.1.pdf).

dpark01:
Otherwise, looks great!

Issue #290, opened 2016-05-03
should conda installs be travis-cached?


dpark01:
Travis caching was primarily meant to speed up the build for the tools that required compilation (and also for pip installs that involved compiling C code e.g. numpy). But does it provide speedup for conda-installed tools? If not, we may want to avoid caching conda so that the installers actually get tested more often than the tagged releases.

yesimon:
It seems like conda tools installation caching is causing a bunch of coverage noise `+/-` (depending on if everything is cached or not - if not that requires executing more code), and it's leading to spurious increases/decreases in coverage percentage.

tomkinsc:
It's not much of a performance advantage to cache conda installs, especially considering the time it takes to add them to the cache at the end of a build. We can probably omit caching of conda installs, or perhaps travis cache the `conda-cache/` directory to include any local package archives.

tomkinsc:
We're not currently caching installed conda tools, but rather [only](https://github.com/broadinstitute/viral-ngs/blob/31d8e5ae993444dbb2979eb35440ad45a900d0b0/.travis.yml#L5) the miniconda directory since it can take take some time to download and unpack/install.

PR #359, opened 2016-06-21
build optimizations

dpark01:
- remove a lot of non-conda-based fallback tool installers
- remove Travis caching of conda-installed tools. also keeps coveralls test coverage consistent.
- add standalone install_tools.py script that installs tools w/o the need for requirements-tests.txt
- remove pytest.ini and move into Travis-specific env variables so that developers can specify their own options appropriate to their local dev environment
- run flake8 only on pull requests and master branch (skip branch commits)
- add a requirements-conda.txt file that can be used with conda create to optimize/speed up tool installation


dpark01:
closes #290

PR #360, opened 2016-06-21
a few landscape-inspired changes

tomkinsc:
Now that landscape is working again upstream, we can start to make
changes/improvements again based on its reports (see:
https://github.com/landscapeio/landscape-issues/issues/220 )


dpark01:
Looks good, and good to hear that landscape is working again. I think we should delete flake8 from our Travis setup now, if Landscape.io is already running pyflakes checks... shave a minute off the Travis build.

test/unit/test_assembly.py
```diff
@@ -139,8 +139,7 @@ def test_lassa_protein(self):
         assembly.order_and_orient(
             os.path.join(inDir, 'contigs.lasv.fasta'),
             os.path.join(inDir, 'ref.lasv.fasta'),
-            outFasta,
-            aligner='promer')
+            outFasta)
```
dpark01: this is actually incorrect, but maybe we should just delete the test_lassa_protein unit test anyway since assembly.order_and_orient has not had this functionality for a while, and the unit test has just been skip-decorated this whole time.

test/unit/test_interhost.py
```diff
@@ -84,7 +84,7 @@ def test_unequal_genomes_error(self):
         genomeA = makeTempFasta([('chr1', 'ATGCACGTACGTATGCAAATCGG'), ('chr2', 'AGTCGGTTTTCAG'),])
         genomeB = makeTempFasta([('first_chrom', 'GCACGTACGTATTTGCAAATC')])
         with self.assertRaises(Exception):
-            cm = interhost.CoordMapper(genomeA, genomeB)
+            cm = interhost.CoordMapper2Seqs(genomeA, genomeB)
```
dpark01: hm... maybe we should be raising a more specific exception than just Exception....

test/unit/test_assembly.py
tomkinsc: Do you mean my change is incorrect, or the test prior to the change? In `master`, `aligner` is commented out from the `order_and_orient` signature as of [this](https://github.com/broadinstitute/viral-ngs/commit/090bf5c9f89f75ae96affc32b514cbbcba5d1bf6) commit. I'll remove the test if we don't need it.

test/unit/test_interhost.py
tomkinsc: Yeah. I'll work on this. Turns out Landscape can be helpful!

test/unit/test_assembly.py
dpark01: The change to the test is incorrect: it should be aligner=promer. But as you've noted, the current implementation no longer supports this option so the test is commented out. Why don't we leave this as it originally was and leave it skipped for some eventual day when maybe we reimplement the feature.

tomkinsc: Ok!

PR #364, opened 2016-06-21
Fix JSON serializability bug

haydenm:
When passing information into params, snakemake requires that it be
JSON serializable. However, `read_samples_file` is a generator and
is not serializable. snakemake fails with `TypeError: <generator
object read_samples_file at ... is not JSON serializable>`. This
fix wraps the call to `read_samples_file` with list(..) to obtain
the output of the generator, which can be serialized.


haydenm:
It looks like this created downstream intrahost problems, even though all tests pass. We might actually want `' '.join(read_samples_file(..))` rather than `list(read_samples_file(..))`. I'll take a closer look later.

haydenm:
c0519b8 resolves the problem I was seeing.

dpark01:
Ugh, this appears to be a change in snakemake's behavior (it appears that it treats params differently from inputs when inserting them into a string in a run/script block). Well, this is the more conservative/defensive way to code it, so let's run with this.

PR #370, opened 2016-06-23
change num_threads=0 for IlluminaBasecallsConverter

tomkinsc:
pass num_threads=0 to picard’s IlluminaBasecallsToSam so that
IlluminaBasecallsConverter uses all available cores rather than the 4
cores previously hard-coded

### DIFF
--- a/tools/picard.py
+++ b/tools/picard.py
@@ -292,7 +292,7 @@ class IlluminaBasecallsToSamTool(PicardTools):
         'adapters_to_check': ('PAIRED_END', 'NEXTERA_V1', 'NEXTERA_V2'),
         'max_reads_in_ram_per_tile': 100000,
         'max_records_in_ram': 100000,
-        'num_processors': 4,
+        'num_processors': 0, # 0 = all available
         'force_gc': False,
         'include_non_pf_reads': False,
     }


tomkinsc:
@dpark01: Should we have a constraint on core count? Based on [this commit](https://github.com/broadinstitute/viral-ngs/commit/7db18408e85ce1f4b5a827596a1cf166f3b57c10) it looks like you reduced the count to decrease the memory footprint. Should we have a better heuristic to set core count based on available RAM as well as available processors?

dpark01:
I think on almost every tool, we should let it autoscale to the number of processors. But in the case of Picard demultiplexing, we have to fix it to a small number, or the memory footprint gets ridiculous. It's really tool-by-tool... some tools don't scale memory usage much based on thread count, others do...

dpark01:
I think I'd prefer to close this PR

dpark01:
Or it's probably right to expose the parameter, as long as we make sure it defaults the same way as before...

I think most tools don't have this problem, and I've wondered why we have a threads parameter in our config.yaml instead of just letting things autoscale naturally to available hardware. Maybe we do need to revisit our threads strategy in general.

PR #374, opened 2016-06-23
OSX compatibility

dpark01:
This makes the test suite pass on OSX, mostly by skipping a few of the ones that depend on Kraken, Diamond, or VPhaser2. It prepares some aspects of Travis testing, but does not actually enable them because of slow boot-up and build times in their current infrastructure.


tomkinsc:
I suppose this means I can delete my `enable-osx-builds` branch from a while ago? It was mostly [working](https://travis-ci.org/broadinstitute/viral-ngs/builds/118000377), albeit with 3x longer test time and often significant delays in the Travis queue waiting for OSX build slots to be available.

dpark01:
Yeah, I copied a lot of snippets straight out of your branch actually. I mostly got mine working, with the exception of coveralls I think... but I also noticed that the queue wait times and actual build times meant that I really didn't want to wait around for OSX builds on the average PR for this project. But I figured we might as well make it so that devs can successfully py.test everything on their local machines and start putting some of the travis infrastructure in place.

PR #375, opened 2016-06-24
join accessions list into space-separated string

tomkinsc:
fairly sure this was a non-issue before. bug introduced by a newer version of Snakemake?


dpark01:
Yeah I think we're seeing this in multiple places.. Fix as we find them!

PR #380, opened 2016-06-28
Fix various issues in interhost rules

haydenm:
This fixes issues in the interhost rules that prevented targets
from running. Now, the `all_ref_guided` target runs successfully,
along with all the others it invokes.

The fix in `assembly.py` is needed for segmented genomes. Without
this fix, the fasta file would give the same header for each segment
and downstream steps would fail because of this.

### DIFF
--- a/assembly.py
+++ b/assembly.py
@@ -1062,12 +1062,13 @@ def main_vcf_to_fasta(args):
 
     with open(args.outFasta, 'wt') as outf:
         chr_idx = 0
-        for header, seq in vcf_to_seqs(util.file.read_tabfile(args.inVcf),
-                                       chrlens,
-                                       samples,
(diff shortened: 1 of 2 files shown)


dpark01:
I'm looking through the changes to assembly.py now. The snakemake rule changes look great. The snakemake changes regarding the custom unrolling of lists in params is an unfortunate behavioral change in snakemake itself, thanks for making that change. More comments to follow soon.. (if any)

dpark01:
Oops, that change to assembly.py looks like a straight up bugfix--there was some code that was set up to use chr_idx, but it never actually counted up.  Thanks. I'm going to merge this now, at some future time maybe we should consider a multi-segment test coverage of the fasta header naming that comes out of this method... when we have time of course...

PR #383, opened 2016-06-28
SampleSheet reader now uses csv.reader() rather than str.split()

tomkinsc:
Previously, commas in sample names could cause problems when reading in sample sheets, since the reader was naively splitting on commas. This incorporates changes to uses Python’s csv.reader() to parse SampleSheet.csv rows, with a custom dialect to support escaping quotes with backslash characters. It also adds a unit test with a SampleSheet that was problematic prior to these changes.


dpark01:
Thanks!

dpark01:
Merge in some of my branch, I got started a bit

--
Daniel J. Park, PhD
Tel: +1-617-714-8526
dpark@broadinstitute.org

tomkinsc:
Ok, I merged in what you had in `dp-samplesheet`, and removed some duplication since I'm already stripping whitespace from all elements in a row.

tomkinsc:
This should be OK to merge.

Issue #361, opened 2016-06-21
Add a convenience function for printing a list of the most common barcodes in a run in descending order


tomkinsc:
Add a convenience function for printing a list of the most common barcodes in a run in descending order, using barcodes extracted from raw reads by picard's `ExtractIlluminaBarcodes`. This can be useful for puzzling out which barcodes in a sample sheet may be incorrect.

PR #384, opened 2016-06-29
add common_barcodes() to illumina.py

tomkinsc:
Add a convenience function (with argparse) for writing a list of the most common barcodes in a run in descending order, using barcodes extracted from raw reads by picard's ExtractIlluminaBarcodes. This can be useful for puzzling out which barcodes in a sample sheet may be incorrect.

This addresses #361.


tomkinsc:
I'm open to suggestions for a better function name, but `common_barcodes` seemed to fit the use case.

PR #398, opened 2016-07-07
spike-in reports via BWA

dpark01:
Change spike-in reports to use BWA. Also change a few of the simple bam->fastq conversions in taxon_filter to use samtools instead of picard. Update samtools version. Add BWA tool.


dpark01:
@tomkinsc do we have to add anything anywhere when we change conda tool versions (samtools 1.2 to 1.3.1) or add new tools (bwa), or is that handled elsewhere?

tomkinsc:
Beyond the changes in this PR, we'll [need to change the version in the bioconda recipe](https://github.com/broadinstitute/viral-ngs/blob/master/DEVELOPMENT_NOTES.md#changing-dependency-versions) for viral-ngs, but that should happen after the next tagged release.

PR #399, opened 2016-07-08
common barcodes and other fixes

tomkinsc:
* —commonBarcodes is added to demux.rules invocation of illumina_demux
* better handling of rows with end-of-line spaces in samplesheets and barcode files
* common_barcodes now uses ExtractIlluminaBarcodesTool as source of jvmMemDefault 
* easy install now correctly prepends miniconda to PATH for environment loading

### DIFF
--- a/easy-deploy-script/easy-deploy-viral-ngs.sh
+++ b/easy-deploy-script/easy-deploy-viral-ngs.sh
@@ -106,6 +106,23 @@ else
     set_locale "en_US.utf8"
 fi
 
+function prepend_miniconda(){
+    if [ -d "$MINICONDA_PATH/bin" ]; then
+        echo "Miniconda installed."
+
+        echo "Prepending miniconda to PATH..."
+        export PATH="$MINICONDA_PATH/bin:$PATH"
+        hash -r
+
+        # update to the latest conda this way, since the shell script 
+        # is often months out of date
+        conda update -y conda
+    else
+        echo "Miniconda directory not found."
+        exit 1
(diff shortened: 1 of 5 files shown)


pipes/rules/common.rules
```diff
@@ -15,9 +15,15 @@ def set_env_vars():
 
 def read_tab_file(fname):
     with open(fname, 'rt') as inf:
-        header = [item.strip() for item in inf.readline().rstrip('\n').split('\t')]
+        header = [item.strip() for item in inf.readline().strip().rstrip('\n').split('\t')]
```
dpark01: let's see, while this new parsing of the header will lose any trailing empty columns (e.g. \t\t\t\n) in theory, the header is the one place where we shouldn't allow that anyway, so that's okay... trailing empty columns look good in the data rows below. It also handles trailing spaces within the interior columns itself. The csv case above is okay because the delimiter is not affected by strip()...

easy-deploy-script/easy-deploy-viral-ngs.sh
```diff
@@ -177,6 +186,8 @@ function create_project(){
 }
 
 function activate_env(){
+    prepend_miniconda
```
dpark01: This means that easy-deploy usages of our scripts are now pegged to miniconda's python, not the system python, right? That's probably a good thing.

tomkinsc: Exactly.

Issue #402, opened 2016-07-13
Add a convenience functions for creating coverage plots 


tomkinsc:
It would be helpful if viral-ngs could generate coverage plots. Two functions should be provided: one that accepts an aligned bam file and tabulates read depth (`plot_coverage()`?), and another (`align_and_plot_coverage()`?) that accepts an unaligned bam file or fastq files (converting where necessary), performs a quick alignment via`bwa mem`, and then calls the first function to calculate coverage.

The function `plot_coverage()` could use samtools' `sort`, `index`, and `depth` since they're fast, or it could use pysam. The former would include something like:

```
samtools sort -O bam -T sort mapped.bam > mapped_sorted.bam
samtools index mapped_sorted.bam
samtools depth -aa -m 1000000 mapped_sorted.bam > mapped_sorted_depth_summary.tsv
```

The function `plot_coverage()` should expose parameters to set the base quality and mapping thresholds (`-q` and `-Q`, respectively), and maximum coverage depth (`-m`). It should also expose parameters to set the plot type (png/pdf/txt), plot dimensions, title, and output path.

The function `align_and_plot_coverage()` should expose various bwa parameters such as min score threshold (`-T`). It could also expose a `--sensitive` option that uses @haydenm's param set, `-k 12 -A 1 -B 1 -O 1 -E 1 -T`, though that may mean [additional filtering is required](https://github.com/broadinstitute/hayden-work/blob/master/personal/bin/align_reads_to_fasta#L75).

Staying in Python space, if we're ok with including matplotlib as a dependency, we can use it to generate the coverage plots. Otherwise the most simple way to create plots would probably be `gnuplot`, which is available as a conda package. In addition to being able generating images or postscript plots, gnuplot supports a [fun text mode](http://www.calliopesounds.com/2011/12/i-have-to-say-again-i-find-gnuplots.html) suitable for terminal display, and an SVG mode suitable for display in a browser.

I'm not sure where `plot_coverage()` should live, but `read_utils.py` could make sense.

Once the functions have been added, the Snakemake rules should be updated to include them, possibly in `reports.rules`.

haydenm:
[...]

PR #403, opened 2016-07-14
add plot_coverage command

tomkinsc:
This addresses #402. Input about the visual display is welcome. As it is the matplotlib stylesheet can be set as an argparse param, and the data display type can be filled, line, or dots. If it would be helpful I could add a sliding-window average option.

12 1f coverage_plot4

12 1f coverage_plot4

12 1f coverage_plot4

12 1f coverage_plot4


tools/bwa.py
```diff
@@ -46,18 +48,28 @@ def index(self, inFasta, algorithm=None):
...
-        self.execute('mem', ['-t', str(threads), refDb, fq1, fq2], stdout=aln_sam)
+        self.execute('mem', opts + ['-t', str(threads), refDb, fq1, fq2], stdout=aln_sam)
         os.unlink(fq1)
         os.unlink(fq2)
-        samtools.sort(aln_sam, outAlign)
+        samtools.sort(aln_sam, aln_sam_sorted)
         os.unlink(aln_sam)
-        samtools.index(outAlign)
+        # cannot index sam files; only do so if a bam is desired
+        if outAlign.endswith(".bam"):
```
dpark01: Or endswith .cram

read_utils.py
```diff
@@ -1052,6 +1057,267 @@ def parser_bwamem_idxstats(parser=argparse.ArgumentParser()):
...
+    aln_sam_filtered = mkstempfname('.filtered.sam')
+
+    bwa.mem(in_bam, ref_indexed, aln_sam, opts=bwa_opts)
+
+    # @haydenm says:
+    # For some reason (particularly when the --sensitive option is on), bwa
+    # doesn't listen to its '-T' flag and outputs alignments with score less
+    # than the '-T 30' threshold. So filter these:
+    os.system("grep \"^@\" " + aln_sam + " > " + aln_sam_filtered)
+    os.system("grep \"AS:i:\" " + aln_sam + " | awk -v threshold=" + str(map_threshold) + " '{split($14, subfield, \":\"); if(subfield[3]>=threshold) print $0}' >> " + aln_sam_filtered)
```
dpark01: Why grep+awk on a sam file vs samtools view -f or -F on a bam file?

read_utils.py
```diff
@@ -1052,6 +1057,267 @@ def parser_bwamem_idxstats(parser=argparse.ArgumentParser()):
...
+    if out_summary is None:
+        coverage_tsv_file = mkstempfname('.summary.tsv')
+    else:
+        coverage_tsv_file = out_summary
+
+    bam_aligned = mkstempfname('.aligned.bam')
+
+    if in_bam[-4:] == ".sam":
+        # convert sam -> bam
+        samtools.view(["-b"], in_bam, bam_aligned)
```
dpark01: Samtools.sort can read a sam as input I thiught

read_utils.py
tomkinsc: Yeah, those lines can definitely be refactored. They were pulled from Hayden's known-working script. `samtools view` is definitely the answer.

read_utils.py
tomkinsc: `samtools sort` can accept sam, but `samtools index` expects `.bam` or `.cram`, and we need a sorted and indexed file for `samtools depth`, so input bam files get copied to tmp, and sam files get converted to bam format

dpark01: But you always need to sort and index, and since the samtools sort step takes anything as input (and produces bam output), the sam to bam conversion seems unnecessary

dpark01:
Can we move to reports.py?

read_utils.py
tomkinsc: True. I'll remove the sam->bam conversion and add `-O bam` to the `samtools.sort()` call.

tomkinsc:
Absolutely.

dpark01:
If this proves to be popular with folks, consider adding it to snakemake rules in a future PR

tomkinsc:
It may also be nice for the DNAnexus implementation via svg out.

tools/bwa.py
```diff
@@ -46,18 +48,28 @@ def index(self, inFasta, algorithm=None):
...
         os.unlink(fq1)
         os.unlink(fq2)
-        samtools.sort(aln_sam, outAlign)
+        samtools.sort(aln_sam, aln_sam_sorted)
         os.unlink(aln_sam)
-        samtools.index(outAlign)
+        # cannot index sam files; only do so if a bam is desired
+        if outAlign.endswith(".bam") or outAlign.endswith(".cram"):
+            # convert sam -> bam
+            samtools.view(["-b"], aln_sam_sorted, outAlign)
```
dpark01: Can we also get rid of this sam->bam conversion here (with the aln_sam_sorted temp file created above)? Takes a lot of excess tempdir disk space and runtime compared to before (where samtools.sort did the sam->bam conversion during sorting).  My suggestion would be to remove the aln_sam_sorted temp file, revert the samtools.sort to write directly to outAlign, and put only the samtools.index command in the if block (with nothing else, and no elif/else block)

tools/bwa.py
```diff
@@ -7,6 +7,8 @@
 import os
 import os.path
 import subprocess
+import shutil
```
dpark01: can probably remove import shutil if you make the other change below

tools/bwa.py
```diff
@@ -46,18 +48,28 @@ def index(self, inFasta, algorithm=None):
...
+
         threads = threads or util.misc.available_cpu_count()
         samtools = tools.samtools.SamtoolsTool()
         fq1 = util.file.mkstempfname('.1.fastq')
         fq2 = util.file.mkstempfname('.2.fastq')
         aln_sam = util.file.mkstempfname('.sam')
+        aln_sam_sorted = util.file.mkstempfname('sorted.sam')
         samtools.bam2fq(inReads, fq1, fq2)
-        self.execute('mem', ['-t', str(threads), refDb, fq1, fq2], stdout=aln_sam)
+        self.execute('mem', opts + ['-t', str(threads), refDb, fq1, fq2], stdout=aln_sam)
```
dpark01: may need to pull the threads param out of the self.execute array here and do something like:
```
if '-t' not in opts:
    threads = threads or utils.misc.available_cpu_count()
    opts.extend(('-t', str(threads))
```
then just `self.execute('mem', opts + [refDb, fq1, fq2], stdout=aln_sam)`

dpark01:
BTW this is pretty awesome and I think it'll be really useful.

One quick question: in the case of svg/pdf/vector outputs, what happens to the width/height defaults of 1024x768? Does it end up being 1024 inches x 768 inches?  I'm less familiar with matplotlib, but in R/ggplot, the more portable way to specify defaults is to do it in inches and set the default dpi as well to be something reasonable, so that it applies either way for vector or raster output (otherwise, the text ends up all messed up as it tries to put a 12pt font on an 8 foot canvas). If it isn't easy to do it in a generic way with matplotlib, don't worry about it, we can leave it to the user to specify dimensions every time they want a vector output.

haydenm:
This looks really great to me! A few nitpicky comments:

- It might be nice to reset genome position on the x-axis for each segment. When I do a Google image search for `genome manhattan plot` I only see chromosome labels on the x-axis (precise genome position is probably not so important there), but you might be able to find an example of this to base it on if you keep digging. If it seems like it'd be a lot of work in `matplotlib`, I'd forget about it; it's not super important.
- This is going to work great for small viruses. For viruses with a larger genome, it's possible `samtools depth -aa` could become problematic. You'll get a large `samtools depth` output and a large PDF that won't open because shows every point. Binning depth is a way around this; see [https://github.com/broadinstitute/hayden-work/blob/master/personal/bin/bin_depth](https://github.com/broadinstitute/hayden-work/blob/master/personal/bin/bin_depth) for an easy way to do this with `samtools`. But I would definitely ignore this for now because I don't think we work with anything large enough for this to be a problem. Let's just wait to see if this indeed becomes an issue, and keep it in mind. (I just looked back at what you wrote before I submit this comment, and I think you had the same idea with the sliding window average option.)

tools/bwa.py
tomkinsc: Sure, I'll change it.

tomkinsc:
@dpark01 RE: output sizes, matplotlib defaults to working in inches and DPI (or at least the interface does). I coerce it to pixels because it made sense for the raster output. For vector output, the dimensions currently get divided by the default DPI exposed by matplotlib. I'll make it an option to pass in a DPI for vector mode, with the matplotlib default DPI as the default for the parameter. This way the user can set pixel dimensions and DPI. The tiny text problem is currently addressed by manually scaling the labels based on the height of the plot and the DPI; they don't really seem to scale with the plot size otherwise.

dpark01:
Hey @tomkinsc this looks great. My inclination would be to merge it in and then if there are minor improvements suggested by anyone along the way, handle that separately in the future.

tomkinsc:
@dpark01: I'm in the process of exploring different matplotlib arrangements based on @haydenm's feedback, to address his first point of resetting the coordinate for each chromosome/segment. Unfortunately it means some careful application of sub-plots, and it's not straightforward to get something that looks good, especially considering how inconsistent the matplotlib API is for changing label sizes on subplots vs. figures. If I'm not pleased with the results, I'll roll back to 0f47ccc and merge this in. 

Regarding his second point, that's something I encountered while testing samtools' `depth` with an alignment to hg19—the resulting depth TSV file was ~50GB. Binning or averaging over a sliding window is probably the answer, but this is viral-ngs and not human-ngs, so I'm inclined to ignore the issue until we sequence something where data size/density is an issue.

dpark01:
I see. Personally, I'm not a big fan of resetting coords on each segment--I haven't seen too many manhattan plots that actually do this, though I see how it could potentially be useful for small genomes. It'd probably have more value to overlay gene annotations below the axis, but that's a much bigger feature add on.

The size is definitely going to be an issue for larger genomes, and binning is definitely the answer. If you're close, sure why not, but if you haven't started down that road yet, it can be added later.

tomkinsc:
Merging this in, but leaving the branch open for future improvements.

Issue #407, opened 2016-07-19
perform pre-Trinity subsampling in bam space


tomkinsc:
We should consider downsampling reads in bam space using Picard, samtools, or sambamba, rather than using our custom file seek-based subsample routine on fastq files. It would likely be faster, and remove complication associated with unpaired reads.

`samtools view -s $seed.$fractionOfReads -b $testBam > $subsampledTestBam`

`sambamba view -h -t $numThreads -s $fractionOfReads -f bam --subsampling-seed=$seed $testBam -o $subsampledTestBam`

https://broadinstitute.github.io/picard/command-line-overview.html#DownsampleSam

tomkinsc:
In #417.

PR #408, opened 2016-07-19
Keep unpaired trimmomatic output

haydenm:
Trimmomatic can output single-end reads if mates are dropped. Previously, its single-end output was stored in tmp files and trashed. This PR keeps its unpaired output and allows those reads to be used as input for Trinity assembly.

This is particularly useful for low quality and quantity samples. In these samples, often the second mate of a pair has a drop in quality (at least on MiSeq sequencing) and is tossed by Trimmomatic; this PR would allow the first mate of the pair to be used in assembly.

### DIFF
--- a/assembly.py
+++ b/assembly.py
@@ -67,37 +67,51 @@ def trim_rmdup_subsamp_reads(inBam, clipDb, outBam, n_reads=100000):
     n_input = util.file.count_fastq_reads(infq[0])
 
     # Trimmomatic
-    trimfq = list(map(util.file.mkstempfname, ['.trim.1.fastq', '.trim.2.fastq']))
+    trimfq = list(map(util.file.mkstempfname,
+                      ['.trim.1.fastq', '.trim.2.fastq']))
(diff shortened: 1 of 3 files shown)


dpark01:
Cool. Haven't looked at the diff yet but I assume this lets unmated reads flow down the filtering steps which would mean that they still get tossed out downstream if the read count is above the sub sample threshold, right? (I think that's desired behavior: we only want to recover single end reads if we don't have that many anyway)

haydenm:
Right, more or less. Specifically, if the number of reads (paired + single) is greater than the subsample threshold, it will purge unmated reads (tossing all the single ones) and then subsample from the paired-end reads.

This isn't ideal because you can imagine a case where there are a lot of single reads and, in purging unmated reads, you're tossing more single reads than necessary to get below the threshold -- potentially leaving you with very few paired-end reads from which to assemble. I don't think this will come up much, if at all, in practice. Regardless, @tomkinsc and I talked about this and figured out a fix that he'll implement as part of #407.

PR #411, opened 2016-07-25
select UGER queue for lastal rule based on read count; raise memory

tomkinsc:
The `filter_to_taxon` Snakemake rule now selects the UGER queue to use based on the number of reads present in the {sample}.cleaned.bam file, because previously HiSeq runs were timing out on the ‘short’ queue. Now bam files with >1.5M reads get pushed to the ‘long’ queue. Additionally, the memory limit has been raised for `filter_to_taxon` from 3G to 8G.


dpark01:
Hm, I don't think this will work.. your new python function will execute during snakemake's DAG construction, but the required input file won't exist yet by that point. This might be fundamentally hard in the snakemake way of describing workflows, because it doesn't really have a way to describe branching logic for pipelines based on information that doesn't exist up front...

dpark01:
Just to be safe, maybe it's fine if we just say the filter_to_taxon rule should always be on UGER long queue...

Also, just to be sure, when you say HiSeq runs are timing out, is that because filter_to_taxon is running on merged bams? Or is it even timing out on the per-lane pre-merge bams?

yesimon:
I actually implemented this feature in a PR to snakemake to defer resolution of callables until inputs exist for a rule here https://bitbucket.org/snakemake/snakemake/pull-requests/88/defer-evaluation-of-resources-callables/diff, but it looks like JK is thinking of another way to handle it. Regardless, support for such a feature should be arriving soon.

PR #414, opened 2016-07-26
add bam-space subsampling methods to Picard and Samtools

tomkinsc:
This adds bam-space subsampling functions to the Samtools and Picard tool wrappers in preparation of shifting from fastq to bam-based sampling prior to *de novo* assembly.

The following new functions are available:
picard.py::DownsampleSamTool()::downsample_to_approx_count()

samtools.py::SamtoolsTool()::downsample_to_approx_count()

The picard-based routine yields a number of reads closer to the desired count.


tools/picard.py
```diff
@@ -116,20 +119,17 @@ def per_read_group(self, inBam, outDir, picardOptions=None, JVMmemory=None):
                 os.mkdir(outDir)
         else:
             picardOptions = picardOptions or []
-            opts = ['INPUT=' + inBam, 'OUTPUT_DIR=' + outDir, 'OUTPUT_PER_RG=true']
+            opts = ['INPUT=' + inBam, 'OUTPUT_DIR=' + outDir, 'OUTPUT_PER_RG=true', 'RG_TAG=ID']
```
tomkinsc: @dpark01: I can use your input here. Is this OK? Upgrading Picard from 1.126 to 1.141 changes the usage of `OUTPUT_PER_RG`, and seems to require `RG_TAG=ID` to conform to prior usage. The change seems to have appeared in [`1.127`](https://github.com/broadinstitute/picard/commit/4fef32ef397e33114a3711469a8c4e66507fafb9)

dpark01: Hm, sounds fine to me. Sounds like they wanted to specify whether to break it up by PU or ID (if only they could break it up by LB.. oh well). Tests broke before you added this option, tests fixed after you added it... sure, let's go with that!

tomkinsc:
This is step one of addressing #407.

PR #417, opened 2016-07-27
perform pre-Trinity subsampling in bam space

tomkinsc:
Perform pre-Trinity subsampling in bam space. If the number of paired reads (after purge, or not) is lower than the Trinity subsampling threshold, subsample the unpaired reads to try to reach the total. Otherwise, if the post-purge paired read count is greater than the Trinity input, downsample to reach it. A few small improvements to the samtools and Picard downsample functions. This addresses #407.

### DIFF
--- a/assembly.py
+++ b/assembly.py
@@ -58,6 +58,9 @@ def trim_rmdup_subsamp_reads(inBam, clipDb, outBam, n_reads=100000):
         This should probably move over to read_utils or taxon_filter.
     '''
 
+    downsamplesam = tools.picard.DownsampleSamTool()
+    samtools      = tools.samtools.SamtoolsTool()
+
     if n_reads < 1:
         raise Exception()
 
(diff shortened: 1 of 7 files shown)


assembly.py
```diff
@@ -93,120 +96,110 @@ def trim_rmdup_subsamp_reads(inBam, clipDb, outBam, n_reads=100000):
...
-    if n_subsamp == 0:
-        # FastqToSam cannot deal with empty input
-        # but Picard SamFormatConverter can deal with empty files
-        opts = ['INPUT=' + tmp_header, 'OUTPUT=' + outBam, 'VERBOSITY=ERROR']
-        tools.picard.PicardTools().execute('SamFormatConverter', opts, JVMmemory='50m')
+
+    # convert paired reads to bam
+    # stub out an empty file if the input fastqs are empty
+    tmp_bam_paired = util.file.mkstempfname('.paired.bam')
+    if all(util.file.count_fastq_reads(x) > 0 for x in  purgefq) > 0:
```
dpark01: I'm not sure I understand the if expression here, there appears to be an extraneous `> 0` at the end. Should it be this?
```
if all(util.file.count_fastq_reads(x) > 0 for x in purgefq)
```

Also, if you're just testing for emptiness, can you just use os.path.getsize or similar? That should be O(1) instead of O(n) on the input.

PR #426, opened 2016-07-29
corrections related to picard 1.141's FastqToSam

tomkinsc:
previously picard’s FastqToSam used PU if present to name files, and fell back to ID. As of 1.141, it uses the field specified by RG_TAG. Since we explicitly use RG_TAG=ID now, we can remove the PU override from the file collection section of read_utils.rmdup_mvicuna_bam. Also add a unit test for rmdup_mvicuna_bam.


dpark01:
Yeah... that looks like it should do it. The unit test will need to be on non-empty data though in order to catch anything. This should become a tagged release once it's merged because it's a significant breaking issue on pretty much any use case.

tomkinsc:
The function name was borrowed—the actual test data is non-empty, and derived from @haydenm's failing case. It's now renamed!

PR #430, opened 2016-08-02
Update diamond to 0.8.22

yesimon:
Diamond requires libgcc<5 so it's going into its own special env.

### DIFF
--- a/tools/diamond.py
+++ b/tools/diamond.py
@@ -12,8 +12,8 @@ import subprocess
 import tools
 import util.file
 
-TOOL_VERSION = '0.7.10'
-CONDA_VERSION = tools.CondaPackageVersion('0.7.10', 'boost1.60_1')
+TOOL_VERSION = '0.8.22'
+CONDA_VERSION = tools.CondaPackageVersion('0.8.22', '2')
 
 log = logging.getLogger(__name__)
 
--- a/requirements-conda.txt
+++ b/requirements-conda.txt
@@ -1,7 +1,7 @@
 blast=2.2.31
 bmtagger=3.101
 bwa=0.7.15
-diamond=0.7.10=boost1.60_1
+diamond=0.8.22=2
 kraken-all=0.10.6_eaf8fb68
 krona=2.7
 last=719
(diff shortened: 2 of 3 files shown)


tomkinsc:
Nothing we're using requires libgcc>5, and we're currently using libgcc=4.8.5. Why is it necessary to put diamond in its own environment? 

I'd prefer to keep everything in the same environment for the sake of the viral-ngs bioconda recipe.

yesimon:
Samtools depends on ncurses which depends on libgcc, although it doesn't seem to specify version, the current one I had installed was `libgcc 5.2`. I think it may have to do with the ordering of installing something that requires `libgcc`, hence simply installing the newest version. Then later conda install with `--no-update-dependencies` fails due to version conflic.

```
yesimon ~/s/viral-ngs> conda install --no-update-dependencies  -y diamond=0.8.9 -p tools/conda-tools/default 
Using Anaconda Cloud api site https://api.anaconda.org
Fetching package metadata: ......
Solving package specifications: ....

The following specifications were found to be in conflict:
  -  libgcc 5.2.0 0
  - diamond 0.8.9* -> libgcc >=4.8.5,<5
```

tomkinsc:
If you update the diamond version in the [viral-ngs bioconda recipe](https://github.com/bioconda/bioconda-recipes/blob/master/recipes/viral-ngs/meta.yaml), do the dependencies resolve correctly? Libgcc is not pinned in the recipe, so it should try to find a version that will work with the various packages. If so, we should be able to keep diamond in the same environment. It may be an issue for Travis if samtools is installed first, since we aren't currently relying on conda to set up the environment on Travis. It would be nice to move to a conda environment file to define the dependencies to install on Travis, but there [are some issues there](https://github.com/conda/conda/issues/2800)...

tools/diamond.py
```diff
@@ -26,7 +26,7 @@ class Diamond(tools.Tool):
     def __init__(self, install_methods=None):
         if not install_methods:
             install_methods = [
-                tools.CondaPackage("diamond", version=CONDA_VERSION)
+                tools.CondaPackage("diamond", env="diamond", version=CONDA_VERSION)
```
tomkinsc: Let's omit this, and change the diamond version in `requirements-conda.txt`, which will handle global dependency resolution on Travis.

yesimon:
I'm holding off on this because diamond 0.8 has some segfault issues.

yesimon:
The fixed diamond build with `-march=x86-64` is now built as `0.8.22-2`:  bioconda/bioconda-recipes#2353.

There's no need for a separate env and OS X builds are now available.

PR #443, opened 2016-08-10
rework assembly.trim_rmdup_subsamp_reads()

tomkinsc:
reworked assembly.trim_rmdup_subsamp_reads() so unpaired reads from trimmomatic and singletons from prinseq are both de-duplicated together and used as the subsampling source for cases where the total paired read count is less than the desired number.

### DIFF
--- a/assembly.py
+++ b/assembly.py
@@ -152,7 +152,7 @@ def trim_rmdup_subsamp_reads(inBam, clipDb, outBam, n_reads=100000):
                                                   tmp_bam_unpaired)
 
             tmp_bam_unpaired_subsamp = util.file.mkstempfname('.unpaired.subsamp.bam')
-            reads_to_add = (n_reads - (n_rmdup_paired))
+            reads_to_add = (n_reads - (n_rmdup_paired*2))
             downsamplesam.downsample_to_approx_count(tmp_bam_unpaired, tmp_bam_unpaired_subsamp, reads_to_add)
             n_unpaired_subsamp = samtools.count(tmp_bam_unpaired_subsamp)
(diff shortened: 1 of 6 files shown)


dpark01:
Looks good. One question though. Are all the counts and numbers we're using "pairs" for read pairs and "singles" for unpaired reads? If the subsample threshold is 200, it sounds like we'd allow for either 200 paired read pairs or, for example, 180 read pairs plus 20 single unpaired reads?

If that is the case, maybe we should multiply all the paired counts by two and just increase the default subsample cutoff by a factor of two so that we're counting equal things.

tomkinsc:
> Are all the counts and numbers we're using "pairs" for read pairs and "singles" for unpaired reads? If the subsample threshold is 200, it sounds like we'd allow for either 200 paired read pairs or, for example, 180 read pairs plus 20 single unpaired reads?

That should be how it is behaving now. When we're in fastq space, we're only counting one of the two paired-read files, and in bam space the count subsample threshold is multiplied by two (though the Picard docs act on read pairs for accept/reject, it appears to consider each read of a pair separately).

You're saying we should count the paired reads twice to count each mate individually? I was just working off how the code was structured in the older fastq-based subsampling block. The log message is fairly explicit:

```
Pre-Trinity read filters:
  71 read pairs at start
  27 read pairs after Trimmomatic (44 unpaired)
  27 read pairs after Prinseq rmdup (0 unpaired)
  71 reads for trinity (27 read pairs + unpaired subsampled 44 -> 44)
  98 individual reads
```

tomkinsc:
Let's talk more in person about how we want to interpret the subsample threshold.

Issue #451, opened 2016-08-16
Add bwa as alternative aligner for align_and_fix


tomkinsc:
Add bwa as alternative aligner for align_and_fix

PR #456, opened 2016-08-17
add aligner option (novoalign or bwa) to read_utils.align_and_fix()

tomkinsc:
This addresses #451.


read_utils.py
```diff
@@ -1067,12 +1068,42 @@ def align_and_fix(
...
+            inBam, refFasta, bam_aligned,
+            options=aligner_options.split(),
+            JVMmemory=JVMmemory
+        )
+    elif aligner=='bwa':
+        bwa = tools.bwa.Bwa()
+        bwa.index(refFasta)
+
+        aln_bam_prefilter = util.file.mkstempfname('.prefiltered.bam')
+        bwa.mem(inBam, refFasta, aln_bam_prefilter, opts=aligner_options.split()+['-R', '@RG\tID:bwa\tSM:sample'])
```
dpark01: Oh nuts.... this is why vphaser won't work... because our BWA aligner doesn't preserve read groups.

dpark01: Okay, how about this. Take a cue from the Novoalign Tool wrapper and modify the `mem` function in the BWA tool to split out by read groups, align each read group one at a time with BWA-MEM, reheader each output BAM file, then merge all output BAMs together.  Sigh... that should solve it though, and you can use the Novoalign wrapper as an example.

tomkinsc: Perfect; thanks for the comment. Forcing a read group stood out as incorrect, and I had hoped you'd weigh in.

tools/bwa.py
```diff
@@ -47,7 +48,155 @@ def index(self, inFasta, algorithm=None):
...
+            for line in inf:
+                if line.startswith("@RG"):
+                    readgroup_line = line
+
+        assert len(readgroup_line) > 0
+        
+        aln_bam_prefilter = util.file.mkstempfname('.prefiltered.bam')
+        # rather than reheader the alignment bam file later so it has the readgroup information
+        # from the original bam file, we'll pass the RG line to bwa to write out
+        self.mem(one_rg_inBam, refDb, aln_bam_prefilter, opts=options+['-R', readgroup_line.rstrip("\n").rstrip("\r")], min_qual=min_qual, threads=threads)
```
tomkinsc: @dpark01: Is it fair game to pass the individual `@RG` line to BWA so we don't have to reheader downstream?

dpark01: I think there may or may not be other headers as well.. And samtools reheader is really fast actually (supposedly). Does our demux process emit any other headers other than RG?

tomkinsc: Samtools' `reheader` is particularly fast when it segfaults 😕, which I've seen on BWA output with our one-RG header. I'm trying to figure out why, but others online have reported similar failures.

A `.bam` file demuxed by our process and selected at random from DNAnexus included the standard `@HD` header line and then `@RG`:
```
@HD     VN:1.4  SO:queryname
@RG     ID:ARJYU.1      DT:2016-08-05T00:00:00+0000     PU:ARJYU.1.AGGCAGAA-GCGTAAGA    LB:BWH001       SM:BWH001       CN:M04004       PL:illumina
```
According to the [spec](https://samtools.github.io/hts-specs/SAMv1.pdf), `@HD` will always be present, `@SQ` will be present for an aligned sam/bam, we can assume to have `@RG`, the `@PG` program info line may be present (do we care about it?), and a `@CO` comment may be present. Along with passing in an `@RG` line to BWA via `-R` we can pass in the other lines via `-H`, which accept an `@line`, or a set of lines from a file (we'd have to remove `@HD` for the file option).

dpark01: I'm assuming our BWA code is emitting `@HD` properly. `@SQ` should not exist upstream (and will be added by BWA) so that shoudl be fine. `@PG` gets added by every thing that touches it along the way (the ailgner, GATK, etc). You can decide whether you want to add a `@PG` in a future iteration of your human depletion process.

So in sum, `@PG` might be the only thing to watch out for in your input, but otherwise, yeah, might as well do what you're doing?

PR #468, opened 2016-09-01
update encrypted gatk tarball to v3.6, remove novoalign from tarball since it is now installed by conda

tomkinsc:


dpark01:
Huh, what's in that `.tar.gz` file? Isn't it just a single `.jar` file?

tomkinsc:
It currently has the contents of the GATK download as provided by DSDE,  but it could definitely contain just the jar file (or encrypt the jar directly, since it's already a zip file).

PR #477, opened 2016-09-09
Kraken results on cleaned reads.

yesimon:
Also remove diamond from default 'all_metagenomics'.

### DIFF
--- a/pipes/Snakefile
+++ b/pipes/Snakefile
@@ -51,7 +51,7 @@ if config.get('spikeins_db'):
 if config.get('kraken_db'):
     all_inputs.extend(
         expand(os.path.join(config["data_dir"], config["subdirs"]["metagenomics"],
-                    "{sample}.kraken.report"), sample=samples_metagenomics))
+                    "{sample}.raw.kraken.report"), sample=samples_metagenomics))
 
 rule all:
     input: all_inputs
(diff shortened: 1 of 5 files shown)


tomkinsc:
Is this mainly to reduce runtime? In the DNAnexus implementation we actually moved from running kraken on cleaned reads to running it on raw reads, for a couple reasons: 1) it allowed us to kick off kraken in parallel with the depletion step, allowing the user to see the kraken report earlier (this should apply to the Snakemake implementation as well). Even with the larger input data size, it was still faster than waiting on kraken after running depletion and 2) to some degree, it can be insightful to see the human fraction in the krona report.

tomkinsc:
Maybe we should have separate sets of rules? `all_metagenomics` and `all_metagenomics_host_depleted`

yesimon:
No this is mainly a request for @haydenm just so we can compare the cleaning process. I guess I should add another rule so it won't block when asking for `all_metagenomics`.

haydenm:
This change should run kraken on _both_ raw and cleaned reads. I completely agree that seeing results on raw reads is most often what we want, but there are cases where we might want to see results on cleaned reads too. (For example, if Zika hybrid select probes contaminate a sample where we're sequencing Zika, we would want to see metagenomics after depleting those probes. There's no way to get those numbers by doing some arithmetic directly from the metagenomic output on raw reads.)

dpark01:
I like the idea of having both data sets analyzed in separate targets and output folders. Why did you remove diamond?

yesimon:
It's removed from the default for now while I resolve some instruction set issues with the binary on the heterogenous cluster nodes

PR #478, opened 2016-09-13
Fix usage of samples-runs.txt in reports.rules

haydenm:
Previously snakemake rules in reports.rules assumed that
samples-runs.txt would include only sample names. This
fix allows the correct usage of samples-runs.txt -- i.e.,
in which the lines include sample names and run info.


dpark01:
Huh, this seems to be a bugfix... thanks! Not sure how that happened to the spike-ins rule

haydenm:
A bugfix for my own commit (c615f02a), which broke it. (I had been using `samples-runs.txt` incorrectly.)

PR #479, opened 2016-09-13
Blacklist problematic UGER nodes

haydenm:
To bypass transient issues on a set of UGER nodes, maintain a
list of nodes that should be blacklisted. This list is maintained
by jobscript.sh and used in calls to qsub by cluster-submitter.py.

### DIFF
--- a/pipes/Broad_UGER/cluster-submitter.py
+++ b/pipes/Broad_UGER/cluster-submitter.py
@@ -2,6 +2,7 @@
 import os
 import sys
 import re
+import getpass
 from snakemake.utils import read_job_properties
 
 LOGDIR = sys.argv[-2]
@@ -10,6 +11,15 @@ mo = re.match(r'(\S+)/snakejob\.\S+\.(\d+)\.sh', jobscript)
 assert mo
 sm_tmpdir, sm_jobid = mo.groups()
 props = read_job_properties(jobscript)
+
+# Blacklist problematic nodes; this list is stored as filenames
+# in /broad/hptmp/[username]/blacklisted-nodes/
+whoami = getpass.getuser()
+blacklisted_node_dir = os.path.join("/broad/hptmp", whoami, "blacklisted-nodes")
+if os.path.isdir(blacklisted_node_dir):
+    blacklisted_nodes = os.listdir(blacklisted_node_dir)
+else:
+    blacklisted_nodes = []
 
 # set up job name, project name
 jobname = "{rule}-{jobid}".format(rule=props["rule"], jobid=sm_jobid)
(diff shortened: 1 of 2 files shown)


pipes/Broad_UGER/jobscript.sh
```diff
@@ -15,11 +15,30 @@ export PATH="$MINICONDADIR/bin:$PATH"
...
-# since the node with the job doesn't have the NFS share mounted.
-# Hopefully it will be sent to a node that has it mounted.
-# For this to work, re-run must be set to y via "qsub -r y"
-ls "$DATADIR" || exit 99
+
+# As a simple solution to transient UGER problems, maintain a list of blacklisted
+# nodes. When a node fails a check, add its hostname to the list and exit 99
+# to reschedule the job. Blacklist these nodes via qsub
+# Maintain the list of blacklisted nodes as filenames in /broad/hptmp/$(whoami)/blacklisted-nodes
+BLACKLISTED_NODES="/broad/hptmp/$(whoami)/blacklisted-nodes/"
```
yesimon: I would recommend leaving final slash out as convention

pipes/Broad_UGER/cluster-submitter.py
```diff
@@ -27,9 +37,17 @@
 if mem:
     # on UGER, memory requests are per-core (according to BITS as of Sept. 6, 2016)
     mem_per_core = round((int(mem)*1024)/int(cores), 2)
-    cmdline += ' -l h_vmem={}M,h_rss={}M '.format(  mem_per_core, round(1.2 * mem_per_core, 2)  )
+    if blacklisted_nodes:
+        cmdline += " -l h_vmem={}M,h_rss={}M,h='!({})' ".format(
```
tomkinsc: We should probably have comment lines here explaining the various resource parameters, or at least for `h=` mention that it's the hostname param, that it [accepts a regex](http://wiki.gridengine.info/wiki/index.php/EvaluationExpressionSupport), and we take advantage of that invert the match and blacklist hostnames.

tomkinsc:
Let's hold off on merging this until we see if we can modify snakemake to avoid semaphore-based locks. It would still be helpful to have the blacklist ability, but we may be able to get away without checking the available shared memory.

dpark01:
If you end up solving it that way, then perhaps it's good anyway to add a manually-curated blacklist (that is read-only as far as our scripts are concerned). Or maybe you just maintain it based on swappyness.

haydenm:
I think we should maintain this list dynamically in its current framework. It's an easy way to fix various other node-specific issues that could (and likely will) arise in the future. But it's also a list we can manually add to and delete from. Perhaps we can add a flag to the top of `jobscript.sh` that specifies whether we want to check shared memory (and the other checks)? Of course, if the modification of snakemake solves it, we would set that flag to false.

tomkinsc:
A [PR is open](https://bitbucket.org/snakemake/snakemake/pull-requests/109/add-force-use-threads-option-to-force/diff) to add a `--force-use-threads` option to Snakemake. If/when it is accepted, we can add the flag to our Snakemake call [in `run-pipe.sh`](https://github.com/broadinstitute/viral-ngs/blob/master/pipes/Broad_UGER/run-pipe.sh#L25-L32) once we update to a new release of Snakemake that includes the feature.

tomkinsc:
Maybe change the [conda config](https://github.com/broadinstitute/viral-ngs/blob/master/travis/install-conda.sh#L43) to also depend on the `conda-forge` channel:
```shell
conda config --add channels bioconda
conda config --add channels conda-forge # <===
conda config --add channels r
```
The [docs](https://github.com/broadinstitute/viral-ngs/blob/master/docs/install.rst#configure-conda), and the easy-deploy script should also be updated if we add another channel dependency. In the script, multiple channels can be specified with additional `-c` params: `-c <channel> -c <channel2>`

PR #484, opened 2016-09-15
YAPF all Python files.

yesimon:


yesimon:
That's strange because I was just running it based on the default repo
style. In general I don't really care as long as it's consistent, and the
whole repo has been fairly consistent on style.

tomkinsc:
I generally just conform to whatever style is in use. That said, it is nice to have a consistent style, but autoformatting is one of those things that's best to start using yesterday. The git blame issue is real. Without having autoformatting from the start and requiring it on a pre-commit hook, I'm inclined to close this PR for now, at least until we can all get on the same page about style and commit hooks.

PR #486, opened 2016-09-15
Metagenomics BWA alignment + MarkDuplicates + LCA

yesimon:
An alternative metagenomics analysis aligning to a bwa index with encoded taxids in the reference name, followed by Picard's MarkDuplicates and subsequent LCA.

Note that it's not possible to create `pysam.AlignedSegment` objects with custom reference names within python for testing (has to be read from file), so we're leaving that to the integration tests.

### DIFF
--- a/metagenomics.py
+++ b/metagenomics.py
@@ -18,27 +18,34 @@ import operator
 import queue
 import shutil
 import sys
+
+import pysam
+
 import util.cmd
 import util.file
 import util.misc
+import tools.bwa
+import tools.diamond
 import tools.kraken
 import tools.krona
-import tools.diamond
 import tools.picard
 
 __commands__ = []
 
-
 log = logging.getLogger(__name__)
+
+
+class TaxIdError(ValueError):
+    '''Taxonomy ID couldn't be determined.'''
 
 
 class TaxonomyDb(object):
 
-    def __init__(self, tax_dir=None, gis=None, nodes=None, names=None,
-                 gis_paths=None, nodes_path=None, names_path=None):
+    def __init__(
+        self, tax_dir=None, gis=None, nodes=None, names=None, gis_paths=None, nodes_path=None, names_path=None
+    ):
         if tax_dir:
-            gis_paths = [join(tax_dir, 'gi_taxid_nucl.dmp'),
-                         join(tax_dir, 'gi_taxid_prot.dmp')]
+            gis_paths = [join(tax_dir, 'gi_taxid_nucl.dmp'), join(tax_dir, 'gi_taxid_prot.dmp')]
             nodes_path = join(tax_dir, 'nodes.dmp')
             names_path = join(tax_dir, 'names.dmp')
         self.gis_paths = gis_paths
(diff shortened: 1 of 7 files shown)


dpark01:
@yesimon just so I understand, did you pull out the more infrastructural (conda/travis/pip) bits and submit as PR #489 ? If so, is the idea that we'll tackle #489 first and then resubmit #486 isolated to just the metagenomics.py new code?

yesimon:
Yes - didn't want to have to support old pysam then migrate to new pysam again.

dpark01:
Okay so then my thought is let's close this PR until #489 gets merged in. Then we can resubmit so that it's just the metagenomics pieces.

From a quick skim through, if you remove the infrastructural pieces and all the whitespace changes, the actual new content as advertised (adding BWA-LCA metagenomics) should be only inserting new lines into metagenomics.py alone, with no other changes. The signal-to-noise ratio on these diffs is really low with all that other stuff in the way and makes it hard to review.

One more comment is that, although we can add testing later on, we should include some documentation and/or scripts about how to generate the desired databases (e.g. list of accessions -> pull the fastas -> modify the headers to include taxids -> index with bwa, etc).

Also, is this still using GI numbers? We have 2 weeks to switch to accession.version, right?

yesimon:
@dpark The new stuff has no dependency on gi numbers because it's relying on the bwa index to have taxids encoded in every reference sequence name.

dpark01:
Okay @yesimon now that the other PRs have merged, I'll leave it to you to rebase these changes on current master, and try to leave out whitespace/stylistic changes on the completely unrelated code so as to increase the SNR of the github diff. This should mostly stay in metagenomics.py and related tests & pipes.

pipes/config.yaml
```diff
@@ -55,7 +55,7 @@ diamond_db: "/idi/sabeti-scratch/yesimon/db/diamond_db/nr"
 krona_db: "/idi/sabeti-scratch/yesimon/db/krona_taxonomy"
 
 # BWA index prefix for metagenomic alignment.
-align_rna_db: "/idi/sabeti-scratch/yesimon/db/rna_bwa/rna_bwa"
+align_rna_db: "/idi/sabeti-scratch/yesimon/db/rna_bwa/human_viral_16s"
```
tomkinsc: Should this, the krona db, and the taxonomy db point to databases stored in the newly-created `/idi/sabeti-scratch/shared-resources/`? (I see you already have an `rna_bwa` directory there.)

yesimon: Yes I'll switch it to that one.

metagenomics.py
```diff
@@ -563,15 +623,109 @@ def parser_diamond(parser=argparse.ArgumentParser()):
...
+    outLca=None,
+    sensitive=False,
+    excludeDuplicates=True,
+    JVMmemory=None,
+    numThreads=None,
+    picardOptions=None,
+    min_score_to_output=None
+):
+    '''
+        Take reads, align to reference with BWA-MEM, and generate a coverage plot
```
dpark01: This comment is probably incorret?

PR #489, opened 2016-09-16
Use conda python for python packages on travis.

yesimon:
Python packages installed to the 'default' environment.

### DIFF
--- a/.coveragerc
+++ b/.coveragerc
@@ -2,6 +2,7 @@
 # source is not used here because it is specified in pytest.ini
 [run]
 branch = True
+omit = tools/conda-tools/*
 
 [report]
 
--- a/.travis.yml
+++ b/.travis.yml
@@ -45,8 +45,8 @@ before_install:
 
 install:
   - source travis/install-conda.sh
-  - travis/install-pip.sh
   - source travis/install-tools.sh
+  - travis/install-pip.sh
 
 script:
   - travis/tests-unit.sh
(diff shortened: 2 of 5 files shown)


yesimon:
I'm not sure why coveralls is upset

yesimon:
I'm setting the appropriate Python version in 'conda create'

yesimon:
Another note here is that Travis's python versions are basically bypassed and we're installing our own versions of python. I think this is relatively reasonable since a move to conda would suggest that installing conda's python works well and is necessary.

dpark01:
@yesimon yeah been pondering that move anyway, I just didn't realize conda came with multiple versions of py3

yesimon:
@dpark01 So I double checked and it looks like the 3.4 tests on travis are picking up python3.5 from the miniconda cache. We'll need to have a cache per python version?

dpark01:
Wait--the only persistent cache is handled by Travis, and that's definitely separate for each point on the build matrix (which in our case, is just the three Python versions). Not sure how that got mixed up but I assume you've tried clearing the Travis cache and rebuilding?

yesimon:
Sorry had to correct a typo 😦

yesimon:
@tomkinsc don't forget to update the bioconda recipes, the biopython and matplotlib versions were also updated

tomkinsc:
Thanks. Already in the staged commit! I'm the one who added biopython 1.68 to bioconda. :)

PR #491, opened 2016-09-16
Set shell executable to /bin/bash on snakemake rules

yesimon:


tomkinsc:
This seems fine, but were there issues previously?

dpark01:
If you set this in common.rules and reports_only, why is it needed elsewhere?

yesimon:
@tomkinsc If you don't specify it just uses the user's shell, but I think we want to standardize on bash.

@dpark01 let me try that

yesimon:
Why did the first build fail? 😕

tomkinsc:
It was a transient Travis error (couldn't access GitHub).

PR #493, opened 2016-09-16
Update Picard to 2.5.0

yesimon:


yesimon:
So we have 2 separate things going on here:

1. Set `$JAVA_HOME` correctly for travis tests (otherwise it picks up system java)
2. Set `$JAVA_HOME=` blank for picard invocations because the conda supplied wrapper script around it finds the conda-packaged version of java and executes. With `$JAVA_HOME` overriding if present.

The correct thing is to probably do both.

travis/install-tools.sh
```diff
@@ -31,6 +31,9 @@ else
     cp requirements-conda.txt $HOME
 fi
 
+# Set to conda's java
+# export JAVA_HOME="$(pwd)/tools/conda-tools/default/jre"
```
dpark01: So is this `export JAVA_HOME` statement commented out because it already happens anyway when we activate the conda environment?

dpark01: This is a nice clean diff and looks pretty good to me...

yesimon: Forgot to implement  point 1 from above.

yesimon:
Yes. Conda's picard is actually a shell wrapper script that finds conda's java and calls it with its own `picard.jar`. However it gets overriden by `$JAVA_HOME` if present. I tested the above and if neither fix from above is there then it would fail.

dpark01:
@tomkinsc should our easy deploy script set JAVA_HOME after activating the conda env?

travis/install-tools.sh
```diff
@@ -31,6 +31,9 @@ else
     cp requirements-conda.txt $HOME
 fi
 
+# Set to conda's java
+export JAVA_HOME="$(pwd)/tools/conda-tools/default/jre"
```
tomkinsc: Considering the Picard wrapper locates the JRE, do we need to override `JAVA_HOME`? Could we simply `unset JAVA_HOME`?

yesimon:
The Picard wrapper can figure it out but not always for other tools I
believe. I don't think there's any harm in setting it on just Travis to
make sure Java things work.

PR #495, opened 2016-09-19
Move empty kraken test to integration

yesimon:
Reuse the same process to build a db from raw data.

This doesn't remove the tarred db which should be purged at a later time


test/integration/test_kraken.py
```diff
@@ -113,8 +113,9 @@ def test_kraken_tool(tmpdir, kraken, kraken_db, input_bam):
     kraken.classify(input_bam, kraken_db, out)
     kraken.filter(out, kraken_db, out_filtered, 0.05)
     kraken.report(out_filtered, kraken_db, out_report)
+    with util.file.open_or_gzopen(out_filtered, 'r') as inf:
+        assert len(inf.read()) > 0
```
dpark01: Can you convert this call (`len(inf.read()) > 0`) to something that is O(1) memory and runtime (like `os.path.getsize`)? Especially memory. Also in `test_kraken`.

test/integration/test_kraken.py
```diff
@@ -173,3 +175,25 @@ def test_kraken_krona(tmpdir, kraken_db, krona_db, input_bam):
     parser = metagenomics.parser_krona(argparse.ArgumentParser())
     args = parser.parse_args([out_reads, krona_db, out_html])
     args.func_main(args)
+
+
+
+@pytest.mark.skipif(tools.is_osx(), reason="kraken osx binary does not yet exist on bioconda")
+def test_kraken_on_empty(kraken_db, input_bam):
```
dpark01: Why didn't the remaining tests move over as well? Those all failed at some point before we hardened the code for empty data handling.

test/integration/test_kraken.py
yesimon: The reads file is actually gziped so it has a small nonzero size.

test/integration/test_kraken.py
yesimon: The remaining tests mock out the actual kraken invocation so it's more for just ensuring that the right things are happening in `metagenomics.py`

yesimon:
As far as the time goes, it was taking ~2 min which is >10% of test time. Now when consolidated into the integration test it utilizes the shared kraken db fixture so there's no extra cost for database setup anymore.

dpark01:
How long does the fixture setup take? I don't see py.test on Travis reporting any fixture setups taking more than a few seconds -- it seems to be all the function calls themselves.

Are you saying this change means the kraken binary never gets executed anywhere in `test/unit`?

yesimon:
It takes a few seconds. It's unfortunate that py.test doesn't break down timings on fixtures and just groups them all under 'setup'. This is annoying especially when you have lots of dependent fixtures. 

Yes the kraken binary should not be executed in `test/unit`

test/unit/test_tools_kraken.py
```diff
@@ -85,64 +85,3 @@ def test_classify_num_threads(self):
...
-
-    def setUp(self):
-        super().setUp()
-        input_dir = os.path.join(util.file.get_test_input_path(), 'TestMetagenomicsViralMix')
-        self.kraken_db_viral_mix = tempfile.mkdtemp('-kraken_db_viral_mix')
-        cmd = ['tar', '-C', self.kraken_db_viral_mix, '-xzf', os.path.join(input_dir, 'kraken_db.tar.gz')]
-        subprocess.check_call(cmd)
-        self.kraken = tools.kraken.Kraken()
-
-    def test_kraken_execution(self):
```
dpark01: Okay I think I understand all the bits and pieces of this PR now. The main change I want to see is that [`test_kraken`](https://github.com/broadinstitute/viral-ngs/blob/master/test/integration/test_kraken.py#L121) gain some of the output checking that is now lost by removing `test_kraken_execution` here. And I'm up for eliminating `test_kraken_tool`, `input_bam`, and `fastq_to_sam` from there as well to help clean it up. And if you want to work into this PR (or a future one) the testing of kraken-build, that'd be huge...

yesimon:
Addressed in #653

PR #501, opened 2016-09-21
allow GATK and Novoalign license file to be passed in to functions requiring them

tomkinsc:
This supports WDL/CWL, which make it difficult or impossible to mount arbitrary host volumes in Docker containers. Note that for the Docker use case, the --GATK_PATH or --NOVOALIGN_LICENSE_PATH need to be within the Dockerfile’s CWD.

### DIFF
--- a/assembly.py
+++ b/assembly.py
@@ -495,7 +495,8 @@ def impute_from_reference(
     replaceLength,
     newName=None,
     aligner='muscle',
-    index=False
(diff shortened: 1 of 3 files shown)


tomkinsc:
The behavior from before should be unchanged if the new args are omitted. There are no changes to `tools/gatk.py` because the constructor signature already supported passing in a path.

PR #504, opened 2016-09-22
retry conda install

tomkinsc:
add retry option to run_and_print; this is to support retrying conda
installs when anaconda.org errors out

### DIFF
--- a/travis/install-tools.sh
+++ b/travis/install-tools.sh
@@ -37,11 +37,8 @@ export JAVA_HOME="$(pwd)/tools/conda-tools/default/jre"
 echo "Installing and validating bioinformatic tools"
 export CONDA_ENVS_PATH=tools/conda-cache:tools/conda-tools/default
 
-tries=0
-until [ $tries -ge 4 ]
-do
+for i in $(seq 3); do
   conda create -y -m -c bioconda -p tools/conda-tools/default --file $HOME/requirements-conda.txt python="$TRAVIS_PYTHON_VERSION" && break
-  tries=$[$tries+1]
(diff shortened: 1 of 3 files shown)


travis/install-tools.sh
```diff
@@ -36,7 +36,14 @@ export JAVA_HOME="$(pwd)/tools/conda-tools/default/jre"
 
 echo "Installing and validating bioinformatic tools"
 export CONDA_ENVS_PATH=tools/conda-cache:tools/conda-tools/default
-conda create -y -m -c bioconda -p tools/conda-tools/default --file $HOME/requirements-conda.txt python="$TRAVIS_PYTHON_VERSION"
+
+tries=0
+until [ $tries -ge 4 ]
```
yesimon: for i in $(seq 3)

util/misc.py
```diff
@@ -198,80 +199,91 @@ def run(args, stdin=None, stdout=None, stderr=None, shell=False,
 def run_and_print(args, stdout=None, stderr=None,
                   stdin=None, shell=False, env=None, cwd=None,
                   timeout=None, silent=False, buffered=False, check=False,
-                  loglevel=None):
+                  loglevel=None, expected_return_code=0, retries=0, retry_delay_sec=2):
```
yesimon: So we should have an `assert not (check and retries)` ?

yesimon: Also need docs that expected_return_code is only for retries?

dpark01:
(unless I'm misunderstanding how the problem is manifesting itself...)

PR #508, opened 2016-09-23
Unify column for krona reports 

yesimon:
* Print when loading each taxonomy db part.
* Use updated shared taxonomy db.
* Convert krona snakemake rule to `shell` over `run` using new params
  callable functionality.
* Add krona mem req


metagenomics.py
```diff
@@ -67,16 +67,22 @@ def __init__(
             elif gis_paths:
                 self.gis = {}
                 for gi_path in gis_paths:
+                    print('Loading taxonomy gis: {}'.format(gi_path))
+                    sys.stdout.flush()
```
dpark01: can this be log.debug or log.info instead of print?

pipes/rules/metagenomics.rules
```diff
@@ -85,12 +85,12 @@ rule krona_import_taxonomy:
     input: lambda wildcards: join(config["data_dir"], config["subdirs"]["metagenomics"], \
            ".".join([wildcards.sample, wildcards.adjective, method_props[wildcards.method]['reads_ext']]))
     output: join(config["data_dir"], config["subdirs"]["metagenomics"], "{sample}.{adjective,raw|cleaned}.{method,kraken|diamond|rna_bwa|rna_bwa_dupes}.krona.html")
-    run:
-        # method_props key can be changed to allow for different tools,
```
dpark01: Let's actually make this deleted comment correct (make the code reflect what it says) instead of deleting it.

yesimon:
For the deduped report unmapped and/or duplicate reads are excluded because otherwise the non-deduped unclassified reads are an inflated fraction of unique reads.

I'll set `sensitive` to true and see how that goes.

dpark01:
For the unmapped reads, I was primarily referring to the tsv (per-read) output.
For the non-deduped summary report, you are counting unmapped as well?

yesimon:
For non-deduped, unmapped go into unclassified.

haydenm:
I'd suggest not making `sensitive` the default, or even getting rid of it. Those bwa parameters are ones I've used when aligning against a single reference genome. When I tried them here via `sensitive`, bwa ran significantly slower (~days) than with its default parameters. I think aligning against many tens of thousands of genomes should help be more sensitive anyway, at least for reads from the diverse taxa for which there exists a lot of data.

PR #516, opened 2016-10-06
Add plugin to measure top N fixture executions

yesimon:


yesimon:
Doesn't seem to print in combination with `pytest-xdist` although testing without works fine

PR #521, opened 2016-10-14
Add --wait-submit option to run-pipe

yesimon:
This results in not submitting all jobs to the cluster scheduler with
dependencies right away, instead letting snakemake take control of
submitting jobs. Useful because the errored jobs in Eqw count against
your quota even though snakemake is unaware of this.

Will need to wait on snakemake update as well.

### DIFF
--- a/pipes/Broad_UGER/jobscript.sh
+++ b/pipes/Broad_UGER/jobscript.sh
@@ -65,6 +65,7 @@ echo $JOB_ID
 echo "=============================="
 
 {exec_job}
+EXIT_STATUS=$?
 
 # Report resource consumption because it's not reported by default
 echo "------------------------------"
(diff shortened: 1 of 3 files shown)


dpark01:
Sounds good, but why does it need a snakemake update?

yesimon:
I PRing snakemake to provide the `immediate_submit` variable in the jobscript template so that it knows whether or not to `exit 100`. Going with immediate_submit toggles on exit 100 but otherwise we don't want that because it takes up quota

dpark01:
Oh well if you want this now (depending on how speedy Johannes is with that PR), just make a jobscript2 or something and specify that instead in run-pipe if wait-submit.

dpark01:
What's the status of this? If upstream snakemake hasn't merged the PR, then just create two separate jobscripts on our end and call it a day..

yesimon:
Snakemake has been updated and now has the `immediate-submit` property.

tomkinsc:
This is confirmed to work on the Broad cluster, @yesimon?

PR #524, opened 2016-10-17
No prepend column in dfs reports

yesimon:


dpark01:
Can't think of a good reason for this

yesimon:
I think this was accidentally added to unify different report types but this is actually for the final tree type report that is the same across the board. Right now it just adds a spurious tab to the first line of the report.

dpark01:
Oh, the summary human readable indented file? Yeah

Issue #527, opened 2016-10-20
avoid RevertSam in deplete_human when possible


dpark01:
Right now, we call RevertSam at the beginning of `deplete_human`. That's mostly just defensive for a few scenarios (e.g. someone gives us a aligned bam, etc), but in most cases (e.g. this come straight after demux) is pointless and wastes a lot of time.

Perhaps we should only RevertSam if we notice certain features of the bam coming in (e.g. it looks to be aligned) which can be queried quickly just by looking at the bam header...

tomkinsc:
Looking at the [sam/bam spec](https://samtools.github.io/hts-specs/SAMv1.pdf), we should be able to detect aligned bam files by looking for `@SQ` lines in the header, right?

Are there other cases where we'd want to call `RevertSam`?

tomkinsc:
In #529.

PR #529, opened 2016-10-20
avoid RevertSam in deplete_human

tomkinsc:
RevertSam is expensive, since it reads, alters, and sorts all records in a bam file. This changes `main_deplete_human` so RevertSam is only executed if the input bam file is aligned. It determines if the file is aligned by checking whether the `@SQ` line exists and is populated in the bam header (via pysam). Since RevertSam also sanitizes, we lose some functionality: “_Reads discarded include (but are not limited to) paired reads with missing mates, duplicated records, records with mismatches in length of bases and qualities._”

revertBam has been made an optional argument via `nargs=‘?’`, so calls to`deplete_human` will not need to be changed. A test case has been added for aligned input.

This addresses #527.


dpark01:
Just so I understand: if the user specifies four bam files, the second one is the revert.bam, and if they specify three, the revertbam is a tempfile. Sounds good to me.

tomkinsc:
Right; I didn't want the API to change, and RevertSam only happens if the file is aligned (or rather, if it has an `@SQ` line).

PR #531, opened 2016-11-02
changes from Aaron Lin to better support multiple annotations with snpEff

tomkinsc:


dpark01:
Sorry I've been negligent on the Slack chatter about this, but I'd be happy to take a look. This is a pretty familiar issue for me, and last time I looked at it, it required manual corrections to the DB by snpEff's author.

tomkinsc:
Is it worthwhile for us to reach out to the snpEff author via: http://snpeff.sourceforge.net/about.html#bugs ?

dpark01:
No, it was a hack to get around how the gene model was coded in Genbank... if it had been coded more like alternative transcripts of a eukaryote, it would've been easier, but I think the original virologists wanted to stick closer to the molecular truth (there's no actual mRNA) instead of bioinformatic convention.

PR #534, opened 2016-11-28
Raise exception if inputs cannot be found for merge_one_per_sample

yesimon:
Previously would try to execute with empty input.

### DIFF
--- a/pipes/rules/hs_deplete.rules
+++ b/pipes/rules/hs_deplete.rules
@@ -63,6 +63,11 @@ rule filter_to_taxon:
             logid="{sample}"
     shell:  "{config[bin_dir]}/taxon_filter.py filter_lastal_bam {input[0]} {params.lastalDb} {output}"
 
+
+class MergeInputException(Exception):
+    '''Cannot find merging multiplexed sample inputs'''
+
+
 def merge_one_per_sample_inputs(wildcards):
     if 'seqruns_demux' not in config or not os.path.isfile(config['seqruns_demux']):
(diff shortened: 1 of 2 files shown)


dpark01:
In general for new features / changed behavior, it's probably worth discussing in an Issue or on Slack before putting in a PR (to gain a sense from others on the team of both *whether* and *how* before implementing). But this seems okay to me and the right approach: it keeps all the atomic python scripts tolerant of empty inputs, but it puts the use-case sanity checks in the pipeline itself. A few questions:

1. Has it been tested?
2. If the criteria are not met, does it throw an Exception during dry runs (`snakemake -n`) / DAG-creation, or only during execution? My guess (and hope) would be the former--we'd want to report this as early as possible.
3. Instead of our own custom exception shouldn't we throw snakemake's MissingInputException (or perhaps a subclass of it)?

Any thoughts and comments from @tomkinsc @haydenm ?

tomkinsc:
Similar feedback: 1) it would be nice to have a test for the empty input case, and 3) subclassing Snakemake's MissingInputException seems like the way to go.

In general, we need documentation of what `flowcells.txt` and `barcodes.txt` should look like.

yesimon:
I added a simple test to check failure with snakemake dry run on missing input.

Snakemake's `MissingInputException` is a class that expects certain information in the constructor, like `rule` and `files` that is not known from within the input function so it wouldn't be a good fit. We can't know the input files because they are dependent on what's in flowcells and barcodes.

PR #535, opened 2016-12-05
Update db and resources used for bwa metagenomics

yesimon:
Convert from just 16s to all rRNA (including SSU and LSU) from SILVA


yesimon:
Added the dbs and comments

PR #549, opened 2016-12-13
raise trinity rule memory 4g->10g, JVMmemory 5g

tomkinsc:


dpark01:
In principle, this seems alright, but I'm wondering two things:
1. In what scenario does Trinity require 5G?
2. In what scenario does the `assemble_trinity` command require more memory than the `JVMmemory`? (shouldn't they be the same, or close to same?)

tomkinsc:
We've seen Trinity fail with the prior limits memory limit, and raising the `h_vmem` solved the issue. For @haydenm, the scenario may be that amplicon-seq data from many pooled samples results in a larger than usual De Bruijn graph in the Chrysalis stage. Katie has also seen failures in the butterfly stage.

For Trinity, the JVMmemory argument is only heap ([`--bflyHeapSpace`](https://github.com/broadinstitute/viral-ngs/blob/master/tools/trinity.py#L49)), so presumably with per-thread stack, etc. the `--JVMmemory` value could differ from the total required, especially with eight threads on large graphs. Raising JVMmemory in this PR is mostly for insurance since we have a bit more headroom with the increase to `h_vmem`.

dpark01:
Okay sounds great, though let's not be too Broad_UGER-centric in our tweaks to the resource requests (asking for more just because it doesn't hurt). 7-7.5GB is a bit of a sweet spot on AWS/DNAnexus (as well as your average desktop/laptop), so if we can get away with that here, that would be nice. The stack and other non-heap memory shouldn't be > 2GB, even with a bunch of threads...

And as long as it's easily modifiable by the end user, the default command line params don't have to satisfy every scenario, just the vast majority of them (amp-seq is a little special).

tomkinsc:
@yesimon's suggestion of replacing hard-coded resource values with functions that can scale resources to input size is likely the right approach, but maybe only worth characterizing in the post-UGER world. Maybe we should log the data size and memory high-water mark for all current jobs so we can collect the necessary information as a secondary product of normal operation.

dpark01:
I agree. For now though, can we get away with 7-8GB?

PR #553, opened 2016-12-22
Update pytest version to conda-available verison

yesimon:
Necessary to support fixture executions plugin but 2.9.3 is not in conda channels


tomkinsc:
Do we need to update pytest upstream in bioconda?

yesimon:
Nope!

PR #566, opened 2017-01-24
Set --tmp_dir in metagenomics rules from config[tmp_dir]

yesimon:
Use config's `tmp_dir` as command line arg `--tmp_dir`.

### DIFF
--- a/pipes/rules/common.rules
+++ b/pipes/rules/common.rules
@@ -74,3 +74,5 @@ def webfile_readlines(uriToGet):
         cleanedLine = line.decode("utf-8").strip()
         if len(cleanedLine) > 0:
             yield cleanedLine
+
+os.makedirs(config.get('tmp_dir'), exist_ok=True)
--- a/pipes/rules/metagenomics.rules
+++ b/pipes/rules/metagenomics.rules
@@ -45,7 +45,7 @@ rule diamond:
             UGER=config.get('UGER_queues', {}).get('long', '-q long')
     shell:
         """
-        {config[bin_dir]}/metagenomics.py diamond {input} {config[diamond_db]} {config[taxonomy_db]} {output.report} --outLca {output.lca} --numThreads {params.numThreads}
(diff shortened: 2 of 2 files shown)


yesimon:
Yup running out of /local/scratch/ space. They are different things but two sides of the same coin? We should be controlling where tmpfiles are generated and cleaned up by default - maybe we can make the "uninteresting" tmp files a subdir.

dpark01:
The uninteresting intermediates (the config[tmp_dir]) are meant to stick around for various reasons.. sometimes for debugging failures, sometimes even for debugging successful runs. Things like trinity contigs are actually quite useful and don't take much space. There may be some stuff in there that we want to be more proactive about cleaning up though. But one thing is for sure: this definitely cannot be on instance/ephemeral storage, because snakemake relies on these files being NFS-reachable.

The --tmp_dir command line option (and TMP_DIR environment variable, which was easier to manually override in LSF) are intended to describe where to put truly ephemeral files (like fastqs, uncompressed things, etc) that are often large and there's no point burdening the network interfaces for them.  /local/scratch is almost always the correct answer on the Broad, but hptmp works as well and qscratch should be explored as an option as well. The problem under UGER (vs LSF) is that it's harder to override this for a specific run. So we should probably add a config variable that's some sort of ordered list of directories to try/prefer and the wrapper shell scripts can just "export TMP_DIR=xxx" and be done with it?

Also, intermediate fastq files (for command line tools that require fastq) really ought to be `gzip -1`'ed. Any reason not to?

yesimon:
Yeah, the problem is sometimes /local/scratch is filled or doesn't have enough space, but right now it's hard to override `$TMPDIR` env variable for snakemake usage. Since we probably shouldn't be hardcoding logic about selecting /local/scratch vs hptmp based on known quotas (we can't use regular disk usage checks to determine if there's going to be enough space), one complicated solution would be to have user-defined python functions or shell scripts to determine the correct path - this is very complicated.

Perhaps it's just easier to add a separate config variable for ephemeral `$TMPDIR` and/or check the user shell's '$TMPDIR` for an override.

Ideally many of the intermediate outputs should be piped but some programs don't like piped inputs/outputs and/or writing to compressed files directly (i.e. `bwa mem` can't directly output bam format files, and kraken can't operate in piped inputs)

Issue #238, opened 2016-03-17
tolerate single-end reads throughout pipeline


dpark01:
All throughout the pipeline, there are places that assume paired-end reads, and it's mostly how we wrap underlying tools, probably not the underlying tools themselves. For example, BAM->fastq conversion for bmtagger, Trinity, subsampling, etc, all assumes a pair of fastqs in the output. It's likely that the final alignment and refine steps are fine though (Novoalign, Picard, GATK) since they are entirely BAM-based. It's the steps that have fastq intermediates that are the most likely to be broken. I think we even have a place where we explicitly toss out unpaired reads--we may need to make that an optional boolean flag.

We should tolerate BAMs with single-end reads throughout as well as BAMs that have a mixture of both single and paired end reads. It would make analyses in the field (where sequencing runs might fail halfway through) more robust.

This is a lower priority enhancement for later down the road.

dpark01:
@biocyberman I'm a bit surprised that you ran into issues with unpaired read data on the `refine_assembly` steps (but not at all surprised that there were errors on `deplete_human`). To better assist, do you have some (publicly shareable) example read data? Perhaps a small BAM file of reads that we could use for debugging. We can generate fake data ourselves pretty easily if you don't have something readily available, but it doesn't hurt to try something more realistic.

dpark01:
Hm... looks like https://github.com/broadinstitute/viral-ngs/blob/master/tools/novoalign.py#L180 ought to be changed from `BAMPE` to `BAM`. I think this may be a somewhat recent addition to Novoalign and greatly simplifies its use on mixed data (it doesn't seem to exist in the v3.00 manual). That should solve the `refine_assembly` bits. As far as I can guess, all of the remaining problems with single-end data should be limited to the stuff in `taxon_filter.py`.

@biocyberman if you have example data, we can make the change and codify a unit test around it.

dpark01:
I've made some changes to the `dp-unpaired` branch that should address issues in the lastal filtering step as well as the refine_assembly step. This does not fix any of the bmtagger/blast human depletion steps yet. It is also untested, pending some test case data.

biocyberman:
@dpark01 Here are some ion torrent files. I will leave them online for about a week: https://drive.google.com/drive/u/1/folders/0ByYppRKDpYO_WGlrV3FRR3l3SDA

I will be able to code some parts of the adjustment to accommodate unpaired reads. How do we arrange so that we do not repeat each other?

[...]

PR #583, opened 2017-02-16
handle unpaired read data

dpark01:
This should mostly address issue #238. Unit tests for single-end reads have not been written, though some small data sets have been tried by hand, and existing tests on paired and empty inputs are unbroken.

This also includes some cleanup of legacy fastq commands in various places, and some moving around of code.

The following steps were not previously working on single-end read data: bmtagger, blastn, lastal, trimmomatic, novoalign


tools/mvicuna.py
```diff
@@ -63,6 +63,34 @@ def rmdup(self, inPair, outPair, outUnpaired=None):
...
 
+    def rmdup_single(self, inFastq, outFastq):
+        """
+        Run mvicuna's duplicate removal operation on single-end input reads in
+            fastq format, producing output in fastq format.
+        Notes on weird behaviors of M-Vicuna DupRm:
+            For some reason, it requires you to specify both -opfq and -drm_op even for
+            single-end input.
+
+            Actually, more interestingly, this doesn't work at all.
```
tomkinsc: DumpRm does not work at all on single-end reads?

tools/novoalign.py
```diff
@@ -177,7 +177,7 @@ def align_one_rg_bam(self, inBam, refFasta, outBam, rgid=None, options=None, min
         tmp_sam = util.file.mkstempfname('.novoalign.sam')
         tmp_sam_err = util.file.mkstempfname('.novoalign.sam.err')
         cmd = [self.install_and_get_path(), '-f', one_rg_inBam] + list(map(str, options))
-        cmd = cmd + ['-F', 'BAMPE', '-d', self._fasta_to_idx_name(refFasta), '-o', 'SAM']
+        cmd = cmd + ['-F', 'BAM', '-d', self._fasta_to_idx_name(refFasta), '-o', 'SAM']
```
tomkinsc: We should probably have a comment here noting that the use of `-F BAM` is possible via the newer Novoalign version, since it does not seem to be mentioned in the current reference manual.

tools/samtools.py
```diff
@@ -69,7 +69,7 @@ def view(self, args, inFile, outFile, regions=None):
 
     def bam2fq(self, inBam, outFq1, outFq2=None):
         if outFq2 is None:
-            self.execute('bam2fq', ['-s', outFq1, inBam])
+            self.execute('bam2fq', ['-n', inBam], stdout=outFq1)
```
tomkinsc: Why the change to use stdout over `-s <file>`? Does it conflict with `-n`?

tools/trimmomatic.py
```diff
@@ -15,6 +20,58 @@ def __init__(self, install_methods=None):
...
+        '''Trim read sequences with Trimmomatic.'''
+        trimmomaticPath = self.install_and_get_path()
+        unpairedFastq1 = unpairedOutFastq1 or util.file.mkstempfname()
+        unpairedFastq2 = unpairedOutFastq2 or util.file.mkstempfname()
+        javaCmd = [trimmomaticPath]
+
+        if inFastq2 is None or os.path.getsize(inFastq2) < 10:
+            # Unpaired reads
+            javaCmd.extend([
+                    'SE', '-phred33',
```
tomkinsc: Should we expose this (and the `PE` quality encoding setting) as a parameter?

tools/trimmomatic.py
```diff
@@ -15,6 +20,58 @@ def __init__(self, install_methods=None):
...
+        else:
+            # Paired reads
+            javaCmd.extend([
+                    'PE', '-phred33',
+                    inFastq1, inFastq2,
+                    pairedOutFastq1, unpairedFastq1, pairedOutFastq2, unpairedFastq2
+                ])
+
+        # all the options
+        javaCmd.extend(
```
tomkinsc: Should we explicitly expose the ability to set `-threads <int>`? As per the [manual](http://www.usadellab.org/cms/uploads/supplementary/Trimmomatic/TrimmomaticManual_V0.32.pdf), "If not specified, it will be chosen automatically." On the cluster this could be an issue if it detects the number of system cores rather than the number of cores available to the job.

read_utils.py
```diff
@@ -408,53 +408,8 @@ def main_filter_bam(args):
 
 __commands__.append(('filter_bam', parser_filter_bam))
 
-# =======================
-# ***  bam_to_fastq   ***
-# =======================
-
-
-def bam_to_fastq(
```
tomkinsc: The argument for removal is that users can simply use samtools' `bam2fq`?

assembly.py
```diff
@@ -58,7 +58,7 @@ def __init__(self, n_start, n_trimmed, n_rmdup, n_output, n_subsamp, n_unpaired_
 
 def trim_rmdup_subsamp_reads(inBam, clipDb, outBam, n_reads=100000):
```
tomkinsc: No issues in this function with single-end reads?

tools/novoalign.py
dpark01: weirdly, if you google for "Novoalign manual", there are a few top hits referring to different versions. The one for [v3.07](http://www.novocraft.com/userfiles/file/Novocraft.pdf) describes this. The one for [v3.02](http://www.novocraft.com/wp-content/uploads/Novocraft.pdf) does not.

tools/mvicuna.py
dpark01: I tried valiantly... maybe a bug or something, maybe it was never tested in that scenario. I figure a nop is better than a failure.

tools/trimmomatic.py
dpark01: I think I prefer autodetection where possible, partly because this is really a fastq-specific problem, and the user's input files (BAM) can always be paired, single, or a mixture of both. If it's a mixture, I wouldn't know how to expose the parameter to the user.

read_utils.py
dpark01: Yeah, and also I'm just starting to clean out a few exposed things that are not used in the pipeline (from the user's point of view).

assembly.py
dpark01: Once you fix trimmomatic, apparently not! It's kind of beautiful, especially watching the little report at the end with all the counts.

tools/novoalign.py
tomkinsc: I see...serves me right for assuming the version linked from the Novocraft docs website would be the most recent version.

tools/trimmomatic.py
tomkinsc: Ok, makes sense!

PR #584, opened 2017-02-17
add conda recipe w/ rendering script, refactor dependency files

tomkinsc:
WIP PR, submitted early to kick off long-running tests on Travis.

In this PR, dependencies are restructured so a conda recipe can source dependencies from the same files used for CI testing. It adds a jinja2 conda recipe template, and a script to render the recipe with package lists, along with release tag as version number, and an md5 hash from the tagged release package. conda push is untested, as is branch/tag interaction (since this is the initial commit to GitHub).


tomkinsc:
NB: We'll need to [update](https://readthedocs.org/dashboard/viral-ngs/advanced/) ReadTheDocs to use the new docs dependency file, `requirements-py-docs.txt`.

tomkinsc:
To Do: trigger `viral-ngs-deploy` to build+push the Docker container after the conda build+upload succeeds.

requirements-py-docs.txt
```diff
@@ -0,0 +1,4 @@
+Sphinx==1.4.4
```
dpark01: Did you update the RTD/Sphinx code to use this file instead?

packaging/conda-recipe/viral-ngs-template/meta.yaml
```diff
@@ -0,0 +1,93 @@
...
+    - "read_utils.py -h &> /dev/null"
+    - "reports.py -h &> /dev/null"
+    - "taxon_filter.py -h &> /dev/null"
+
+    # test tool install
+    #- "cd $PREFIX/opt/$PKG_NAME-$PKG_VERSION && py.test -v test/unit/test_tools.py"
+
+    # various unit tests
+    #cd 
+    - "cd $PREFIX/opt/$PKG_NAME-$PKG_VERSION && py.test -v test/unit/test_tools_vphaser2.py"
```
dpark01: is it better to run these individually instead of lumping them all together in parallel (e.g. travis/tests-*.sh)?

requirements-conda.txt
```diff
@@ -23,13 +23,6 @@ vphaser2=2.0
 biopython=1.68
 matplotlib=1.5.3
 future>=0.15.2
-PyYAML=3.11
 pysam=0.9.1
-pycodestyle
-mock==2.0.0
-six<2
-pytest=3.0.5
```
dpark01: I was trying to see if there was any significant negative impact on build time from moving these packages from conda to normal pip install, but it doesn't seem like it: the pip install seems to take < 20 seconds anyway. That said, it was kind of nice that folks could previously set up a dev environment by using conda only (and not having to pip install anything). We could add a note about it, but we probably need to improve the "how to set up your dev environment" documentation anyway, so we could leave that for a separate PR later on.

requirements-py-docs.txt
tomkinsc: Not yet; I figured we'd wait until this PR is merged (though I did note the need to change it in a comment above).

packaging/conda-recipe/viral-ngs-template/meta.yaml
tomkinsc: The test lists in this recipe were copied over from the bioconda version, which as a general rule only tests executability and not function (relying on upstream tests to check the latter). I did include more tests than that basic level, however when running almost everything I has having issues with Travis timing out. This test set seemed like a reasonable compromise, but it's certainly open to revision. There's also the GATK issue.

requirements-conda.txt
tomkinsc: That's true. I'll move these test-related packages back to the conda requirements.

PR #613, opened 2017-02-28
pin tbl2asn, update development notes

tomkinsc:
pin tbl2asn version, also small refactoring of versions in diamond and snpeff


tomkinsc:
Not sure why Travis is not being triggered for this PR...

PR #621, opened 2017-03-08
Fix the nodupes report incorrectly being dupes

yesimon:
Increase picard jvm memory for metagenomics dedup

Increase from the default value to increase stability and speed


dpark01:
I think since the previous integration tests passed with the old JVMmemory, I propose that the integration tests be modified to manually specify the old value, and let the higher value be the new default?

PR #622, opened 2017-03-08
Move pytest plugin requirements to py-tests

yesimon:
Not available in conda

### DIFF
--- a/requirements-conda.txt
+++ b/requirements-conda.txt
@@ -33,7 +33,5 @@ pycodestyle
 mock==2.0.0
 six<2
 pytest-cov==2.4.0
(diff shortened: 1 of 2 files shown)


tomkinsc:
They are in conda-forge, which we rely on.

yesimon:
Gotcha, gotta add it to my channels list!

tomkinsc:
A case could be made for those dependencies to be in the `requirements-py-tests.txt` file anyway, but that file would then need to be listed as `--test-reqs` [in `deploy.sh`](https://github.com/broadinstitute/viral-ngs/blob/master/travis/deploy.sh#L45).

Issue #532, opened 2016-11-14
Expose Picard tool for CollectIlluminaLaneMetrics


tomkinsc:
Similar to how we have a tool to call Picard's `ExtractIlluminaBarcodesTool`, we should consider exposing [`CollectIlluminaLaneMetrics`](https://broadinstitute.github.io/picard/command-line-overview.html#CollectIlluminaLaneMetrics) to write out cluster density metrics for a run directory. This has been requested by members of our lab.

After this has been added, the report should probably be created by default in the [DNAnexus implementation of viral-ngs](https://github.com/dnanexus/viral-ngs).

Related to this, are there other metrics we'd like to report that we are not already writing out?

tomkinsc:
Or we may want `CollectIlluminaBasecallingMetrics` instead.

PR #624, opened 2017-04-18
add tool for CollectIlluminaLaneMetrics

tomkinsc:
This adds a tool to picard.py for CollectIlluminaLaneMetrics, and a parser to illumina.py, `lane_metrics`, for reporting cluster density. This is to address #532.


tomkinsc:
It get the read structure from `RunInfo.xml` if the argument is not specified, same as the other tools.
https://github.com/broadinstitute/viral-ngs/pull/624/files#diff-235f4acb69c1f097332ce7931bf3bd14R184

PR #626, opened 2017-05-25
made assembly more  modular

notestaff:
Broke up assemble_trinity into separate trim_and_rmdup and assembly steps.  Broke up orient_and_impute into separate steps of scaffolding contigs, filling gaps, and imputing from reference.  Added placeholder step for determining a sample-specific reference closest to each sample.  Added config params to choose implementation methods of various steps.  Added various small utils.

From this branch it should be easier to branch out and merge improvements to individual steps without causing conflicts.


dpark01:
Thanks @notestaff -- I will review, though with some extended travel around the holiday weekend (I'm out of town Fri-Tues), I may be a little slow, but I'll try to stay on it!

assembly.py
```diff
@@ -241,7 +241,9 @@ def trim_rmdup_subsamp_reads(inBam, clipDb, outBam, n_reads=100000):
                 os.unlink(f[i])
 
     # multiply counts so all reflect individual reads
-    return (n_input * 2, n_trim * 2, n_rmdup * 2, n_output, n_paired_subsamp * 2, n_unpaired_subsamp)
+    outStats_tuple = (n_input * 2, n_trim * 2, n_rmdup * 2, n_output, n_paired_subsamp * 2, n_unpaired_subsamp)
+    if outStatsFile: util.file.dump_file(outStatsFile, outStats_tuple)
```
tomkinsc: Maybe this should be a csv/tsv file that includes a header to disambiguate the column values. Something like this?
```python
with open(outStatsFile, 'w') as f:
    csvwriter = csv.writer(f, delimiter='\t', quoting=csv.QUOTE_MINIMAL)
    csvwriter.writerow(["paired_input","paired_after_trimmomatic","paired_after_prinseq_rmdup","individual_output_reads", "paired_output_reads", "unpaired_output_reads"])
    csvwriter.writerow(outStats_tuple)
```

util/file.py
```diff
@@ -473,3 +494,36 @@ def line_count(infname):
...
+        pass
+
+def dump_file( fname, value ):
+    """store string in file"""
+    with open( fname, 'w' )  as out: 
+        out.write( str( value ) )
+
+def slurp_file( fname, noErr=False ):
+    """Read entire file into one string."""
+    if not fname or not isinstance(fname,str) or not os.path.isfile( fname ):
```
tomkinsc: Maybe frame this as a try/except, with a `return None` in the exception handler if `noErr=True`

util/file.py
```diff
@@ -473,3 +494,36 @@ def line_count(infname):
...
+    """store string in file"""
+    with open( fname, 'w' )  as out: 
+        out.write( str( value ) )
+
+def slurp_file( fname, noErr=False ):
+    """Read entire file into one string."""
+    if not fname or not isinstance(fname,str) or not os.path.isfile( fname ): 
+        if noErr: return None
+        else: raise IOError( 'File not found: %s' % fname )
+    with open( fname ) as f:
```
tomkinsc: To be consistent within viral-ngs, maybe use`open_or_gzopen()`

assembly.py
```diff
@@ -278,45 +284,42 @@ def assemble_trinity(
         First trim reads with trimmomatic, rmdup with prinseq,
         and random subsample to no more than 100k reads.
     '''
-    if outReads:
-        subsamp_bam = outReads
+
+    if already_preprocessed_bam:
+        subsamp_bam = inBam
+        read_stats = eval(util.file.slurp_file(preprocessed_bam_stats, noErr=True) or '(0,0,0,0,0,0)')
```
tomkinsc: `eval()` is generally considered [dangerous](https://nedbatchelder.com/blog/201206/eval_really_is_dangerous.html). Even if our input here is usually trusted, can we read in `preprocessed_bam_stats ` in a different way? If it is encoded in CSV/TSV format we can use the Python `csv` module, or else we could encode as JSON/YAML. Even parsing with a regular expression would be preferable to `eval()`

assembly.py
```diff
@@ -278,45 +284,42 @@ def assemble_trinity(
...
-            log.warn("denovo assembly (Trinity) failed to assemble input, emitting empty output instead.")
-            util.file.touch(outFasta)
-        else:
-            raise DenovoAssemblyError(*read_stats)
-    os.unlink(subsampfq[0])
-    os.unlink(subsampfq[1])
-
-    if not outReads:
-        os.unlink(subsamp_bam)
+    with tools.picard.SamToFastqTool().execute_tmp(subsamp_bam, sfx='.subsamp', illuminaClipping=True) as subsampfq:
```
tomkinsc: I like the pattern of using a context manager for intermediate fastq files from bam precursors. It would be nice to have a (perhaps optional) post-`yield` cleanup of the temp files in the context manager.

notestaff: Cleanup is here: https://github.com/broadinstitute/viral-ngs/pull/626/files#diff-3db8e9fe58bc71080acfba25411aa600R113

assembly.py
```diff
@@ -278,45 +284,42 @@ def assemble_trinity(
...
     parser.add_argument(
         '--n_reads',
         default=100000,
         type=int,
-        help='Subsample reads to no more than this many pairs. (default %(default)s)'
+        help='Subsample reads to no more than this many pairs. Ignored if --already_preprocessed_bam.  (default %(default)s)'
     )
-    parser.add_argument('--outReads', default=None, help='Save the trimmomatic/prinseq/subsamp reads to a BAM file')
+    parser.add_argument('--outReads', help='Save the trimmomatic/prinseq/subsamp reads to a BAM file')
+    parser.add_argument('--already_preprocessed_bam', default=False, action='store_true',
```
tomkinsc: Thoughts on renaming this argument to something like `--doNotTrimRmdupAndSubsample` (or `--doNotProcess`) and/or changing the help text to explain that when the option is not specified the input bam will be trimmed, rmdup'd, and subsampled?

pipes/config.yaml
```diff
@@ -192,6 +192,27 @@ vphaser_max_bins: 10
 # allele freq > 0.005. If false, no filtering is performed.
 vcf_merge_naive_filter: false
 
+# Preprocessing method for reads for de novo assembly
+assembly_preproc_method: 'subsamp'
+
+# Method of de novo assembly
+assembly_de_novo_method: 'trinity'
+
+# Method of choosing the reference to use for scaffolding
```
tomkinsc: Maybe list the method choices here?

pipes/config.yaml
```diff
@@ -192,6 +192,27 @@ vphaser_max_bins: 10
...
+refchoice_for_scaffold_method: 'common'
```
tomkinsc: Can we spell out "reference"? 
`reference_choice_for_scaffold_method `
or maybe:
`reference_selection_method_for_scaffolding`

pipes/config.yaml
```diff
@@ -192,6 +192,27 @@ vphaser_max_bins: 10
...
+
+# Method of scaffolding contigs
+scaffold_method: 'scaffolded'
+
+# Method of filling gaps between contigs from the reads
+gapfill_method: 'none'
+
+# Method of choosing the reference to use for imputation
+refchoice_for_impute_method: 'common'
```
tomkinsc: Can we spell out "reference" in full here too?

assembly.py
notestaff: The name was by analogy with  already_realigned_bam in refine_assembly.   Ideally I'd remove it from assemble_trinity and just have  separate  commands  for  preprocessing and assembly, per [unix philosophy](http://www.faqs.org/docs/artu/ch01s06.html).  But I thought removing the preprocessing part might break backwards compatibility.

pipes/rules/assembly.rules
```diff
@@ -30,67 +30,116 @@ rule all_assemble_failures:
...
+    run:
+            shell("{config[bin_dir]}/assembly.py trim_rmdup_subsamp {input} {params.clipDb} {output.subsamp_bam}  --outStatsFile={output.subsamp_bam}.stats --n_reads={params.n_reads}")
 
 rule assemble_trinity:
     ''' This step runs the Trinity assembler.
         First trim reads with trimmomatic, rmdup with prinseq,
         and random subsample to no more than 100k reads.
     '''
-    input:  config["data_dir"]+'/'+config["subdirs"]["per_sample"]+'/{sample}.taxfilt.bam'
+    input:  config["tmp_dir"]+'/'+config["subdirs"]["assembly"]+'/{sample}.'+config.get("assembly_preproc_method","subsamp")+'.bam'
```
tomkinsc: I do wonder if the preprocessing method should be a parameter of `assemble_trinity()` (or `assemble(aligner='trinity', preprocessing='subsamp')` as we could refactor it), so we can keep the logic out of Snakemake and in the core viral-ngs code.

pipes/config.yaml
notestaff: Right.  On this branch there's only one method choice  for each stage, since I just added placeholders.  Plan was to merge this branch and then make pull requests for additions of individual methods; these pull  requests would  then be less  likely to create merge conflicts  since they'd affect different placeholders.

pipes/config.yaml
notestaff: Sure.   Just thought  it  easier to have a  one-word  verb  for each stage: scaffolding, assembly, refchoice :)

pipes/rules/assembly.rules
notestaff: Can do that.   For the pipeline, I'd keep preprocessing as a separate step, since e.g. multiple assemblers might use  the preprocessed  reads.  I also like the Unix way of having well-defined individual commands that can  be chained.   But if people are  also routinely running individual CLI tools, then aggregate commands like preprocess-and-assemble make sense.

assembly.py
tomkinsc: Gotcha. `already_realigned_bam` is a file path rather than a boolean though.

We try to be modular and unixy with the viral-ngs python functions, but also expose convenience functions that "pre-chain" functions that are (almost always) called together, like subsampling (or preprocessing) as an inherent part of the `assemble_trinity()`call.

I'd probably vote for keeping each *de novo* assembly preprocessing method as a separate function (exposed via CLI), like how the subsample function is now in `master`, with invocation by default in `assembly(aligner='trinity')` so the most common use case is covered without additional input from the user (with option to disable like you have). This allows savvy users to chain things in a unixy way, but provides "easy" methods for the most common use case.

pipes/config.yaml
tomkinsc: That's fair. I generally tend to lean toward being verbose and explicit, since we have high resolution monitors now. :)

notestaff: Could params be grouped hierarchically, by stage?

assembly:
    - preproc_method: 'subsamp'
    - de_novo_method: 'trinity'
scaffolding:
    - reference_selection_method: 'common'
    - scaffolding_method: 'viral-ngs-scaffolder'
finishing:
    - gapfill_method: 'gap2k'

etc

tomkinsc: Absolutely! 👍 Just have to find & update the various places where we access the params.

notestaff:
@tomkinsc re: coveralls, other people have also recently had problems:
https://github.com/lemurheavy/coveralls-public/issues/281
not sure it's the same problem, but the link points to an alternative to coveralls

dpark01:
Sorry I'm late to the game, but one quick question and one more comment as I orient myself.

Question: should I be looking at #627 before this one? That is, should we sort that out first, because some of that stuff appears to be in here as well?

Comment: There's a bit of discussion here about breaking out the pre-processing steps to assembly (trim/rmdup/subsample). Makes a lot of sense, but I'll describe my own thinking and hesitations around that. In general, I like modularity, particularly at the code level (python functions). I also like simplicity at the UI level (command-line accessible subcommands), which sometimes implies wrapping together steps that always want to happen together regardless, and also presenting the units of work to the user in logical chunks that correspond to how they scientifically think about it. So there's some tension. Also throw in the additional consideration that I prefer not to put important logic in the snakemake code itself (keep it within the python code) because it is only one possible way that our code gets run.

But here's the bigger issue around the pre-assembly processing steps that led me to bundle it with trinity's execution. It has to do with the fact that I've strongly pushed our code to operate only on BAM files from a UI perspective. There are reasons for this I can explain elsewhere, but it's pretty important to me. Each of the commands we provide that operate on and emit BAM files always preserve all the metadata that it was presented with. So all the way through to the end, all read group info is maintained, and all of the various tags and such remain in there. The exception to this is the output of the trim/rmdup/subsamp. While rmdup and subsamp could be done in a lossless way (in terms of metadata), trimming reads is inherently destructive and I don't think there's a great way to preserve information there. As such, the BAM output of this step is the only time in our entire code base that we ever call Picard's FastqToSam. Notice that it's also not even written out by default: if someone runs assemble_trinity with default parameters, it consumes the reads and emits a fasta, and no one ever sees the molested FastqToSam output. That said, looking at that output is useful for debugging things, so all of the pipeline wrappers (Snakemake, DNAnexus) around our code save this file anyway for future inspection and debugging. But I sincerely hope that no one ever uses it for anything important other than debugging.

So that's one of the main reasons why I had avoided breaking it entirely apart. If you do break apart the pre-processing from the assembly, the FastqToSam output becomes a first class citizen input/output file in the pipeline.

notestaff:
@dpark01   I've merged #627 into #626 since both have to do with modularizing assembly, so can just look at #626

notestaff:
My understanding was that files under data/ are "official" inputs/outputs, while files under tmp/ are "unofficial/under-the-hood/use-at-your-own-risk" stuff that can be anything.   Is that incorrect?

notestaff:
"trimming reads is inherently destructive" -- if Trimmomatic loses information, can we add code after Trimmomatic to put the information back?  The read names after trimming still match to the original reads, so the read groups and SAM tags can be restored.

notestaff:
"I like modularity, particularly at the code level (python functions). I also like simplicity at the UI level (command-line accessible subcommands)," -- he Snakemake pipeline used CLI, but is itself used as a black-box UI.   So does it really break abstraction, to use "internal" CLI commands in pipelines?  Could it be a consistent description to say, the UI consists of "public" CLI commands and what pipelines put in data/, while internals consist of "private" CLi commands and what pipelines put in tmp/?

notestaff:
@dpark01 a question about degree to which backwards compatibility should be maintained: if I change the names of some files which the pipeline creates under tmp/, would that cause problems for existing users?

dpark01:
re: #627 I prefer looking at smaller units of change in the diffs, so if #627 is a subset of the changes in #626 then I'll probably review that first, merge it, then this branch can merge back and have a smaller, more topically focused diff.

data-vs-tmp: in the snakemake description of the pipeline, "data" are the main inputs and outputs of each sample that you generally care about when everything is working well, and "tmp" files are not unstable nor unsupported--on the contrary, they're meant for scientist consumption, especially when something goes wrong. (this is not to be confused with the --tmp_dir that our python code creates for ephemeral stuff like fastqs and stuff no one should care about seeing.. this is actually often on local instance storage, not permanent/network storage). But at the end of a large analysis / project, it's expected that the user wouldn't really care about what's in tmp longer term (they'd probably archive and backup everything but that directory). In the DNAnexus incarnation of the pipeline, this directory is called "intermediates".

None of the *pipelines* are meant to be black box to the user / scientist. They can be executed as a single invocation for simplicity / scalability. But each snakemake rule or DNAnexus applet (or WDL task eventually) is meant to encapsulate the level of abstraction we want to present, and in most cases, that's one viral-ngs CLI command per box (but there are some exceptions to that). Currently, none of the CLI commands we expose are intended to be private/internal. Private/internal things tend to be straight python functions. Take for example, some of the taxonomic depletion commands. We have a few python functions that will take bam in, write bam out, and run bmtagger or blast or such against a database. But we don't expose those as CLI -- we expose calls to `multi_db_deplete_bam` wrapped around it, with all the fastq conversion machinery hidden away. But it lets us as developers glue together various combinations of these bits and pieces with different tools.

From a UI point of view, I would envision that we would have a denovo_assembly command that takes reads (BAM) in and spits contigs (FASTA) out, and let's you specify from a few different assemblers, along with various tool-specific options. Same for scaffolding. The main thing we need to sort out is whether there's any tool-specific reason to want to change how pre-processing of reads prior to assembly is done. My current intuition is that most of the pre-processing is scientifically driven, not Trinity-driven, and we'd probably keep it mostly the same no matter what assembler people choose.

I do somewhat like the idea of breaking out scaffolding from the post-scaffolding (impute / gapfill).. though part of me would probably lump the last two together as a single filling step, presenting options to the user about which strategies (if any) they want to employ to fill out the empty parts of the genome. These two steps are already separate in the CLI, but not in the snakemake or DNAnexus wrappers. One caveat to keep in mind though is that each time you break up a step, you're incurring more cluster or cloud queuing and instantiation overhead, and some of these steps only take a few minutes of runtime in the average case.

re: backwards compatibility with filenames in tmp/ -- we may need to change how code is called elsewhere (e.g. by DNAnexus wrappers) but if it's a sensible/necessary change (like changing ".trinity.fasta" to ".denovo.fasta") we should probably just do it.

re: trimming reads being destructive: basically, the BAM file philosophy is such that no tool should ever change the sequence of a read. Trimming and clipping a BAM file happens by applying tags to each read that indicate the positions that should be clipped--I don't know of any tool out there that actually hard trims reads in a BAM file. Nothing really stops us from being the first ones to try though, but the painful part would be that if you delete the first few bases of a read, you'll have to manually shift any metadata tags on that read that refer to a bp position on that read, which doesn't sound fun. It also seems relatively unnecessary to go through all that effort to preserve all that in this case because we're only doing this prior to feeding it to a de novo assembler, which consumes the read data, but wouldn't care about such metadata.

@notestaff @tomkinsc I'll be continuing to look at this from Boston next week. Let me know if you ever want me to call you guys in Florida about any of this.

notestaff:
Thanks  @dpark01 , that clarifies the design.  Maybe add these notes to DEVELOPMENT_NOTES.md .

Finer-grained pipeline steps are useful when exploring the parameter space.  E.g. assemble_trinity and assemble_spades both start with trimmed/rmduped reads, and factoring out trim_rmdup lets one try out different params for assembly_* without redoing trim_rmdup.  But that could be solved with a caching-to-tmp_dir mechanism instead.   It's just, why add a separate mechanism when normal Snakemake operation already does it.  Also, if "pre-processing is scientifically driven" then wouldn't its output be "meant for scientist consumption"?  If assembly fails we report stats from the preproc step; these stats and output might be useful for tuning non-failed runs too.

Some downstream steps like refine could use trim_rmdup reads in lieu of .cleaned ones.  Is there a reason they don't?

"none of the CLI commands we expose are intended to be private/internal" -- private commands needn't be exposed by default; they can be in a separate parser.  But for exploratory analysis it can be useful to have finer-grained steps accessible via CLI.

" "tmp" files are not unstable nor unsupported--on the contrary, they're meant for scientist consumption, especially when something goes wrong." -- there could be, in addition to those, internal files not _typically_ meant for scientist consumption, but occasionally useful, under tmp/internal/ .  The exposed interface includes the  config file and the filenames.   Does it also include the exact set of rules?   Normally one sets config params and asks the pipeline to produce certain target filenames; whether a target is made in one step or two seems more implementation detail than interface.

dpark01:
Ah, if there's ever a reason to explore parameter space in assembly, we should definitely promote the preprocessing bam and break that step out. That totally makes sense. Does spades have parameters to explore? Trinity doesn't have much that was ever worth exposing to the user--it either works or doesn't, and that's mostly a function of the input reads.

Yeah we'd never want to use the trim/subsamp reads for assembly refinement for multiple reasons... can discuss more later. We used to use the raw reads actually, but have been using cleaned reads since about 2014 or so.

notestaff:
Choice of assembler(s) is one parameter.  No assembler dominates all others on all samples.  For small genomes can just always run all assemblers and combine the contigs, eliminating the assembler-choice parameter; but I'm not yet sure this would _never_ confuse the scaffolder.

(Side note: we sometimes get contigs that aren't from target genome, e.g.  I've seen influenza contigs assembled from HIV samples.  Maybe worth adding a contig filtering step, based on BLAST results for contigs, before scaffolding them?  Taxon filtering of reads could then be more permissive.)

Related to choice of assembler, some assemblers can take other' output as a starting point.  Whether to do that, and how much to trust the earlier assembler's contigs, is a parameter in SPAdes.

kmer size is an important parameter of many assemblers, to the point that there are papers on automatically choosing this parameter.   (There's also some work on multi-scale de Bruijn graphs that use multiple kmer sizes in one graph, but I haven't yet seen assemblers based on that.) Like with running different assemblers, could run the same assembler under a range of values then combine contigs.

I agree that external interface should be kept simple and knobs minimized.  I just wanted to understand what, exactly, counts as "external interface".  If it's the generated docs and the help messages and the documented input/output files and the DNAnexus steps, then maybe it's ok to have internal/advanced CLI commands and internal pipeline steps as long as the user-visible interface is unchanged.  But if the Snakemake rules themselves are part of the external interface, then adding internal stuff would clutter that.

notestaff:
Re: exploring assembly parameters, there are some reference-free tools that score the quality of contigs given the reads:
http://hibberdlab.com/transrate/
http://genome.cshlp.org/content/26/8/1134
http://deweylab.biostat.wisc.edu/detonate/
So could assemble under several parameter settings then pick best contigs by these metrics.

notestaff:
Re: pipeline organization, in current version of this branch, the output of a given phase doesn't have a standard filename; next phase gets the filename based on params, e.g. output of de novo is either sample.assembly1-trinity.fasta or sample.assembly1-spades.fasta .  May be better to have standard names for the output of each stage, regardless of method, e.g. sample.assembly1-denovo.fasta .

notestaff:
"each time you break up a step, you're incurring more cluster or cloud queuing and instantiation overhead" -- for Snakemake the localrules: directive might mitigate that; not sure with WDL.

dpark01:
If you're interested in seeing more about how this is utilized by scientist-users, and the UI that is exposed and interacted with, you might consider attending (as an observer, instead of a trainee) a weeklong training I'm giving from 6/12-6/16 at the Broad.

Snakemake localrules pushes execution onto the head node and doesn't parallelize anything, right?

Scaffolding starts with an aligner already, so there was never a reason to taxonomically filter things again (and the lastal taxon filter step was always deliberately promiscious.. we see all sorts of contigs come out of trinity that don't bother us because they get tossed at scaffolding). That said, if you're trying to introduce a new step (Sealer) between de novo assembly and reference-based scaffolding, then I can see a reason to add another filtration step on the contigs. But even then, I don't think it's necessary--the computational expense of sealing a few unintended contigs isn't really that costly.

notestaff:
Workshop would be great, is there a signup page?

Stray contigs can affect the determination of closest reference, which is now based on the set of kmers in contigs.  So stray contigs might themselves not align, but cause us to pick a reference to which other contigs and reads align less well.

notestaff:
I'll close this pull request and make some new, more granular ones.

PR #627, opened 2017-05-25
Added ability for config files to include other config files. 

notestaff:
Added ability for config files to include other config files.   Useful for adding new config params while preserving backwards compatibility; params from pipes/config.yaml get included implicitly, if not overridden by user's config file.


notestaff:
@tomkinsc about coveralls issues: could the cause be that on a pull request that is not on the master branch, unit tests aren't run, but on a master branch they are?  So code that's covered only by unit tests shows a drop in coverage, for a pull request on a non-master branch.

dpark01:
Sorry folks, I could've told you that, yes that's exactly the known reason why coveralls is flaky for PRs ever since we tried to speed up the build.

I promise I'll dig in more time into these PRs tomorrow, I'm still catching up from the out of town travel

notestaff:
To speed up build, could skip tests when code they use hasn't changed?  strace can tell which files are used; python's trace module can tell which python functions.

notestaff:
Could try https://github.com/tarpas/pytest-testmon or https://pypi.python.org/pypi/pytest-incremental .

travis/tests-unit.sh
```diff
@@ -1,7 +1,9 @@
 #!/bin/bash
 #set -e
 
-if [ $TRAVIS_PULL_REQUEST == "false" ]; then
+UNIT_TESTS_ALWAYS=1
```
tomkinsc: This is for debugging CI coverage/testing?

tomkinsc:
Thoughts on making this a PyYAML `Loader` (meta)class?
https://stackoverflow.com/a/9577670
https://gist.github.com/joshbode/569627ced3076931b02f

travis/tests-unit.sh
notestaff: @tomkinsc This fixes the issue of coveralls coverage drop, at the cost of again doing redundant computations on travis.  Maybe it could stay this way, till we figure out how to avoid redundant computation without causing incorrect coverage measurements?  @yesimon ?

notestaff:
Using Loader would put the included mappings under a named node, so they'd have to be accessed under that node's name: config["included_config_A"]["key"] rather than just config["key"].

travis/tests-unit.sh
dpark01: Yeah I can go either way on this. On one hand I thought this switch last year seemed worthwhile even though it broke coveralls for PRs (but not for commits). But I can see the reason to want to keep that meaningful at the expense of build time. My suggestion, if we revert back to the original as Ilya proposes, would be to reverse the order so that the unit tests fire after the integration tests so that at least we can hit failures at the same speed as before (and Travis turns red and sends email notifications on the first fail, it doesn't wait for the build to finish).

util/misc.py
```diff
@@ -410,4 +412,106 @@ def which(application_binary_name):
...
+    '''Return a tuple containing the items in `x`, or containig just `x` if `x` is a non-string sequence.  Convenient
+    for uniformly writing iterations over parameters that may be passed in as an item or a tuple/list of items.
+    '''
+    return tuple(x) if is_nonstr_seq(x) else (x,)
+
+def flatten_dict(d):
+    '''Return a new dict `r`, where r[(k1,k2,...,kn)]==v iff d[k1][k2]...[kn]==v.'''
+    result=dict()
+    for k, v in d.items():
+        assert not isinstance(k,tuple), 'Dict already flattened? {}'.format(d)
```
yesimon: This is a strange API. You can't flatten dicts with tuples in them?

util/misc.py
```diff
@@ -410,4 +412,106 @@ def which(application_binary_name):
...
+
+def make_seq(x):
+    '''Return a tuple containing the items in `x`, or containig just `x` if `x` is a non-string sequence.  Convenient
+    for uniformly writing iterations over parameters that may be passed in as an item or a tuple/list of items.
+    '''
+    return tuple(x) if is_nonstr_seq(x) else (x,)
+
+def flatten_dict(d):
+    '''Return a new dict `r`, where r[(k1,k2,...,kn)]==v iff d[k1][k2]...[kn]==v.'''
+    result=dict()
```
yesimon: Spaces between = (and elsewhere)

util/misc.py
```diff
@@ -410,4 +412,106 @@ def which(application_binary_name):
...
+    for k, v in d.items():
+        assert not isinstance(k,tuple), 'Dict already flattened? {}'.format(d)
+        if isinstance(v, collections.Mapping):
+            for k_rest, v_actual in flatten_dict(v).items():
+                result[(k,)+k_rest]=v_actual
+        else:
+            result[(k,)]=v
+    return result
+#    return dict(((k,)+k1,v1) for k, v in d.items() for k1, v1 in \
```
yesimon: Delete?

util/misc.py
```diff
@@ -410,4 +412,106 @@ def which(application_binary_name):
...
-            return link_resolved_path
\ No newline at end of file
+            return link_resolved_path
+
+def is_nonstr_seq(x):
+    '''Tests whether `x` is a Sequence other than a string'''
+    return isinstance(x, collections.Sequence) and not isinstance(x,(str,bytes))
+
+def make_seq(x):
+    '''Return a tuple containing the items in `x`, or containig just `x` if `x` is a non-string sequence.  Convenient
```
yesimon: Typo

yesimon: Add comment that this concretizes iterators

test/unit/test_util_misc.py
```diff
@@ -176,3 +178,69 @@ def testMultiChrWithPayloadIntervals(self):
...
+            z = rand_dict()
+            z_flat=util.misc.flatten_dict(z)
+            assert all( isinstance(k,tuple) and isinstance(v,int) for k,v in z_flat.items())
+            z_unflat=util.misc.unflatten_dict(z_flat)
+            z_reflat=util.misc.flatten_dict(z_unflat)
+            
+            self.assertEqual(z_unflat, z)
+            self.assertEqual(z_reflat, z_flat)
+
+class TestConfigIncludes(unittest.TestCase):
```
yesimon: Also test empty included config files. Yaml load returns None for empty files which is a gotcha.

util/misc.py
```diff
@@ -410,4 +412,106 @@ def which(application_binary_name):
...
+        if fname.upper().endswith('.YAML'): return yaml.safe_load(f)
+        if fname.upper().endswith('.JSON'): return json.load(f)
+        raise TypeError('Unsupported dict file format: ' + fname)
+
+def load_config(cfg, include_directive='include', std_includes=[], param_renamings={}):
+    '''Load a config file, recursively loading any included config files; load any params specified under legacy names.
+
+    This function implements extensions to the standard way of specifying configuration parameters via (possibly nested)
+    dictionaries.  The extensions let config files include other config files, and let us rename parameters while
+    still letting them be specified under old names for backwards compatibility.
```
yesimon: I think this needs more description. Reading the code there are two things undocumented.

* Included filenames can be the filename without path if they are in the same directory as cfg AND cfg is a filename.
* Param renaming renames parameters in cfg.
* Order of include overrides is std_includes, include_directive, then finally cfg.

pipes/rules/common.rules
```diff
@@ -1,13 +1,26 @@
...
 # imports needed for download_file() and webfile_readlines()
 import re
 # since py3 split up urllib
 try:
     from urllib.request import urlopen
 except ImportError:
     from urllib2 import urlopen
+
+# Ensure that Snakemake rules can import viral-ngs python code
+sys.path.append(os.path.realpath(os.path.join(os.path.expanduser(config['bin_dir']))))
```
yesimon: Might there be a better way of importing the utiltiy functions into the rule? @tomkinsc

util/misc.py
notestaff: You can in general.  The utila here are just subroutines for handling config dicts, which always have string keys, while flattened dicts always have tuple keys; this can be used as a sanity check.  I'move it into a unit test.

notestaff:
thx @yesimon for detailed review, will fix.

pipes/rules/common.rules
tomkinsc: Eh, there's importlib:
```
from importlib.machinery import SourceFileLoader
foo = SourceFileLoader("module.name", "/path/to/file.py").load_module()
```

notestaff: @tomkinsc What makes importlib preferable?   Seems like it involves loading each module manually/individually?

tomkinsc: Not necessarily preferable, just another option.

notestaff: Other option is to put code into common.rules directly, rather than importing it from util.misc; that's what I was doing originally.   This keeps the wall between Snakemake rules and CLI code, but can lead to some code duplication.

Modifying a system variable like sys.path directly feels like a bit of a hack, but I can't think of trouble down the road this would cause.

notestaff:
@yesimon @tomkinsc @dpark01 I've added some commits to address the comments so far; let me know if you notice more issues.

pipes/rules/common.rules
tomkinsc: Should we prepend it instead for higher precedence on the path? 

`sys.path.insert(0, <path/to/add>)`

notestaff: I figured append has the least chance of creating unexpected interactions?  The value can't mask any othet lookup, and only gets used if the next step would be error.

notestaff:
@yesimon @tomkinsc @dpark01 Are there more things that need to be fixed?

dpark01:
Sorry it's been so long and more than one of us have been traveling so let's meet to discuss tomorrow and close this out. I'll send an invite.

tomkinsc:
This has the potential to break our simplistic yaml parsing in the cluster-related shell scripts:
ex. https://github.com/broadinstitute/viral-ngs/blob/master/pipes/Broad_UGER/jobscript.sh#L8
Probably just need to update the scripts.

util/misc.py
```diff
@@ -410,4 +412,118 @@ def which(application_binary_name):
...
+    return result
+
+def load_yaml_or_json(fname):
+    '''Load a dictionary from either a yaml or a json file'''
+    with open(fname) as f:
+        if fname.upper().endswith('.YAML'): return yaml.safe_load(f) or {}
+        if fname.upper().endswith('.JSON'): return json.load(f) or {}
+        raise TypeError('Unsupported dict file format: ' + fname)
+
+def load_config(cfg, include_directive='include', std_includes=[], param_renamings={}):
```
dpark01: change default std_includes to () or to None to avoid issue with mutable default input

tomkinsc: Probably shouldn't set the default to a mutable type; better to set it to `None` and then in the scope of the function initialize it as empty:

```
if std_includes is None:
        std_includes = []
```

notestaff:
@tomkinsc "This has the potential to break our simplistic yaml parsing in the cluster-related shell scripts" -- this only breaks if the user, rather than specifying conda_env_dir/miniconda_dir/bin_dir/data_dir directly in ./config.yaml , tries to inherit the value from another file; and in that case, the code fast-fails.   We're not yet documenting the include-config ability for the user; it's being added for internal use for algorithm-knob params, but would not be used for directory params.  Given this, what scenario could cause a problem?

tomkinsc:
If we stub out a `config.yaml` files that omits directories it would be breaking (i.e. if we update `easy-deploy-viral-ngs.sh` to assume that the `config.yaml` should be imported rather than explicitly included in each project directory). Maybe we should split the config file into two: one with paths (that we still copy into projects), and one with algorithmic knobs to turn. The former could import the latter. Alternatively we could include your YAML import ability in the bash/python one-liners, but that may be inelegant.

notestaff:
@tomkinsc   "If we stub out a config.yaml files that omits directories" -- I was thinking that could be a separate, future pull request.  The stubbing-out isn't an essential element of this PR.  This PR as-is enables specific functionality, the ability to add and rename/regroup algorithm-knob params without breaking compatibility.   What about merging this functionality now, and doing a separate PR for more related functionality later?

tomkinsc:
Yeah, that's reasonable. Let's make a GitHub issue though so we remember we can change the easy-deploy behavior to use an included yaml config rather than a copied yaml file (and that notes it would mean an update to the various shell scripts mentioned above).

tomkinsc:
Maybe the docs should mention this new config import functionality in the [section](https://viral-ngs.readthedocs.io/en/latest/pipeuse.html#setting-up-an-analysis-directory) on setting up a Snakemake analysis directory.

PR #631, opened 2017-06-05
added refsel - reference selection of closest reference to each sample

notestaff:


dpark01:
I'm excited for this direction of work, but I don't think this branch is quite ready for a PR yet.

1. I think we should finish up #627 before doing this one, so that the diff is topically clean.
2. The addition of two new Tools look like they don't have new corresponding entries in requirements-conda.txt or basic binary execution tests in tests/unit. @tomkinsc can discuss a few of the basics with you this week. But I would consider those prerequisite to a PR.
3. Most of all, this adds a decent amount of new custom-written code that certainly should be unit tested. A few simple test cases of input that reliably shows that it can pick the sensible closest reference out of a tiny test database. Including some multi-segment virus examples.
4. How does this handle (if at all) multi-segment viruses? If it doesn't, it's not ready for PR. If it does, there's discussion to be had around the UI of it, because I've thought about it a bit in the past, and there's a few possible approaches (user provides a bunch of fasta files, one per genome, user provides a bunch of fasta files, one per segment/chromosome, user provides an interleaved fasta and a number-of-chromosomes-per-genome integer, or a non-interleaved fasta, etc). I think I've leaned towards single fasta, possibly interleaved... but then it'd be up to our code to split it all out and deal with it properly under the hood.

Separately, is a k-mer counting strategy advantageous over a more simplistic BLAST (or BLAST-like) approach, especially since the databases will always be tiny (compared to `nr`) and maybe we can even save a computation by utilizing our LASTAL output.

I think this also adds a top level python script that isn't represented in the coveralls invocations in the Travis scripts.

I think this branch should continue development, with discussion above the above items offline, and we can re-open the PR at a later time?

notestaff:
Makes sense.  It does handle multi-segment genomes, where the accessions file lists ids as 
ref1_seg1
ref1_seg2
ref2_seg1
ref2_seg2

I'll add unit tests and try simpler BLAST-based strategies.   Reason for not using lastal was that  its accessions sometimes include fragments of genomes or just one segment of a multi-seg genome, which is fine for taxon-filtering reads but not for reference selection.

notestaff:
p.s. reason for kmer approach is, for a set of contigs it's simple to define the match criterion (set of kmers in the set of contigs contains largest number of unique/rare kmers from which ref?); combining blast scores from a bunch of possibly overlapping/duplicative contigs into one score is less straightforward.

notestaff:
I'll close this pull request for now, will re-open when have more tests.

PR #633, opened 2017-06-09
Fix two issues in using bwa mem for alignments

haydenm:


tomkinsc:
Is the bwa `-T` [issue](https://github.com/broadinstitute/viral-ngs/pull/633/files#diff-6f446f26625bfa311ddb27610d17a840L172) no longer present?

haydenm:
I think I had been misunderstanding bwa's behavior in filtering alignments based on `-T`. From what I understand now, an alignment score is per-mate. If one mate has a valid score, then both mates are output in the alignment, even if the other has a score falling below the threshold. So that's why it may output alignment scores that fall below the `-T` value. The documentation is not clear on this and the behavior is probably debatable. Regardless, I think what bwa is doing is valid and is not necessarily incorrect.

yesimon:
I think bwa mem adds together the AS from both mates to compare with -T.

notestaff:
novoalign also adds a term based on insert length, not sure if bwa does

PR #635, opened 2017-06-13
added various file-handling utils, and tests for them

notestaff:
Added some small utils for handling unpaired reads in .bam files, and for using context managers to temporarily convert .bam files to fasta/q for processing.   These utils are used by some other branches I plan to check in.

### DIFF
--- a/tools/picard.py
+++ b/tools/picard.py
@@ -8,6 +8,7 @@ import os.path
 import tempfile
 import shutil
 import subprocess
+import contextlib
 from decimal import *
 
 import pysam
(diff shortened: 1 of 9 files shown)


util/file.py
```diff
@@ -473,3 +494,33 @@ def line_count(infname):
...
+    '''Make each file in fnames a zero-length file with the current timestamp.'''
+    with open(fname, 'w'):
+        pass
+
+def dump_file( fname, value ):
+    """store string in file"""
+    with open( fname, 'w' )  as out: 
+        out.write( str( value ) )
+
+def slurp_file( fname ):
```
dpark01: I'm a little worried that `slurp_file` might get used a lot when it shouldn't be.. I realize it's convenient at times to take a quick pull of inputs that we know to be O(1) in size... but putting it here almost makes it to convenient for developers to avoid good iterator practice.

util/file.py
```diff
@@ -473,3 +494,33 @@ def line_count(infname):
...
+def tabjoin( *args, **kwargs ):
+    """Join args by tab"""
+    sep = kwargs.get( 'sep', '\t' )
+    return sep.join( map( str, args ) )
+
+def tabwriten( f, *args, **kwargs ):
+    """Write a tab-separated line to a file"""
+    f.write( tabjoin( *args, **kwargs ) )
+
+def tabwrite( f, *args, **kwargs ):
```
dpark01: at one point in history I think I had a nice tab file writer in here that I later removed as a part of purging unused code... that said, I feel like the normal way to do it is so terse in python anyway that it was less useful than the file reading functions were. Just a random comment, no request for changes here.

util/file.py
```diff
@@ -94,6 +94,27 @@ def mkstempfname(suffix='', prefix='tmp', directory=None, text=False):
     return fn
 
 
+@contextlib.contextmanager
```
dpark01: Like @tomkinsc I am also a fan of these two new functions.

tools/picard.py
```diff
@@ -101,19 +102,50 @@ class SamToFastqTool(PicardTools):
...
-                pass
+            for f in (outFastq1, outFastq2, outFastq0):
+                if f: util.file.touch_empty(f)
         else:
             picardOptions = picardOptions or []
+
             opts = [
                 'FASTQ=' + outFastq1, 'SECOND_END_FASTQ=' + outFastq2, 'INPUT=' + inBam, 'VALIDATION_STRINGENCY=SILENT'
             ]
+            if outFastq0:
```
dpark01: Curious to see more later on (in future PRs) the scenarios that outFastq0 would be useful, and also am wondering whether we should code for scenarios where outFastq1/2 are None as well.

util/file.py
```diff
@@ -473,3 +494,33 @@ def line_count(infname):
 def touch(fname, times=None):
     with open(fname, 'a'):
         os.utime(fname, times)
+
+def touch_empty(fname):
+    '''Make each file in fnames a zero-length file with the current timestamp.'''
+    with open(fname, 'w'):
```
notestaff: You're right about the name.  Reason I wanted a make-file-empty function, rather than just touch, is to ensure an empty output file even if the previous operation failed midway and left a partially-written file, or there was already a file with that name there.

util/file.py
notestaff: I'll add a default param for max file size to slurp; if you want to slurp a larger file you'll have to override the param explicitly.  I only use it when reading a file I know to be small, like a user-written config file; there it leads to clearer code.

tools/picard.py
notestaff: The SPAdes assembler can take some paired reads and some unpaired reads, so that was the immediate use case.   I can add the outFastq1/2==None case for outputting just the unpaired reads.

util/file.py
dpark01: would you be open to changing the function name to something like `make_empty`?

util/file.py
yesimon: This is similar to `tempfile.NamedTemporaryFile`. Although that provides file-like objects

util/file.py
```diff
@@ -94,6 +94,27 @@ def mkstempfname(suffix='', prefix='tmp', directory=None, text=False):
...
+    '''Create a tempfile name on context entry, delete the file (if it exists) on context exit.'''
+    fn = mkstempfname(*args, **kwargs)
+    try:
+        yield fn
+    finally:
+        if os.path.isfile(fn): os.unlink(fn)
+
+
+@contextlib.contextmanager
+def tempfnames(suffixes, *args, **kwargs):
```
yesimon: If we move to python 3, we can easily stack/nest contextmanagers and cleanup using `contextlib.ExitStack`

util/file.py
yesimon: Maybe `truncate()` ?

util/file.py
```diff
@@ -94,6 +94,27 @@ def mkstempfname(suffix='', prefix='tmp', directory=None, text=False):
...
+    '''Create a setof tempfile names on context entry, delete the files (if they exist) on context exit.'''
```
yesimon: s/setof/set of/

tools/picard.py
```diff
@@ -101,19 +102,50 @@ class SamToFastqTool(PicardTools):
...
+    @contextlib.contextmanager
+    def execute_tmp(self, inBam, sfx='', includeUnpaired=False, **kwargs):
+        '''Output reads from `inBam` to temp fastq files, and yield their filenames as a tuple.
+        If `includeUnpaired` is True, output unpaired reads to a third temp fastq file and yield it as a third
+        element of the tuple.
+        '''
+        if not includeUnpaired:
+            with util.file.tempfnames((sfx+'.1.fq', sfx+'.2.fq')) as (outFastq1, outFastq2):
+                self.execute(inBam, outFastq1, outFastq2, **kwargs)
+                yield outFastq1, outFastq2
```
yesimon: I think the files here get destroyed before the names are returned to the user. When you leave a function context the context managers within are exited as well. Would make a good unit test case.

util/file.py
yesimon: In fact there is an `os.truncate(path, size)`

util/file.py
notestaff: Interesting, didn't know about ExitStack.  Still, it might end up as 2 lines vs 1 here.  Point of these small utils is to reduce bookkeeping clutter so that core algorithmic code becomes clearer, and for that purpose one-liner is better than two-liner.

util/file.py
notestaff: "we're trying to follow pep8" -- ok.   I prefer using fewer lines of code so more of it fits on the screen at once.  I also find that more spaces in a line make it easier to read.  But ok, standards are standards.

tools/picard.py
notestaff: " the files here get destroyed before the names are returned to the user" -- are you sure?  The yield statement only completes when the user's context closes, at which point the user is done with the filenames.

yesimon: Oh yeah, this is a `yield` statement, not return. Carry on!

util/file.py
notestaff: os.truncate takes a file descriptor, rather than a filename, so it's still a two-liner to make an empty file.

PR #640, opened 2017-07-25
remove broken symlinks in runs; call CheckIlluminaDirectory

tomkinsc:
HiSeq4000 and HiSeq X runs write out a single “s.locs” file to Data/Intensities, rather than per-tile cluster location files within Data/Intensities/L\*/\*.locs. Related to this, CheckIlluminaDirectory with LINK_LOCS=‘true’ creates symlinks from where per-tile location files would be to the absolute path of the s.locs file. This can cause problems when moving a run to a new system, since the absolute paths will be incorrect. This addition checks to see if s.locs is present. If so, broken symlinks within the Data/Intensitites/L\* are removed, and LINK_LOCS is specified for the CheckIlluminaDirectory call that now happens before demultiplexing. This call ensures the run looks correct, and in the case of HiSeq4000/HiSeqX runs, also creates symlinks for per-tile location files using absolute paths on the current system. This is all something of a workaround for Broad-derived HiSeq4000/HiSeqX runs, which are delivered with brittle absolute symlinks. With this addition, runs from Broad walkup or elsewhere are more cloud-compatible. In connection with this, a PR has been opened in the Picard repository to request that LINK_LOCS can create relative symlinks rather than absolute:
https://github.com/broadinstitute/picard/issues/877

### DIFF
--- a/illumina.py
+++ b/illumina.py
@@ -117,10 +117,17 @@ def main_illumina_demux(args):
     # file is present
     if os.path.exists(os.path.join(illumina.get_intensities_dir(), "s.locs")):
         # recurse to remove broken links in directory
-        util.file.remove_broken_symlinks(illumina.get_intensities_dir())
+        log.info("This run has an 's.locs' file; checking for and removing broken per-tile symlinks...")
+        broken_links = util.file.find_broken_symlinks(illumina.get_intensities_dir())
+        if len(broken_links):
+            for lpath in broken_links:
+                log.info("Removing broken symlink: %s", lpath)
+                os.unlink(lpath)
+
         # call CheckIlluminaDirectory with LINK_LOCS=true
         link_locs="true"
 
+    log.info("Checking run directory with Picard...")
     tools.picard.CheckIlluminaDirectoryTool().execute(
         illumina.get_BCLdir(),
         args.lane,
--- a/illumina.py
+++ b/illumina.py
@@ -106,7 +106,7 @@ def main_illumina_demux(args):
         samples = illumina.get_SampleSheet(only_lane=args.lane)
 
 
-    link_locs="false"
+    link_locs=False
     # For HiSeq-4000/X runs, If Picard's CheckIlluminaDirectory is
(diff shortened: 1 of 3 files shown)


illumina.py
```diff
@@ -105,6 +105,29 @@ def main_illumina_demux(args):
...
+    # For HiSeq-4000/X runs, If Picard's CheckIlluminaDirectory is
+    # called with LINK_LOCS=true, symlinks with absolute paths
+    # may be created, pointing from tile-specific *.locs to the 
+    # single s.locs file in the Intensities directory.
+    # These links may break if the run directory is moved.
+    # We should begin by removing broken links, if present,
+    # and call CheckIlluminaDirectory ourselves if a 's.locs'
+    # file is present
+    if os.path.exists(os.path.join(illumina.get_intensities_dir(), "s.locs")):
+        # recurse to remove broken links in directory
```
dpark01: Perhaps add a `log.info` message here just so we can trace after the fact whether this is happening or not?

util/file.py
```diff
@@ -516,3 +516,40 @@ def slurp_file(fname, maxSizeMb=50):
...
+        # otherwise traverse the directory hierarchy
+        log.info("Checking for broken symlinks to remove in: %s", rootdir)
+        for rootpath, subfolders, files in os.walk(rootdir, followlinks=followlinks):
+            for filename in files:
+                fpath = os.path.join(rootpath, filename)
+                if is_broken_link(fpath):
+                    broken_links_to_remove.append(fpath.rstrip("/"))
+
+    for lpath in broken_links_to_remove:
+        log.info("Removing broken symlink: %s", lpath)
```
dpark01: Since this is in `util.file`, I wonder if it might be better if the function was silent (from a logging point of view) and it just returned the list `broken_links_to_remove` as the return value. The higher up caller can then quickly interpret the the result as a boolean (did it do anything) or count the values or `log.info` the returned values if the verbosity is desired.

illumina.py
tomkinsc: Done!

PR #648, opened 2017-08-16
enable kraken on osx

dpark01:


yesimon:
Sure - I might move the package back to `kraken-all` and resolve the discrepancy on bioconda (somebody attached `jellyfish==1` to regular kraken pkg even though it's not needed for runtime.

dpark01:
I think I wasn't able to access `kraken-all` from OSX, just `kraken`. But perhaps this is all moot if you are about to replace this with the `broad-viral` channel build.

PR #649, opened 2017-08-16
Speedup tests

yesimon:
Randomly subsampled to 25% as many reads.

Change thread counts for other tests from 4 to 1


test/__init__.py
```diff
@@ -5,13 +5,45 @@
...
 import os
+import re
 import unittest
 import hashlib
 
 # intra-project
 import util.file
 from tools.samtools import SamtoolsTool
 
+def available_cpu_count():
```
dpark01: Why not just use `util.misc.available_cpu_count()`?

intrahost.py
```diff
@@ -293,16 +295,17 @@ def parse_alleles_string(allelesStr):
...
-def get_mpileup_allele_counts(inBam, chrom, pos, inConsFasta):
+def get_mpileup_allele_counts(inBam, chrom, pos, inConsFasta, samtools=None):
     """ Return {allele : [forwardCount, reverseCount], ...}
         allele is:
             Iins for insertions where ins represents the inserted bases
             Dlen for deletions where len is the length of the deletion
             base itself for non-indels
             'i' or 'd', in which case report count for consensus.
     """
+    samtools = samtools or SamtoolsTool()
```
dpark01: My gut is that it's not so much the instantiation of the class that costs time, but rather the `install_and_get_path` invocation every time. I'm not sure if pre-creating this object and passing it around aggressively will necessarily result in the same time savings as perhaps just turning the Conda-only Tool classes into something that doesn't even bother checking/verifying proper installation before executing it.

tomkinsc: Yeah, skipping the check I'd probably the way to go. The conda calls we use to check the install are the slow part. To call `conda list <package>` can take multiple seconds, even with the extra params that help reduce the time. Conda is slow.

PR #653, opened 2017-08-17
Refactor kraken tests

yesimon:
Only test metagenomics.py entry point and snakemake rules
Use metagenomics.py kraken_build to build the database
Test subset_taxonomy with kraken_build
Picard SamToFastq can deal with empty input

Move subset taxonomy to unit taxonomy test


tools/picard.py
```diff
@@ -150,27 +150,22 @@ def execute(self, inBam, outFastq1, outFastq2, outFastq0=None,
         trim reads at the clipping position specified by the Illumina clipping attribute
         (which is defined by the class variable SamToFastqTool.illumina_clipping_attribute).'''
 
-        if tools.samtools.SamtoolsTool().isEmpty(inBam):
-            # Picard SamToFastq cannot deal with an empty input BAM file
```
tomkinsc: We no longer need to harden Picard `SamToFastq ` against empty input?

yesimon: It seems like picard now properly handles empty input in this case - outputting an empty output.

PR #654, opened 2017-08-17
Update diamond to 0.9.10 with taxon LCA

yesimon:
Send everything through pipes

New `bind_pipes` function to bind two subprocess.PIPE objects. Since data sent through the standard pipes is binary encoded, it is necessary to wrap with `io.TextIOWrapper` on any pipes treated like text.


metagenomics.py
```diff
@@ -716,62 +734,68 @@ def parser_krona(parser=argparse.ArgumentParser()):
...
-    diamond_tool.blastx(db, [tmp_fastq, tmp_fastq2], tmp_alignment, options={'--threads': numThreads})
-    diamond_tool.view(tmp_alignment, tmp_m8, options={'--threads': numThreads})
-
-    if outM8:
-        with open(tmp_m8, 'rb') as f_in:
-            with gzip.open(outM8, 'wb') as f_out:
-                shutil.copyfileobj(f_in, f_out)
+    taxonmap = join(taxDb, 'accession2taxid', 'prot.accession2taxid.gz')
+    taxonnodes = join(taxDb, 'nodes.dmp')
+    cmd = '{} blastx --db {db}'.format(diamond_tool.install_and_get_path(), db=db)
```
dpark01: Maybe I missed something in the diff, but this look like a:
```
cmd =
cmd =
cmd +=
cmd +=
```
And I'm wondering if the first line should be removed, or if we missed an exec, or missed a `+=`

metagenomics.py
```diff
@@ -716,62 +734,68 @@ def parser_krona(parser=argparse.ArgumentParser()):
...
 
-    tax_db = TaxonomyDb(tax_dir=taxDb, load_names=True, load_nodes=True, load_gis=True)
-    tmp_lca_tsv = util.file.mkstempfname('.tsv')
-    with open(tmp_m8) as m8, open(tmp_lca_tsv, 'w') as lca:
-        blast_lca(tax_db, m8, lca, paired=True, min_bit_score=50)
+    cmd = '{} blastx --outfmt 102 --sallseqid'.format(diamond_tool.install_and_get_path())
+    cmd += ' --threads {threads}'.format(threads=numThreads)
```
dpark01: Can we make this string append an `if threads is not None` scenario?

metagenomics.py
```diff
@@ -716,62 +734,68 @@ def parser_krona(parser=argparse.ArgumentParser()):
...
-            with gzip.open(outLca, 'wb') as f_out:
-                shutil.copyfileobj(f_in, f_out)
+    if outLca is not None:
+        cmd += ' | tee >(gzip > {out})'.format(out=outLca)
+
+    diamond_ps = subprocess.Popen(cmd, shell=True, stdin=subprocess.PIPE,
+                                  stdout=subprocess.PIPE, executable='/bin/bash')
+
+    def f(input_pipe, output_pipe):
+        output_pipe = codecs.getwriter('ascii')(output_pipe)
```
dpark01: Was this binary pipe thing just for Py3 or also for Py2?

metagenomics.py
```diff
@@ -716,62 +734,68 @@ def parser_krona(parser=argparse.ArgumentParser()):
...
 
+    s2fq.wait()
+    diamond_ps.wait()
+
 
 def parser_diamond(parser=argparse.ArgumentParser()):
     parser.add_argument('inBam', help='Input unaligned reads, BAM format.')
     parser.add_argument('db', help='Diamond database directory.')
     parser.add_argument('taxDb', help='Taxonomy database directory.')
     parser.add_argument('outReport', help='Output taxonomy report.')
```
dpark01: Perhaps for consistency with `parser_kraken`, should we rename `--outLca` to `--outReads`, and choose something consistent for both kraken and diamond in terms of which of required vs optional on the two output files?

util/file.py
```diff
@@ -626,3 +630,26 @@ def find_broken_symlinks(rootdir, followlinks=False):
...
+    num_n = num_n or 31
+
+    # In case input is a stdout pipe (always binary encoded)
+    input_fastq_f = codecs.getreader('ascii')(input_fastq_f)
+    for r1, r2 in util.misc.batch_iterator(SeqIO.parse(input_fastq_f, 'fastq'), 2):
+        if r1.id.endswith('/1'):
+            rid = r1.id[:-2]
+        else:
+            rid = r1.id
+        jseq = Seq(str(r1.seq) + 'A' * num_n + str(r2.seq))
```
dpark01: Shouldn't this be `'N' * num_n` instead of `'A' * num_n`?

util/file.py
```diff
@@ -626,3 +630,26 @@ def find_broken_symlinks(rootdir, followlinks=False):
...
+        labbrevs = None
+        if output_format == 'fastq':
+            # Illumina quality score of 2 indicates unreliable base
+            labbrevs = {
+                'phred_quality': r1.letter_annotations['phred_quality'] + [2] * num_n + r2.letter_annotations['phred_quality']
+            }
+        rec = SeqRecord(jseq, id=rid, description='', letter_annotations=labbrevs)
```
dpark01: If you're really into speed optimization, I've heard it said that creating and destroying a ton of BioPython SeqRecord objects for ephemeral use can be somewhat inefficient (if you're not translating or revcomping or doing other biological things to it) ... might consider just emitting a tuple of strings instead?

metagenomics.py
yesimon: That's a mistake

metagenomics.py
yesimon: I think it's both, but it requires `codecs.getwriter` to support both instead of just creating a `TextIOWrapper` which is available in py3.

metagenomics.py
yesimon: Done

metagenomics.py
yesimon: Done

metagenomics.py
```diff
@@ -602,11 +616,11 @@ def rank_code(rank):
         return "-"
 
 
-def taxa_hits_from_tsv(f, tax_id_column=3):
+def taxa_hits_from_tsv(f, taxid_column=2):
```
tomkinsc: The `hits` column in the *.tsv file has changed? Will that cause any issues elsewhere?

metagenomics.py
```diff
@@ -721,63 +739,69 @@ def parser_krona(parser=argparse.ArgumentParser()):
...
+        db=db,
+        taxonmap=taxonmap,
+        taxonnodes=taxonnodes)
 
-    if outLca:
-        with open(tmp_lca_tsv, 'rb') as f_in:
-            with gzip.open(outLca, 'wb') as f_out:
-                shutil.copyfileobj(f_in, f_out)
+    if outReads is not None:
+        cmd += ' | tee >(gzip > {out})'.format(out=outReads)
```
tomkinsc: Nice pipefitting. For our future selves, it may be worth adding a brief comment at the top describing the pipe setup in/out for this function.

metagenomics.py
yesimon: Nope, it was only used internally anyways

dpark01:
Hm, maybe I was looking at an older build where the lights were green...

PR #655, opened 2017-08-18
Add option to filter bwa output by alignment score

haydenm:
bwa mem apparently ignores its -T option (alignment score threshold)
on paired-end data. This adds an option to calculate an alignment
score for each query by summing the scores across the query's alignments,
and keep only the queries whose score is at least the value of the
specified threshold.

The added option is exposed through the reports.py align_and_plot_coverage
function.


yesimon:
I guess working with raw sam is better in the sense that it can support nonstandard sam files.

PR #656, opened 2017-08-18
added spades assembler

notestaff:
Added the spades assembler.   This PR just adds the spades command to assembly.py ; later PR will integrate spades into the pipeline.


assembly.py
```diff
@@ -339,6 +340,49 @@ def parser_assemble_trinity(parser=argparse.ArgumentParser()):
...
+    threads=1,
+):
+    ''' This step runs the SPAdes assembler.
+    '''
+
+    with tools.picard.SamToFastqTool().execute_tmp(inBam, includeUnpaired=True) as (reads_fwd, reads_bwd, reads_unpaired):
+        try:
+            tools.spades.SpadesTool().assemble(reads_fwd=reads_fwd, reads_bwd=reads_bwd, reads_unpaired=reads_unpaired,
+                                               contigs_out=outFasta, spades_opts=spades_opts)
+            if not os.path.isfile(outFasta) or os.path.getsize(outFasta) == 0:
```
yesimon: I would differentiate between spades program crash versus empty output, unless spades itself cannot differentiate. Similarly check `always_succeed` before raising exception

tools/spades.py
```diff
@@ -0,0 +1,74 @@
...
+    def execute(self, args, stdout=None):    # pylint: disable=W0221
+        tool_cmd = [self.install_and_get_path()] + args
+        log.debug(' '.join(tool_cmd))
+        if stdout:
+            stdout = open(stdout, 'w')
+        subprocess.check_call(tool_cmd, stdout=stdout)
+        if stdout:
+            stdout.close()
+
+    def assemble(self, reads_fwd, reads_bwd, contigs_out, reads_unpaired = None, contigs_trusted = None,
```
yesimon: No spaces around equals in function args definition

tools/spades.py
```diff
@@ -0,0 +1,74 @@
...
+        if os.path.getsize(reads_fwd) == 0:
+            util.file.make_empty(contigs_out)
+            return
+
+        with util.file.temp_dir('_spades') as spades_dir:
+            log.debug('spades_dir=' + spades_dir)
+            args = ['-1', reads_fwd, '-2', reads_bwd ]
+            if reads_unpaired: args += [ '-s', reads_unpaired ]
+            if contigs_trusted: args += [ '--trusted-contigs', contigs_trusted ]
+            if spades_opts: args += spades_opts.split()
```
yesimon: should be `shlex.split()`

tools/spades.py
```diff
@@ -0,0 +1,74 @@
...
+
+    def version(self):
+        return TOOL_VERSION
+
+    def execute(self, args, stdout=None):    # pylint: disable=W0221
+        tool_cmd = [self.install_and_get_path()] + args
+        log.debug(' '.join(tool_cmd))
+        if stdout:
+            stdout = open(stdout, 'w')
+        subprocess.check_call(tool_cmd, stdout=stdout)
```
yesimon: Is there a stderr log or general log file?

util/file.py
```diff
@@ -111,9 +111,29 @@ def tempfnames(suffixes, *args, **kwargs):
...
-        for fn in fns:
-            if os.path.isfile(fn):
-                os.unlink(fn)
+        if  not keep_tmp():
+            for fn in fns:
+                if os.path.isfile(fn):
+                    os.unlink(fn)
+
+@contextlib.contextmanager
+def temp_dir(*args, **kwargs):
```
yesimon: Maybe tmp_dir for consisntent naming

util/file.py
```diff
@@ -111,9 +111,29 @@ def tempfnames(suffixes, *args, **kwargs):
...
+    try:
+        name = tempfile.mkdtemp(*args, **kwargs)
+        yield name
+    finally:
+        if keep_tmp():
+            log.debug('keeping tempdir ' + name)
+        else:
+            shutil.rmtree(name)
+
+def keep_tmp():
```
yesimon: We should figure out a more consistent way to handle tmp and/or keeping tmp

assembly.py
notestaff: Differentiate how?  My understanding was that always_succeed means output empty contigs in either case, and "not always_succeed" means fail in either case.  (Should probably also remove contigs shorter than a given length before checking for emptiness?)

always_succeed is checked on the 'except' branch before raising an exception.  The raise statement on the try branch will move to the except branch where always_succeed is checked.

util/file.py
notestaff: There's a common command-line switch for keeping tmp, but it's not available when calling functions directly as opposed to calling CLI commands, hence the alternate way through the environment.  Also, when debugging a snakemake pipeline, to turn keeptmp on/off you have to change every command in every rule; with the environment, you just set/unset one variable.

tools/picard.py
```diff
@@ -152,6 +152,8 @@ def execute(self, inBam, outFastq1, outFastq2, outFastq0=None,
         trim reads at the clipping position specified by the Illumina clipping attribute
         (which is defined by the class variable SamToFastqTool.illumina_clipping_attribute).'''
 
+        assert os.path.isfile(inBam)
+
         if tools.samtools.SamtoolsTool().isEmpty(inBam):
             # Picard SamToFastq cannot deal with an empty input BAM file
```
tomkinsc: @yesimon says this empty input hardening is no longer necessary, so since it is removed upstream it can be here too.

notestaff: The assertion checks for non-existent input, not empty input.  I added it because without it the error manifested downstream and it took me some time to figure out the cause.  In general, doing sanity-checks earlier can't hurt, no?  Unless there are so many that they obscure the core logic.

yesimon: Do we handle non-existent input in a consistent way? If we do, shouldn't it be at the command level after parsing args, instead of at the tool library level?

notestaff: I added a check at the command level.

notestaff:
thanks for fast review of  #659; now this PR has only SPAdes-specific changes.

assembly.py
```diff
@@ -339,6 +340,55 @@ def parser_assemble_trinity(parser=argparse.ArgumentParser()):
...
+    '''
+
+    with tools.picard.SamToFastqTool().execute_tmp(inBam, includeUnpaired=True) as (reads_fwd, reads_bwd, reads_unpaired):
+        try:
+            tools.spades.SpadesTool().assemble(reads_fwd=reads_fwd, reads_bwd=reads_bwd, reads_unpaired=reads_unpaired,
+                                               contigs_untrusted=previously_assembled_contigs,
+                                               contigs_out=outFasta, spades_opts=spades_opts, mem_limit_gb=mem_limit_gb,
+                                               threads=threads)
+        except subprocess.CalledProcessError as e:
+            raise DenovoAssemblyError('SPAdes assembler failed: ' + str(e))
```
tomkinsc: Is the SPAdes error message helpful enough to identify if the failure is due to insufficient input size? If not, maybe the `DenovoAssemblyError ` should include some input metrics.

notestaff: This is failure due to internal assembler error, not due to input size.  My understanding was that even zero input size should not be an error?

tomkinsc: Ok, and yes generally we've been trying to write out empty output for empty input.

assembly.py
```diff
@@ -339,6 +340,55 @@ def parser_assemble_trinity(parser=argparse.ArgumentParser()):
...
+            several contigs representing different variants of that region.
+
+    Params:
+        mem_limit_gb - max memory to use
+        threads - number of threads to use
+        spades_opts - (advanced) custom command-line options to pass to the SPAdes assembler
+
+    '''
+
+    with tools.picard.SamToFastqTool().execute_tmp(inBam, includeUnpaired=True) as (reads_fwd, reads_bwd, reads_unpaired):
```
dpark01: Similar questions to other PR:
- if unspecified or set to None, can Spades do something reasonable on its own for thread count and memory limit?

Also, if we have a memory limit suggestion given to us already, perhaps we should be passing that to the JVMmemory optional param when running Picard?

tools/spades.py
```diff
@@ -0,0 +1,82 @@
...
+
+log = logging.getLogger(__name__)
+
+class SpadesTool(tools.Tool):
+    '''Tool wrapper for SPAdes tool (St. Petersburg Assembler)'''
+
+    def __init__(self, install_methods=None):
+        if install_methods is None:
+            install_methods = [tools.CondaPackage(TOOL_NAME, version=TOOL_VERSION, executable='spades.py',
+                                                  verifycmd='echo spades.py --test', env='spades_env')]
```
dpark01: Is spades given its own conda environment because of weird python dependencies?

notestaff: I recall seeing problems that were fixed by using a separate env, but not what the problems exactly were.  spades is partly in C++ so it may be about C++ library dependencies.  I can look again into moving it into the common env.

dpark01: Don't worry about it for this PR, let's get it merged in and deal with it later if it ends up being possible.

PR #659, opened 2017-08-23
added a more centralized way to preserve temp files for debugging

notestaff:
Added a way to preserve tempfiles for debugging by setting an environment variable.  (This was part of PR #656 is-spades-2, but orthogonal to that PR so factoring it out here.)

### DIFF
--- a/util/file.py
+++ b/util/file.py
@@ -82,7 +82,6 @@ def get_resources():
         resources = json.load(inf)
     return resources
 
-
 def mkstempfname(suffix='', prefix='tmp', directory=None, text=False):
     ''' There's no other one-liner way to securely ask for a temp file by
         filename only.  This calls mkstemp, which does what we want, except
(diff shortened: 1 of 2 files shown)


util/file.py
```diff
@@ -111,9 +110,29 @@ def tempfnames(suffixes, *args, **kwargs):
...
+            shutil.rmtree(name)
+
+def keep_tmp():
+    """Whether to preserve temporary directories and files (useful during debugging).
+    Return True if the environment variable VIRAL_NGS_TMP_DIRKEEP is set.
+    """
+    return 'VIRAL_NGS_TMP_DIRKEEP' in os.environ
 
 def set_tmp_dir(name):
     proposed_prefix = ['tmp']
```
tomkinsc: Not directly related to this PR, but should we make use of [`util.cmd.find_tmp_dir()`](https://github.com/broadinstitute/viral-ngs/blob/a3079b281a2ae3ebfd56095b4c0eae6775ad4874/util/cmd.py#L216) here to propose the tmp prefix?

yesimon:
What about `--tmp_dirKeep` ?

util/file.py
notestaff: here=within set_tmp_dir?

notestaff:
@yesimon --tmp_dirKeep is a command-line option, not available when calling functions directly as opposed to calling CLI commands, hence the alternate way through the environment. Also, when debugging a snakemake pipeline, to turn keeptmp on/off you have to change every command in every rule; with the environment, you just set/unset one variable.

util/file.py
tomkinsc: Nevermind, I see `find_tmp_dir` is used in `set_tmp_dir`. We could refactor to move `find_tmp_dir()` to`util.file`

PR #660, opened 2017-08-23
added a util to check that paths passed to a command can be read/writen as required

notestaff:
Added a util to check that paths passed to a command can be read/writen as required.  ( This was part of PR #656 is-spades-2, but orthogonal to that PR so factoring it out here.)

### DIFF
--- a/util/file.py
+++ b/util/file.py
@@ -82,6 +82,28 @@ def get_resources():
         resources = json.load(inf)
     return resources
 
+def check_paths(read=(), write=(), read_and_write=()):
+    '''Check that we can read and write the specified files, throw an exception if not.  Useful for checking
+    error conditions early in the execution of a command.  Each arg can be a filename or iterable of filenames.
+    '''
+    read, write, read_and_write = map(util.misc.make_seq, (read, write, read_and_write))
+    assert not (set(read) & set(write))
+    assert not (set(read) & set(read_and_write))
+    assert not (set(write) & set(read_and_write))
+    
+    for fname in read+read_and_write:
+        with open(fname):
+            pass
+
+    for fname in write+read_and_write:
+        if not os.path.exists(fname):
+            with open(fname, 'w'):
+                pass
+            os.unlink(fname)
+        else:
+            if not (os.path.isfile(fname) and os.access(fname, os.W_OK)):
+                raise PermissionError('Cannot write ' + fname)
+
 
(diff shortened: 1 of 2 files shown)


tomkinsc:
It would be more for a follow-on PR, but in addition to this function which is useful for our module-level API, we could make greater use of [`argparse.FileType()`](https://docs.python.org/3/library/argparse.html#filetype-objects) in our parsers where we pass in file paths.

notestaff:
We could extend argparse.FileType() to also accept URIs, s3:// addresses etc, downloading the file to a local cache and returning the path.  Then the CLI will work with remote files, not just the Snakemake pipeline.  For output files this would return a tempfile path and register a callback to upload it to the remote address if command succeeds.

tomkinsc:
That's a great idea, @notestaff! We could even release the `FileType()` extension (or however it is implemented) as a public module available to all who use argparse. A win for us and others, and it would help us decouple from Snakemake for remote execution.. Letting all Python argparse software use remote files would be wonderful in general. It could probably reuse some of the same machinery/dependencies as the Snakemake remote providers. Passing credentials or backend-specific params would require some API design consideration. Remote files almost want to be an argparse "group" with the actual file as a positional and then optional params for credentials, etc. Testing writability could be a challenge on some backends without actually writing out data.

util/file.py
```diff
@@ -82,6 +82,28 @@ def get_resources():
         resources = json.load(inf)
     return resources
 
+def check_paths(read=(), write=(), read_and_write=()):
+    '''Check that we can read and write the specified files, throw an exception if not.  Useful for checking
+    error conditions early in the execution of a command.  Each arg can be a filename or iterable of filenames.
+    '''
+    read, write, read_and_write = map(util.misc.make_seq, (read, write, read_and_write))
```
dpark01: This passes in Py3? I thought `map` returned an iterable in Py3 (and `util.misc.make_seq` would then be lazily executed later on) and the multiple reads of the iterator down below would fail?

notestaff: map returns iterable over (make_seq(read), make_seq(write), make_seq(read_and_write))
then tuple unpacking runs through that iterable once,, assigning read=make_seq(read),, write=make_seq(write) etc
make_seq returns a tuple

dpark01: Ah okay, I didn't read that carefully. Looks good to merge!

PR #661, opened 2017-08-24
adding Gap2Seq tool for filling gaps between contigs using reads.

notestaff:
adding Gap2Seq tool for filling gaps between contigs using reads.  This PR just adds the gapfill_gap2seq command to assembly.py ; later PR will link it into the pipeline.


assembly.py
```diff
@@ -338,6 +339,27 @@ def parser_assemble_trinity(parser=argparse.ArgumentParser()):
...
 __commands__.append(('assemble_trinity', parser_assemble_trinity))
 
+def gapfill_gap2seq(in_scaffold, inBam, out_scaffold, gap2seq_opts='', threads=1, mem_limit_gb=4):
+    ''' This step runs the Gap2Seq tool to close gaps between contigs in a scaffold.
+    '''
+    tools.gap2seq.Gap2SeqTool().gapfill(in_scaffold, inBam, out_scaffold, gap2seq_opts=gap2seq_opts, threads=threads,
+                                        mem_limit_gb=mem_limit_gb)
+
+def parser_gapfill_gap2seq(parser=argparse.ArgumentParser(description='Close gaps between contigs in a scaffold')):
+    parser.add_argument('in_scaffold', help='Scaffold with gaps (FASTA witht Ns)')
```
tomkinsc: Typo in the help string; `s/witht/with/g`

assembly.py
```diff
@@ -338,6 +339,27 @@ def parser_assemble_trinity(parser=argparse.ArgumentParser()):
...
+    parser.add_argument('inBam', help='Input unaligned reads, BAM format.')
+    parser.add_argument('out_scaffold', help='Output assembly.')
+    parser.add_argument('--gap2seq_opts', default='', help='(advanced) Extra command-line options to pass to Gap2Seq')
```
tomkinsc: We have a loose convention of having the argument names exposed via CLI in `--camelCase` with the argparse `dest='snake_case'` (A bit less for the user to type while still striving for PEP 8 within our Python code.)

test/unit/test_assembly.py
```diff
@@ -312,6 +312,18 @@ def test_multi_overlap(self):
...
+    '''Test gap-filling tool Gap2Seq'''
+
+    def test_gapfill(self):
+        from os.path import join
+        inDir = util.file.get_test_input_path()
+        in_scaffold = join(inDir, 'TestOrderAndOrient', 'expected.ebov.doublehit.fasta')
+        with util.file.tempfname(suffix='.filled.fasta') as filled:
+            assembly.gapfill_gap2seq(in_scaffold=in_scaffold,
+                                     inBam=join(inDir, 'G5012.3.testreads.bam'), out_scaffold=filled)
+            self.assertEqualContents(filled, join(inDir, 'TestGap2Seq', 'expected.ebov.doublehit.gapfill.fasta'))
```
tomkinsc: If the output is always the same, maybe we should check the hash rather than store the expected output verbatim. We have [`assert_md5_equal_to_line_in_file`](https://github.com/broadinstitute/viral-ngs/blob/master/test/__init__.py#L55) to help. Simon and I had some brief chatter about this on Slack in relation to his effort toward repo thinning.

notestaff: I remember, but it's a small file, and there's some benefit to being able to look at assertions and see _why_ they should hold.  E.g. here one can align reads to the filled-in scaffold and see that they do align.

tools/gap2seq.py
```diff
@@ -0,0 +1,110 @@
...
+                                '-verbose', '-nb-cores', str(threads), '-max-mem', str(mem_limit_gb)] + shlex.split(gap2seq_opts)
+                        
+                        with util.file.tmp_dir('_gap2seq_run_dir') as gap2seq_run_dir:
+                            with util.file.tmp_chdir(gap2seq_run_dir):
+                                self.execute(args=args)
+
+                        prev_scaffold = gap2seq_filled
+
+                        if not any('N'*min_gap_to_close in str(rec.seq) for rec in Bio.SeqIO.parse(gap2seq_filled, 'fasta')):
+                            log.info('no gaps left, quittting gap2seq early')
```
tomkinsc: Maybe `log.info()` should be `log.error()` (or `log.warning()`) here? Same with the time limit conditional below.

util/file.py
```diff
@@ -128,6 +128,16 @@ def tmp_dir(*args, **kwargs):
         else:
             shutil.rmtree(name)
 
+@contextlib.contextmanager
+def tmp_chdir(name):
```
tomkinsc: This isn't `tmp`-specific, right? (I get what you mean: temporarily change directories within the generator context, but it reads like it deals with temp files) Let's rename it to be more generic. How about `pushd()`, `pushd_and_popd()` or something along those lines? Let's also change the argument `name` to `path`. Perhaps it should `yield path` for convenience downstream.

tools/gap2seq.py
notestaff: This one is more of a log.happy(): all gaps filled before we've tried all parameter combinations, so we needn't try the remaining ones.  Time limit can be a warning.

test/unit/test_assembly.py
tomkinsc: I agree about it being a small file, but could we achieve the same thing by having the assertion raise with an appropriate message on failure? It would mean changing `assert_md5_equal_to_line_in_file` to have an optional argument (maybe called `failure_message`) that would be passed as the third argument of the [`assertEqual()` comparison it uses](https://github.com/broadinstitute/viral-ngs/blob/master/test/__init__.py#L74). 

Or do you mean it would be helpful to have the expected output in case the developer running `pytest` wants to retain the temp output file for manual inspection and comparison to the expected output?

tools/gap2seq.py
tomkinsc: Ohh, got it. I like the notion of `log.happy()` :)
The time limit one below should be a warning/error though, right?

test/unit/test_assembly.py
notestaff: The latter. Small files, besides taking little space, are amenable to manual inspection/comparison.

notestaff: I can gzip it to make it even smaller.

tools/gap2seq.py
notestaff: The whole gapfill stage is optional; if it does nothing, the pipeline still works.  So even an internal crash of gap2seq is not an error or a serious warning, more a log.fyi(); but I'll make it a warning.  Maybe we ahould gather warnings issued during a run and re-print them together at the end of output, like GATK does.

test/unit/test_assembly.py
tomkinsc: Got it, fair enough. Don't bother gzipping it; we'd lose nice/readable git changes.

tools/gap2seq.py
notestaff: Also, timeout doesn't mean it did nothing; it could have filled 3 of 4 gaps then ran out of time.  We don't know if it could have filled the last gap even with more time, so it's not a definite failure/error, more a suggestion to maybe tune your params.

tomkinsc: Re-printing warnings at the end would be great!

tomkinsc: The timeout log message should probably suggest to the user that he or she should tune the params?

notestaff: Right.  Sometimes we can tell that won't help, e.g. if the gaps that are left are too large.  There are other pipeline places where we could also suggest what to try tuning.

tools/gap2seq.py
```diff
@@ -0,0 +1,110 @@
...
+        kmer_sizes = sorted(util.misc.make_seq(kmer_sizes), reverse=True)
+        samtools = tools.samtools.SamtoolsTool()
+        with samtools.bam2fq_tmp(inBam) as (reads1, reads2):
+            done = False
+            prev_scaffold = in_scaffold
+            stop_time = time.time() + 60*time_limit_minutes
+            with util.file.tmp_dir('_gap2seq_dir') as gap2seq_dir:
+                for solid_kmer_threshold in solid_kmer_thresholds:
+                    for kmer_size in kmer_sizes:
+                        gap2seq_filled = join(gap2seq_dir, 'gap2seq-filled.s{}.k{}.fasta'.format(solid_kmer_threshold, kmer_size))
```
yesimon: `itertools.product` ?

tools/gap2seq.py
```diff
@@ -0,0 +1,110 @@
...
+                We try gapfilling for all combinations of values of solid_kmer_thresholds and kmer_sizes.
+            kmer_sizes: kmer sizes to use.  We try gapfilling for all combinations of values of solid_kmer_thresholds and kmer_sizes.
+            min_gap_to_close: stop gap-closing if all gaps are no longer than this many Ns
+            gap2seq_opts: extra command-line flags to pass to Gap2Seq
+            mem_limit_gb: max memory to use, in gigabytes
+            threads: number of threads to use
+            time_limit_minutes: stop trying to close more gaps after this many minutes (currently this is a soft/advisory limit)
+        
+        """
+        from os.path import abspath, join
```
yesimon: imports at top of file

test/unit/test_assembly.py
```diff
@@ -312,6 +312,18 @@ def test_multi_overlap(self):
             str(Bio.SeqIO.read(outFasta, 'fasta').seq),
             str(Bio.SeqIO.read(expected, 'fasta').seq))
 
+class TestGap2Seq(TestCaseWithTmp):
+    '''Test gap-filling tool Gap2Seq'''
+
+    def test_gapfill(self):
+        from os.path import join
```
yesimon: imports at top

tools/gap2seq.py
```diff
@@ -0,0 +1,110 @@
...
+            install_methods = [tools.CondaPackage(TOOL_NAME, version=TOOL_VERSION, executable='Gap2Seq.sh')]
+        tools.Tool.__init__(self, install_methods=install_methods)
+
+    def version(self):
+        return TOOL_VERSION
+
+    def execute(self, args):    # pylint: disable=W0221
+        tool_cmd = [self.install_and_get_path()] + args
+        log.debug('running gap2seq: ' + ' '.join(tool_cmd))
+        subprocess.check_call(tool_cmd)
```
yesimon: If there's an internal crash of gap2seq this would fail

tools/gap2seq.py
```diff
@@ -0,0 +1,110 @@
...
+                        # gap2seq (or rather the gatb library it uses) currently has a bug where tempfiles are left in the
+                        # current working directory.  So we run it in its own temp dir, but then must give absolute paths
+                        # for all files.
+
+                        args = ['-scaffolds', abspath(prev_scaffold), '-filled', abspath(gap2seq_filled),
+                                '-reads', ','.join((abspath(reads1), abspath(reads2))), '-solid', str(solid_kmer_threshold), 
+                                '-k', str(kmer_size), '-all-upper',
+                                '-verbose', '-nb-cores', str(threads), '-max-mem', str(mem_limit_gb)] + shlex.split(gap2seq_opts)
+                        
+                        with util.file.tmp_dir('_gap2seq_run_dir') as gap2seq_run_dir:
```
yesimon: agreed there's so many levels of nesting/indentation. this might be a use case for `contextlib.ExitStack` or condensing the ctx managers

tools/gap2seq.py
notestaff: Right, I was thinking of masking crashes since it's a non-critical path in the pipeline; on the other hand, if crashes are very rare then getting alerted to them might not be bad.  I haven't actually seen it crash.

tools/gap2seq.py
notestaff: Importing individual functions at top of file pollutes the namespace of the overall module, so normally you just import os.path there.  But then the code gets littered with os.path.join's etc, obscuring the core logic.  Importing individual functions locally is reasonable middle ground.

notestaff: P.S. os.path is still imported at the top so module dependencies are not obscured.

yesimon: In general all imports should be at the top for pep8 as well as tooling. If you really want to abbreviate the code I would suggest `join = os.path.join` at the beginning of the function instead.

tomkinsc: +1 To @yesimon's feedback

util/file.py
```diff
@@ -128,6 +128,16 @@ def tmp_dir(*args, **kwargs):
         else:
             shutil.rmtree(name)
 
+@contextlib.contextmanager
```
dpark01: ooh, nice add!

assembly.py
```diff
@@ -338,6 +339,27 @@ def parser_assemble_trinity(parser=argparse.ArgumentParser()):
...
+    '''
+    tools.gap2seq.Gap2SeqTool().gapfill(in_scaffold, in_bam, out_scaffold, gap2seq_opts=gap2seq_opts, threads=threads,
+                                        mem_limit_gb=mem_limit_gb)
+
+def parser_gapfill_gap2seq(parser=argparse.ArgumentParser(description='Close gaps between contigs in a scaffold')):
+    parser.add_argument('inScaffold', help='Scaffold with gaps (FASTA with Ns)')
+    parser.add_argument('inBam', help='Input unaligned reads, BAM format.')
+    parser.add_argument('outScaffold', help='Output assembly.')
+    parser.add_argument('--gap2seqOpts', dest='gap2seq_opts', default='', help='(advanced) Extra command-line options to pass to Gap2Seq')
+    parser.add_argument('--threads', default=1, type=int, help='Number of threads (default: %(default)s)')
```
dpark01: - I could see it being useful to expose the `time_limit_minutes` optional parameter. But maybe to reduce confusion (since "limit" implies a hard limit) you could call it something like "soft_time_limit" or similar.
- If we pass `None` to `--threads` or `--memLimitGb`, can Gap2Seq do something smart like pick a limit on its own? Or is it always mandatory for the user to specify these numbers?

notestaff:
"Are you currently preferring this over Sealer?" -- in my tests yes, and I've seen another project make the same switch.  I still want to try modifying Sealer to use a reference-set-based approach, for cases where de novo gap-filling fails.

assembly.py
notestaff: For memory Gap2Seq's default is 20GB, which seems high, so I put an argparse default of 4GB.
For threads, passing 0 means use all available cores.

PR #662, opened 2017-08-24
assert_md5_equal_to_line_in_file now accepts msg

tomkinsc:
The test assertion `test.assert_md5_equal_to_line_in_file()` now accepts `failure_message`

### DIFF
--- a/test/__init__.py
+++ b/test/__init__.py
@@ -52,7 +52,7 @@ def assert_equal_bam_reads(testCase, bam_filename1, bam_filename2):
             if os.path.exists(fname):
                 os.unlink(fname)
 
-def assert_md5_equal_to_line_in_file(testCase, filename, checksum_file):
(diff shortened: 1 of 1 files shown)


test/__init__.py
```diff
@@ -52,7 +52,7 @@ def assert_equal_bam_reads(testCase, bam_filename1, bam_filename2):
             if os.path.exists(fname):
                 os.unlink(fname)
 
-def assert_md5_equal_to_line_in_file(testCase, filename, checksum_file):
+def assert_md5_equal_to_line_in_file(testCase, filename, checksum_file, failure_message=None):
```
yesimon: Why not `msg` for consistency?

tomkinsc: I always lean toward being more explicit, but consistency here sounds good. I'll change it.

PR #664, opened 2017-08-25
simplified config code, fixed more corner cases

notestaff:
Rewrote config-include code without flattening/unflattening of dicts.  This makes it simpler, more consistent with snakemake's internal handling of config overriding, and fixes corner cases like a scalar in one config overridden by or overriding a dict in another config.

### DIFF
--- a/test/unit/test_util_misc.py
+++ b/test/unit/test_util_misc.py
@@ -180,38 +180,6 @@ class TestFeatureSorter(unittest.TestCase):
         )
 
 
-class TestDictFlatten(unittest.TestCase):
-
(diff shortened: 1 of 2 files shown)


yesimon:
We should have some tests for `_update_config` because yaml config merging is very complicated with lots of edge cases

notestaff:
It's tested as part of 
https://github.com/broadinstitute/viral-ngs/blob/master/test/unit/test_util_misc.py#L215

notestaff:
also, _update_config is copied from snakemake's code, so it's tested there.

PR #668, opened 2017-08-29
Add cd-hit-dup as alternative to mvicuna rmdup

yesimon:
Set default mismatch threshold to 4

`cd-hit-dup` is currently bugged and basically only recognizes perfect matches


test/unit/test_read_utils.py
```diff
@@ -193,6 +193,30 @@ def test_mvicuna_empty_input(self):
         )
         self.assertEqual(samtools.count(output_bam), 0)
 
+    def test_cdhit_canned_input(self):
+        samtools = tools.samtools.SamtoolsTool()
+
+        input_bam = os.path.join(util.file.get_test_input_path(self), 'input.bam')
+        expected_bam = os.path.join(util.file.get_test_input_path(self), 'expected.bam')
```
dpark01: Wow, cd-hit produces the exact same output bam as mvicuna? That's creepy... maybe our canned input was simplistic though.

yesimon: Oh this expected bam is not actually used.

yesimon:
Yeah I opened a PR here: https://github.com/weizhongli/cdhit/pull/52

It's a bug where it doesn't handle mismatches even though it's supposed to

test/unit/test_read_utils.py
dpark01: Oh I see it now.

dpark01:
Ok how about this -- let's merge this code in so we have it, and we can wait to change the default over until the bugfix is in (and perhaps until we do a little memory & runtime comparisons just to be sure).

PR #669, opened 2017-09-18
gapfill: added gapfill step to pipeline.

notestaff:
gapfill: added gapfill step to pipeline.


pipes/config.yaml
```diff
@@ -192,6 +192,9 @@ vphaser_max_bins: 10
 # allele freq > 0.005. If false, no filtering is performed.
 vcf_merge_naive_filter: false
 
+# Random seed, for non-deterministic steps.  0 means use current time.
+random_seed: 23923937
```
dpark01: Hm... did you assign a deterministic number here by default in order to make testing more deterministic? I think this file is often used as a template for user config files, so maybe we should make a config-tests.yaml that gets included only when running tests.

pipes/rules/assembly.rules
```diff
@@ -103,7 +105,8 @@ rule orient_and_impute:
                 ono_extra.extend(["--min_pct_contig_aligned", str(config["scaffold_min_pct_contig_aligned"])])
             ono_extra = " ".join(ono_extra)
             shell("{config[bin_dir]}/assembly.py order_and_orient {input[0]} {params.refGenome} {params.scaffolded_fasta} {ono_extra} --outAlternateContigs {params.alternate_fasta}")
-            shell("{config[bin_dir]}/assembly.py impute_from_reference {params.scaffolded_fasta} {params.refGenome} {output} --newName {params.renamed_prefix}{wildcards.sample} --replaceLength {params.replace_length} --minLengthFraction {params.length} --minUnambig {params.min_unambig} --index")
+            shell("{config[bin_dir]}/assembly.py gapfill_gap2seq {params.scaffolded_fasta} {input.cleaned_reads} {params.gapfilled_fasta} --memLimitGb {resources[mem]} --maskErrors --randomSeed {config[random_seed]}")
```
dpark01: can we do the idiom here where we make a snakemake param for this step called `random_seed=config.get('random_seed', 0)` or something like that, and then use the param in the `shell` invocation?

dpark01: I wonder if there's any value in also gapfilling the `alternate_fasta` just for fun... or does gap2seq require the contigs to be scaffolded first (ie, it only fills gaps, it doesn't extend contigs)?

assembly.py
```diff
@@ -340,22 +340,34 @@ def parser_assemble_trinity(parser=argparse.ArgumentParser()):
...
+                    random_seed=0, gap2seq_opts='', mask_errors=False):
     ''' This step runs the Gap2Seq tool to close gaps between contigs in a scaffold.
     '''
-    tools.gap2seq.Gap2SeqTool().gapfill(in_scaffold, in_bam, out_scaffold, gap2seq_opts=gap2seq_opts, threads=threads,
-                                        mem_limit_gb=mem_limit_gb)
+    try:
+        tools.gap2seq.Gap2SeqTool().gapfill(in_scaffold, in_bam, out_scaffold, gap2seq_opts=gap2seq_opts, threads=threads,
+                                            mem_limit_gb=mem_limit_gb, time_soft_limit_minutes=time_soft_limit_minutes, 
+                                            random_seed=random_seed)
+    except Exception as e:
```
dpark01: Excellent.. from a user point of view, this seems like the most sensible failure mode, almost all the time. Out of curiosity, under what conditions does this tool throw an error?

test/unit/test_assembly.py
```diff
@@ -353,7 +353,7 @@ def test_gapfill(self):
         in_scaffold = join(inDir, 'TestOrderAndOrient', 'expected.ebov.doublehit.fasta')
         with util.file.tempfname(suffix='.filled.fasta') as filled:
             assembly.gapfill_gap2seq(in_scaffold=in_scaffold,
-                                     in_bam=join(inDir, 'G5012.3.testreads.bam'), out_scaffold=filled)
+                                     in_bam=join(inDir, 'G5012.3.testreads.bam'), out_scaffold=filled, threads=1, random_seed=23923937)
```
dpark01: Can you also add explicit unit tests that test:
 - empty fasta input (no contigs)
 - empty bam input (no reads, just headers)

Also, do we know that gap2seq works with both single and paired reads?

pipes/config.yaml
notestaff: I just thought, users want determinism.  But maybe not, since many tools use time as seed by default.  Btw, mummer also makes random choices, with no way to set the seed.  For testing, LD_PRELOAD of srand() can be used to derandomize most programs.

pipes/rules/assembly.rules
notestaff: Setting random_seed to 0 in pipes/config.yaml would do that, no?

assembly.py
notestaff: I haven't seen it error yet, after implementing workarounds for some of its quirks.

test/unit/test_assembly.py
notestaff: I think this version doesn't use read pairing information, but the newsest one does (to localize reads to gaps).  We could use set of references to better localize reads to gaps.

pipes/config.yaml
dpark01: There's a ton of tools that make random choices, and many don't offer a way to fix the seed.. early versions of BWA were definitely like this (maybe even current ones). We use the `Random` option on calls to Novoalign. Let's default to zero?

pipes/rules/assembly.rules
dpark01: The difference is that the current implementation fails on someone's older / pre-existing config.yaml until they hand merge in changes from the latest one.. it'd be nice to fallback to a sensible default if not specified.

dpark01: Also there are some rare scenarios where I've found it useful to specify param values on the command line when I invoked snakemake (which then overrides the config).. but I forget what situations I found that useful in... but that would require exposing it as a `param:` in the rule.

notestaff: you can change config[] values on the snakemake command line.  and with the config file inclusion PR, a legacy config file can omit new param values that are in the updated pipes/config.yaml, which you would get together with the updated pipes/rules that use the new param.

notestaff: "it only fills gaps, it doesn't extend contigs" - yes.  (IVA has a mode to extend contigs, but seems to need higher coverage than ours.)

It might make sense to also gapfill after filling from reference and refining.

dpark01:
Hi @notestaff , this may have gotten lost in the per-diff comments earlier, but I'll post a few bigger picture questions here below.

dpark01:
Hi @notestaff , this may have gotten lost in the per-diff comments earlier, but I'll post a few bigger picture questions here below. I'm curious to hear more info about how this has performed in your testing scenarios on various data sets. If this is easier to discuss in person, we can connect tomorrow. On one hand, it's a pretty conservative change and is unlikely to make anything worse, but on the other, since it is a change to the primary assembly workflow that is used for everything, I'd like to have a sense of how much we've derisked the various use cases and scenarios.

Questions:
- How often did it make a difference (ie output is different from input)?
- How much runtime does it typically add?
- Have you ever noticed scenarios where it makes things worse than our current process (weird indels / gappy effects)? Have you tested on various Lassa data sets?

notestaff:
@dpark01 It can add 1-2 minutes of runtime; the runtime can be capped, as it's a series of separate invocations of the tool, first trying to close gaps using higher-coverage kmers, then where that fails, trying with lower-coverage ones.   The soft time limit can be made a hard time limit by using the timeout argument of subprocess.checked_call().

I haven't seen cases where it makes things worse.  But I'm still gathering/systematizing the benchmarks and will post the compiled results.  If there are specific datasets that would be good to test on, can you point me?

notestaff:
There is one way in which the current gapfill implementation can make things worse: to fill a gap, it uses all reads, not just those localized to the gap.  By contrast, filling in from a reference may fill in the wrong sequence, but at least the sequence comes from the right part of the genome.  Next version of Gap2Seq addresses that by using reads whose mates map to edges of the gap.  Those reads can be too few, but we can also include read pairs where at least one mate maps to the gap on one of the references.  So maybe hold off on merging until that's working.

dpark01:
@notestaff assuming you're still online at this point, can you merge in the most recent origin/master, resolve conflicts, and prep this to merge in if you think it's ready? (and if you don't think it's ready, let's just make it non-default behavior, but leave the code in there so it can be merged in before you go)

dpark01:
Should we close this in favor of #696?

PR #670, opened 2017-09-19
dummy checkin to check current master

notestaff:


dpark01:
@tomkinsc will work on this in #671

PR #671, opened 2017-09-21
WIP travis py27 testing repair

tomkinsc:
commits in attempt to resolve travis failures under py27

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -6,8 +6,8 @@ matrix:
     include:
         - os: linux
           python: 2.7
-        - os: linux
-          python: 3.4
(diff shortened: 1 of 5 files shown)


tomkinsc:
The conda cache thing was broken, but the real issue was seemingly that we were pinning to an older conda that was incompatible with the version of `wheel` we needed for other dependencies. And `pytest-xdist` seemingly doesn't have a build for py34. So there were several issues.

dpark01:
Ok well, I don’t know if we have a Picard “make it so” emoji, but let’s do this..

Issue #674, opened 2017-09-25
tolerate bad line endings in samplesheets


dpark01:
Although most of our csv-reading samplesheet code tolerates MS Excel-generated files (with Mac OS9 line endings), our txt-reading samplesheet code does not. In particular, `util.file.read_tabfile_dict` (and the related `read_tabfile`) are opening files as `'rt'` instead of `'rU'`. This is one of the dumber error modes that frequently happens that Redmond, WA imposes on us, but we must deal with.

tomkinsc:
Do we have example txt sample sheets to test against beyond [the one](https://github.com/broadinstitute/viral-ngs/blob/master/test/input/TestSampleSheet/SampleSheet-custom-1.txt) in the unit test input folder? (If so, we can attach them to this issue thread.)

dpark01:
While we could come up with more, I don't think they'd necessarily explore more corner cases of the code, they'd basically have the same structure.

We could create a file with `\r` and `\r\n` endings, but I'm a little worried that automagic stuff in git might convert them all... we could create a unit test that takes `SampleSheet-custom-1.txt` and rewrites it with a mix of OS9, DOS, and Unix line endings to use as new test inputs. We could do that to the csv files as well.

tomkinsc:
Sounds like a plan. I'll add a helper util somewhere.

tomkinsc:
In #675.

PR #675, opened 2017-09-25
improve newline handling

tomkinsc:
To address #674 

Using the mode ‘U’ to tell Python to open text files with universal
newline support is deprecated in py3, and may be removed in the future.
This commit alters open_or_gzopen to set newline=None (universal
newline) under py3 when ‘U’ is given in the mode, and changes a few
open() calls to use open_or_gzopen(). Add a few test cases for
tab-delimited sample sheets with various line endings. Also replace the
deprecated “py.test” with “pytest”.


util/file.py
```diff
@@ -224,18 +225,25 @@ def mkdir_p(dirpath):
...
 def read_tabfile_dict(inFile):
     ''' Read a tab text file (possibly gzipped) and return contents as an
         iterator of dicts.
     '''
-    with open_or_gzopen(inFile, 'rt') as inf:
+    with open_or_gzopen(inFile, 'rU') as inf:
         header = None
         for line in inf:
-            row = [item.strip() for item in line.rstrip('\n').split('\t')]
+            row = [item.strip() for item in line.rstrip('\n').rstrip('\r').split('\t')]
```
dpark01: this could just be `line.rstrip('\n\r').split('\t')` to reduce function calls (`rstrip` considers its argument to be an unordered list of characters to strip). also below in `read_tabfile`

util/file.py
```diff
@@ -224,18 +225,25 @@ def mkdir_p(dirpath):
             raise
 
 
-def open_or_gzopen(fname, *opts):
-    return fname.endswith('.gz') and gzip.open(fname, *opts) or open(fname, *opts)
+def open_or_gzopen(fname, *opts, **kwopts):
+    # 'U' mode is deprecated in py3 and may be unsupported in future versions, 
+    # so use newline=None when 'U' is specified
+    for opt in opts:
```
dpark01: you could precede the whole `for` block with `if sys.version_info[0] == 3:` so that the whole block gets skipped under py2 and sys.version_info only gets checked once under py3

test/input/TestSampleSheet/SampleSheet-custom-1_macos9-endings.txt
```diff
@@ -0,0 +1 @@
...
NIHCC_19	AAGAGGCA	CTCTCTAT	AEL2
NIHCC_20	GTAGAGGA	CTCTCTAT	AEL2
NIHCC_21	CGTACTAG	TATCCTCT	AEL2
NIHCC_22	AGGCAGAA	TATCCTCT	AEL2
NIHCC_23	TCCTGAGC	TATCCTCT	AEL2
NIHCC_24	TAGGCATG	TATCCTCT	AEL2
NIHCC_25	GCTACGCT	TATCCTCT	AEL2
NIHCC_26	CGAGGCTG	TATCCTCT	AEL2
NIHCC_27	AAGAGGCA	TATCCTCT	AEL2
NIHCC_28	GTAGAGGA	TATCCTCT	AEL2
```
dpark01: I'm always slightly worried about git auto-converting `.txt` files, but it does look like this came through properly.

Issue #686, opened 2017-10-10
make lastal databases ephemeral


dpark01:
Currently, in the snakemake-based workflow (including staging down database files from the cloud), we are storing all of the `lastal` generated database files (bck, des, prj, sds, ssp, suf, tis) in permanent storage and pulling them all down. Even when not pulled from the cloud, we maintain the full set of files in persistent network storage in the `config["lastal_ref_db_dir"]`.

From a user interaction point of view, I'm inclined to make this behave more like the DNAnexus UI, where the user provides a `fasta` file, and all lastal database files are generated on the fly, for each execution, and are ephemeral / never persisted anywhere. A few reasons:

 - it could be one, easy to understand file, instead of a whole bunch of opaque ones
 - it turns out that creating the lastal db files from the fasta is actually quite quick (seems to be less than one minute for typical LASV inputs) so there isn't much of an efficiency reason to save them
 - it seems the binary index formats are changing between each version of the `lastal` tool itself, leading to more things that could go wrong should we choose to update the tool on someone

I propose having the snakemake rules take a single fasta file as input for the lastalDb, modify the `filter_lastal_bam` command to do the same, and have it build and store in an ephemeral tmp_dir. Alternatively keep the CLI the same, but have the snakemake rule do what the DNAnexus applet is doing (invoke `lastal_build_db` and `filter_lastal_bam` in succession, while making sure the db doesn't go to network storage).

Thoughts?

tomkinsc:
We have a snakemake rule, [`build_lastal_db`](https://github.com/broadinstitute/viral-ngs/blob/master/pipes/rules/ncbi.rules#L49), to build a lastal db from a single fasta input file. The fasta file is downloaded on the fly based on the accessions in the file `config["accessions_file_for_lastal_db_build"]`. The only thing stopping the database from being ephemeral is wrapping the db files with Snakemake's [`temp() flag`](https://snakemake.readthedocs.io/en/stable/snakefiles/rules.html#protected-and-temporary-files), and removal of `objectify_remote()` from the [`filter_to_taxon`](https://github.com/broadinstitute/viral-ngs/blob/master/pipes/rules/hs_deplete.rules#L58) rule (since if the files exist, even remotely, they will be downloaded/used rather than built by the upstream `build_lastal_db` rule).

notestaff:
Maybe just move it from root project dir to tmp/lastal_db?

dpark01:
[...]

Issue #685, opened 2017-10-06
parallelize lastal_chunked_fastq


dpark01:
From watching DNAnexus logs, it seems that `lastal_chunked_fastq` only utilizes a single core. In most cases, it's not that slow (because it's given a reduced input to begin with), but in cases with a lot of input, it can take a very long time. There's no real reason it can't go 8x faster just by threading out the chunks that it is already making.

Alternatively, most of the time appears to be spent processing the outputs of lastal and producing a subsetted fastq file for output. Lastal itself is multithreaded--it's the downstream processing that is slow. Since we never actually care about most of that fastq file (we only care about the read IDs), there could be a significant speed up just by parsing the MAF output directly and extracting read IDs and emitting those only.

notestaff:
"we never actually care about most of that fastq file (we only care about the read IDs)" -- I was thinking of using the mapping locations to localize reads to gaps for gapfilling.

dpark01:
the current `lastal_chunked_fastq` throws away mapping locations. This is just for filtering cleaned reads to taxon-specific reads, and since our CLI is exposing BAM inputs and outputs as the API, all that's happening is this conversion of MAF -> fastq -> list of read IDs -> Picard FilterSamReads -> BAM, which means we can skip the MAF -> fastq step, which is where most of the runtime seems to be spent according to the logs (and it's also single threaded, so we're wasting most of the instance we're on).

notestaff:
right, I just meant, could save MAF output under tmp/ to then use it to localize reads to gaps.

tomkinsc:
Access to the input SeqIO object is sequential, so we’ll have to write out the chunks and then kick them over to a process pool. If I remember right we started chunking to reduce RAM usage, so it’s possible that running parallel instances of lastal could be an issue for memory, but we can certainly try it.

dpark01:
I have a suspicion that lastal’s memory usage is not the issue (blast memory usage is independent of query input size), but that it’s the mafsort step that’s the memory issue. That makes me lean again towards trashing the mafsort/convent to fastq code that’s the slow step and just extracting read ids for Picard FilterSamReads.

Interestingly, since you’re using Bio.SeqIO to read the input fastq and write them out in chunks, why are we even converting the input bam to fastq anymore? SeqIO could just read the input bam directly...

dpark01:
[...]

PR #687, opened 2017-10-11
lastal improvements

dpark01:
Addresses #685 and some portion of #686.


read_utils.py
```diff
@@ -92,16 +94,6 @@ def fastq_to_fasta(inFastq, outFasta):
     return 0
 
 
-def parser_fastq_to_fasta(parser=argparse.ArgumentParser()):
```
tomkinsc: I doubt anyone is depending on this, but any reason not to leave it exposed by the CLI?

taxon_filter.py
```diff
@@ -156,115 +156,47 @@ def bmtagger_wrapper(inBam, db, outBam, threads, JVMmemory=None):
...
+    '''
 
-    lastal_path = tools.last.Lastal().install_and_get_path()
-    mafsort_path = tools.last.MafSort().install_and_get_path()
-    mafconvert_path = tools.last.MafConvert().install_and_get_path()
-    no_blast_like_hits_path = os.path.join(util.file.get_scripts_path(), 'noBlastLikeHits.py')
+    with util.file.tmp_dir('-lastal_db') as tmp_db_dir:
+        # auto build db if needed
+        if not all(os.path.exists(db + x)
+            for x in ('.bck', '.des', '.prj', '.sds', '.ssp', '.suf', '.tis')):
```
tomkinsc: Maybe the list of DB extensions should be part of `tools.last`?

tools/last.py
```diff
@@ -88,28 +120,17 @@ def execute(self, inputFasta, outputDirectory, outputFilePrefix):    # pylint: d
...
         # it to the desired output location
-        os.chdir(os.path.realpath(outputDirectory))
-
-        # execute the lastdb command
-        _log.debug(" ".join(tool_cmd))
-        subprocess.check_call(tool_cmd)
-
-        # restore cwd
-        os.chdir(cwd_before_lastdb)
+        with util.file.pushd_popd(os.path.realpath(outputDirectory)):
```
tomkinsc: This such a convenient context manager!

taxon_filter.py
dpark01: done!

read_utils.py
dpark01: This old bit of code seemed to serve one internal purpose, but wasn't really providing any utility for user from a higher level point of view. I don't think we need to provide a comprehensive set of general utils from biopython / pysam -- just common operations for our own use. I have a branch going that I intend to PR later on that actually eliminates this function altogether.

tools/last.py
dpark01: 👍  @notestaff

pipes/rules/hs_deplete.rules
```diff
@@ -55,18 +55,17 @@ rule filter_to_taxon:
     '''
     input:  
         input_bam = config["data_dir"]+'/'+config["subdirs"]["depletion"]+'/{sample}.cleaned.bam',
-        lastalDb  = objectify_remote(expand('{lastal_ref_db_dir}/lastal.{ext}', lastal_ref_db_dir=config["lastal_ref_db_dir"], ext=['bck','des','prj','sds','ssp','suf','tis'])),
+        lastalDb  = objectify_remote(expand('{lastal_ref_db_dir}/lastal.fasta', lastal_ref_db_dir=config["lastal_ref_db_dir"])),
```
dpark01: @tomkinsc can you quickly take a peek and see if these changes I made to pipes/rules/hs_deplete.rules look okay? I also deleted the `build_lastal_db` rule from ncbi.rules (but not the `download_lastal_sources`)--the snakemake rules should be expecting only a `lastal.fasta` file as input/output. `taxon_filter.py filter_lastal_bam` accepts a naked fasta file as input for its db now (and that functionality is now unit tested).

tomkinsc: They look ok! The input variable could be renamed from `lastalDb` to `lastal_db_fasta` or similar to reflect that it is not actually the database. Do we still want to wrap the fasta file with `objectify_remote()` if the assumption is that the DB will usually be built quickly on the fly? Can't hurt to keep remote support I guess.

It could make to change [`lastal_ref_db_dir`](https://github.com/broadinstitute/viral-ngs/blob/master/pipes/config.yaml#L258) in `config.yml` to `lastal_db_fasta`,  and point to the source fasta rather than a directory containing the DB, with the assumption that the DB files would be alongside the fasta in the same directory.

read_utils.py
tomkinsc: Got it. Always nice to remove stale bits.

tools/last.py
```diff
@@ -35,49 +37,80 @@ class Lastal(LastTools):
...
     subtool_name = 'lastdb'
     subtool_name_on_broad = 'lastdb'
 
-    def build_database(self, fasta_files, database_prefix_path): # pylint: disable=W0221
-        output_file_prefix = os.path.basename(database_prefix_path)
-        output_directory = os.path.dirname(database_prefix_path)
-
+    def is_indexed(self, db_prefix):
+        return all(os.path.exists(db_prefix + x)
+            for x in ('.bck', '.des', '.prj', '.sds', '.ssp', '.suf', '.tis'))
```
tomkinsc: This makes more sense here than in `taxon_filter.py`, though it could be handy to have a class property exposing the list of extensions (not that any of the other tools do this yet, but it would be a start).

pipes/rules/hs_deplete.rules
dpark01: hm.. that's a good point on `objectify_remote`. how about this, since it looks functional, I'll go ahead and merge this in so we can have a functioning `master` branch again, and then I'll let you improve that rule at a later time after we get a few other necessary bits in.

tomkinsc: Sounds good!

PR #699, opened 2017-10-19
a simple form of reference selection

notestaff:
Just runs order_and_orient for several references (in parallel) then picks the scaffold with the most non-N bases.


notestaff:
Currently this takes references from refsel_db/refsel.fasta (which can be generated from accessions_file_for_refsel_db_build in the same way the lastal_db is generated from accessions_file_for_lastal_db_build .  If accessions_file_for_refsel_db_build is not specified, accessions_for_ref_genome_build is used (same as for lastal).  This means that if accessions_file_for_refsel_db_build is given, accessions_for_ref_genome_build is ignored -- again, same as for lastal, but may be somewhat confusing.

notestaff:
https://broadinstitute.slack.com/files/U1RQ3JN68/F7N6QMBG9/simple_refsel_stats.tsv

assembly.py
```diff
@@ -471,6 +472,52 @@ def order_and_orient(inFasta, inReference, outFasta,
...
+    log.info('n_genome_segments={} n_refs={}'.format(n_genome_segments, n_refs))
+
+    with util.file.tempfnames([ '.{}.ref.fasta'.format(ref_num) for ref_num in range(n_refs)]) as refs_fasta, \
+         util.file.tempfnames([ '.{}.scaffold.fasta'.format(ref_num) for ref_num in range(n_refs)]) as scaffolds_fasta, \
+         util.file.tempfnames([ '.{}.altcontig.fasta'.format(ref_num) for ref_num in range(n_refs)]) as alt_contigs_fasta:
+
+         for ref_num in range(n_refs):
+             Bio.SeqIO.write(ref_segments_all[ref_num*n_genome_segments : (ref_num+1)*n_genome_segments], refs_fasta[ref_num], 'fasta')
+
+         pool = multiprocessing.Pool(threads or None)
```
yesimon: `threading` might be a better choice here, especially if it's shelling out to other processes anyways

assembly.py
```diff
@@ -471,6 +472,52 @@ def order_and_orient(inFasta, inReference, outFasta,
     #    os.unlink(trimmed)
     return 0
 
+def _call_order_and_orient_orig(kwargs):
+    _order_and_orient_orig(**kwargs)
+
+def order_and_orient(inFasta, inReference, outFasta,
+        outAlternateContigs=None, outReference=None,
+        breaklen=None, # aligner='nucmer', circular=False, trimmed_contigs=None,
```
yesimon: Are these used?

assembly.py
```diff
@@ -471,6 +472,52 @@ def order_and_orient(inFasta, inReference, outFasta,
...
+    ''' This step cleans up the de novo assembly with a known reference genome.
+        Uses MUMmer (nucmer or promer) to create a reference-based consensus
+        sequence of aligned contigs (with runs of N's in between the de novo
+        contigs).
+    '''
+
+    ref_segments_all = tuple(Bio.SeqIO.parse(inReference, 'fasta'))
+    n_genome_segments = n_genome_segments or len(ref_segments_all)
+    assert (len(ref_segments_all) % n_genome_segments) == 0
+    n_refs = int(len(ref_segments_all) / n_genome_segments)
```
yesimon: Isn't `int(` not necessary here because of the assert?

assembly.py
```diff
@@ -471,6 +472,52 @@ def order_and_orient(inFasta, inReference, outFasta,
...
+    with util.file.tempfnames([ '.{}.ref.fasta'.format(ref_num) for ref_num in range(n_refs)]) as refs_fasta, \
+         util.file.tempfnames([ '.{}.scaffold.fasta'.format(ref_num) for ref_num in range(n_refs)]) as scaffolds_fasta, \
+         util.file.tempfnames([ '.{}.altcontig.fasta'.format(ref_num) for ref_num in range(n_refs)]) as alt_contigs_fasta:
+
+         for ref_num in range(n_refs):
+             Bio.SeqIO.write(ref_segments_all[ref_num*n_genome_segments : (ref_num+1)*n_genome_segments], refs_fasta[ref_num], 'fasta')
+
+         pool = multiprocessing.Pool(threads or None)
+         pool.map(_call_order_and_orient_orig,
+                  [dict(inFasta=inFasta, inReference=refs_fasta[ref_num], outFasta=scaffolds_fasta[ref_num],
```
yesimon: Prefer `{}` over `dict()` constructor.

pipes/config.yaml
```diff
@@ -140,6 +140,13 @@ samples_per_run: "samples-runs.txt"
 # to insufficient input data, consider expanding the list of accessions given to lastal.
 accessions_file_for_lastal_db_build: ""
 
+# File containing the list of accessions for reference genomes to use for reference selection.
+# Defaults to the same list as `accessions_for_ref_genome_build`.
+# For multi-segment genomes, the file must list all segments for the first reference, then all segments
+# for the second reference, and so on; for each reference, segments must be listed in the same order as for
+# `accessions_for_ref_genome_build`; all segments must be listed for each reference.
+accessions_file_for_refsel_db_build: ""
```
yesimon: Should this be a file? config["accessions_for_ref_genome_build"] seems to not be so?

assembly.py
```diff
@@ -471,6 +472,52 @@ def order_and_orient(inFasta, inReference, outFasta,
...
+        contigs).
+    '''
+
+    ref_segments_all = tuple(Bio.SeqIO.parse(inReference, 'fasta'))
+    n_genome_segments = n_genome_segments or len(ref_segments_all)
+    assert (len(ref_segments_all) % n_genome_segments) == 0
+    n_refs = int(len(ref_segments_all) / n_genome_segments)
+    log.info('n_genome_segments={} n_refs={}'.format(n_genome_segments, n_refs))
+
+    with util.file.tempfnames([ '.{}.ref.fasta'.format(ref_num) for ref_num in range(n_refs)]) as refs_fasta, \
```
yesimon: Are these temp filenames hidden files (start with `.` )?

pipes/config.yaml
```diff
@@ -274,6 +281,11 @@ ref_genome_dir: "ref_genome"
 # Can be a remote path prefix (s3://, gs://, etc.). The path should contain lastal.{ext} files
 lastal_ref_db_dir: "lastal_db"
 
+# The directory where the reference selection files should be stored
+# Can be a remote path prefix (s3://, gs://, etc.). The path should contain a refsel.fasta file.
+refsel_ref_db_dir: "refsel_db"
```
yesimon: Why not just be the path to the fasta directly?

assembly.py
```diff
@@ -471,6 +472,52 @@ def order_and_orient(inFasta, inReference, outFasta,
...
+        min_pct_id=0.6, min_contig_len=200, min_pct_contig_aligned=0.3, n_genome_segments=0, threads=0):
+    ''' This step cleans up the de novo assembly with a known reference genome.
+        Uses MUMmer (nucmer or promer) to create a reference-based consensus
+        sequence of aligned contigs (with runs of N's in between the de novo
+        contigs).
+    '''
+
+    ref_segments_all = tuple(Bio.SeqIO.parse(inReference, 'fasta'))
+    n_genome_segments = n_genome_segments or len(ref_segments_all)
+    assert (len(ref_segments_all) % n_genome_segments) == 0
```
tomkinsc: Could be helpful to have a message set in case the assertion fails.

assembly.py
yesimon: Nevermind - it's actually a suffix!

tomkinsc: They're not hidden as long as the prefix is set (defaults to `tmp`, so `tmp.{}.ref.fasta`...)
https://github.com/broadinstitute/viral-ngs/blob/master/util/file.py#L116

pipes/config.yaml
tomkinsc: Or perhaps `accessions_for_ref_genome_build` should be a file...

assembly.py
notestaff: args are passed through to _order_and_orient_orig()

assembly.py
notestaff: you mean to use multiprocessing.dummy ?

it's not just shelling out to other processes, it's calling tools.mummer.MummerTool.scaffold_contigs_custom() which does work in Python.  My understanding was that Python's Global Interpreter Lock limits parallelism?

pipes/config.yaml
notestaff: If it's non-blank then it must be a file, but if it's blank then it defaults to a file containing the list of accessions specified in accessions_for_ref_genome_build , as for lastal.

pipes/config.yaml
notestaff: It's the same interface as for lastal_db .  Right now refsel_db contains just the one fasta file, but in the future the reference selection data can get more complex, so it makes sense to have a directory for it.

assembly.py
notestaff: why?

yesimon: It's a google python style guide. I also see it in the wild much more than `dict()`

notestaff: If it's important to follow the Google python style guide I'll change it, but I prefer dict() for dictionaries that represent parameters of a function call, as opposed to more general dicts.  The syntax looks more like a function call, and permits adding **kwargs .

assembly.py
dpark01: What about `concurrent.futures.ProcessPoolExecutor`, which is used in a few other places already?

notestaff: That's not in Python 2.6, right?

notestaff: 2.7

tomkinsc: It's not, but the 2.7 build of viral-ngs [depends](https://github.com/broadinstitute/viral-ngs/blob/master/requirements-py2.txt) on the [`futures` package](https://pythonhosted.org/futures/) which provides it.

notestaff: changed it to use concurrent.futures.ProcessPoolExecutor

tomkinsc:
Does this allow for reassortment? i.e. Given a list of reference genomes, will it pick the best reference sequence for each segment, even if the segment sequences from GenBank are from disparate source samples? (Do we even want that behavior?) @dpark01?

notestaff:
"Does this allow for reassortment" -- not right now.  "Do we even want that behavior" -- probably virus-specific, since the opportunities for reassortment depend on the typical infection cycle?

tomkinsc:
Let's merge this in and open an issue to add support for reassortment (Boolean in config, probably).

notestaff:
I still need to add unit tests

tomkinsc:
It would be nice to have some of the stat gathering too, so we can regression test in the future.

assembly.py
```diff
@@ -471,6 +473,53 @@ def order_and_orient(inFasta, inReference, outFasta,
     #    os.unlink(trimmed)
     return 0
 
+def _call_order_and_orient_orig(inReference, outFasta, outAlternateContigs, **kwargs):
+    _order_and_orient_orig(inReference=inReference, outFasta=outFasta, outAlternateContigs=outAlternateContigs, **kwargs)
+
+def order_and_orient(inFasta, inReference, outFasta,
```
dpark01: Hm.. one thing. I kind of wonder if it might be preferable to have inReference be a list of fasta files (with argparse nargs='+')? Still thinking about this....

dpark01:
To @tomkinsc 's point, I do like the idea of having a (perhaps optional) output file with the stats that were used to rank the choices. Even prior to this PR (where only one reference was taken as input), users have always wondered what the # bases and % unambiguous were, since it is a failure criteria (at the beginning of impute_from_reference) but the user only sees the stats in the error message if it fails (and no stats if it succeeds).

assembly.py
notestaff: Can allow this and just concat all the listed fastas into one, while also permitting multiple references in each fasta.

dpark01: True... but then maybe you'd want to set `n_genome_segments` automatically based on the pre-concatenated inputs? If you think it's straightforward to make the CLI flexible either way, that's cool...

notestaff: Do we have any actual CLI users we can get input from, as to what CLI would be less confusing?  For pipeline users n_genome_segments gets set automatically from the config file.

notestaff:
"stats that were used to rank the choices" may get more complex if we use reassortment or SHIVER-style compound reference -- not sure how meaningful that'll be for the user...

assembly.py
dpark01: While the CLI may not have a lot of direct users for day to day use (more for one-off small analyses), the CLI is the common denominator between multiple ways of executing pipelines on viral-ngs. The snakemake rules are the oldest in our codebase, but even since their inception, I have never considered them sacred or primary. DNAnexus and Cromwell are both happy with the idea of giving users typed collections for inputs (e.g. `Array[File]`) and since argparse is too, why not let them play together?

I'm thinking more generally of the web user who wants to try a few references, whether it'd be easier to have them just specify a list of fasta files (each referring to one ref genome) or to have them concatenate them all to one file ahead of time (in the same order, etc).

For the lastal step, we have them concatenate, but there are two reasons why. One is that the lastal db often has a huge number of genomes (so it'd be crazy to have them specify hundreds of input files), and the other is that the ordering within the file doesn't matter (so there's nothing for them to mess up in terms of how they put it together).

In the case of multi-reference selection, I could imagine that most use cases (initially at least) would involve selecting from a small handful (maybe less than a dozen) options that span known phylogentic diversity within a species of interest. Which is why I was thinking that exposing this as a list of fasta files might make sense. And in order to expose it at the pipeline level, it's easier if the CLI supports it. We'll have to see how the usage evolves over time, but I initially lean towards having the user specify a list of fastas, at least for the web UI.

notestaff: On the command line you can already use <(cat ref1.fasta ref2.fasta ...) for inReference

notestaff: There's already a consistency check in Python that the number of sequences in inReference is a multiple of n_genome_segments, if n_genome_segments is specified.  If n_genome_segments is not specified, we assume (for backwards compatibility) that inReference contains one reference, so its segment count gives n_genome_segments.   If n_genome_segments is not specified and multiple inReference files are listed, we can assume that each contains one genome and that'll give the n_genome_segments.

Maybe the CLI could recognize a config file from which it could get some parameters if they're not specified on the command-line.  Right now the config file is tied to Snakemake, but loading some parameters makes sense outside that context, e.g. threads, tempdir, n_genome_segments, or other parameters that make sense across several commands.  Or default parameters for a specific CLI command could be specified in a config file.  There could be a project config file, which overrides the user config file, which overrides the system config file, as for .condarc etc.

dpark01: - if we cat multiple fastas together, there's no guarantee that each individual fasta has the right number of segments (only that the sum of them is a multiple of n_genome_segments), unless we enforce it outside of python (e.g. in the shell prior to invocation of viral-ngs)
- I don't like the idea of a CLI-level config file, I have trouble envisioning where that would make sense, and would much rather expose them as argparse params where sensible. The proper default value for `threads` is not at all universal across all the commands and use cases, the proper default value for `--tmp_dir` is already some subdirectory of `$TMPDIR` (in keeping with the behavior of most unix tools). The proper conda defaults should be handled at a layer of abstraction above the CLI. `n_genome_segments` is not only project specific, but potentially sample-specific within a project.

notestaff:
Added tests, and an option to output stats.   Let me know if more changes needed before merging.

assembly.py
```diff
@@ -489,6 +563,15 @@ def parser_order_and_orient(parser=argparse.ArgumentParser()):
                 but were not chosen for the final output.""",
         default=None
     )
+
+    parser.add_argument('--nGenomeSegments', dest='n_genome_segments', type=int, default=0,
+                        help="""Number of genome segments.  If 0 (the default), the `inReference` parameter is treated as one genome.
+                        If positive, the `inReference` parameter is treated as a list of genomes of nGenomeSegments each.""")
+
+    parser.add_argument('--threads', type=int, default=0, help='Number of threads to use (0 for all cpus)')
```
dpark01: can you remove the `--threads` argument and add `threads to the `util.cmd.common_args` invocation a few lines below instead?

PR #708, opened 2017-10-31
docker build optimization

dpark01:
This changes how the docker image is built. Changes include:
 - Change from installing the viral-ngs conda package (`easy-deploy-viral-ngs.sh setup`) to something closer to `setup-git` (that uses the current working copy instead of cloning fresh from github).
 - Divorce the conda build & deploy steps from the docker build & deploy -- put each in their own separate space in the Travis build matrix and cut end-to-end build time in half.
 - Separate out basic infrastructural setup in the Dockerfile to a [separate repository](https://github.com/broadinstitute/viral-baseimage) such that this Dockerfile only installs the components specific to viral-ngs and its dependencies. This also reduces build time by caching all of the apt-get steps.
 - Move Dockerfile to root of git repo. This allows for a direct build of either the github URI or the working copy (e.g. `docker build .` or `docker build https://github.com/broadinstitute/viral-ngs.git`).


.travis.yml
```diff
@@ -56,24 +59,30 @@ env:
   - secure: KX7DwKRD85S7NgspxevgbulTtV+jHQIiM6NBus2/Ur/P0RMdpt0EQQ2wDq79qGN70bvvkw901N7EjSYd+GWCAM7StXtaxnLRrrZ3XI1gX7KMk8E3QzPf0zualLDs7cuQmL6l6WiElUAEqumLc7WGpLZZLdSPzNqFSg+CBKCmTI8=
 
 git:
-  depth: 60
+  depth: 50
```
tomkinsc: The depth increase is so we can infer the most recent tag for setting a version?

.travis.yml
```diff
@@ -56,24 +59,30 @@ env:
...
 
 install:
-  - source travis/install-conda.sh
+  - if [[ "$BUILD_PACKAGE" != "docker" ]]; then source travis/install-conda.sh; fi
   - if [ -n "$PERFORM_TESTING" ]; then source travis/install-tools.sh; fi
   - if [ -n "$PERFORM_TESTING" ]; then travis/install-pip.sh; fi
+  - if [[ "$BUILD_PACKAGE" == "conda" ]]; then
+      conda install -q -y conda-build==3.0.25 &&
+      conda install -q -y anaconda-client &&
+      conda create -m -q -y -p tools/conda-tools/default Jinja2==2.8;
```
tomkinsc: Extra semicolon the result of a one-liner that grew?

docker/install-viral-ngs.sh
```diff
@@ -0,0 +1,21 @@
+#!/bin/bash
+
+set -e -o pipefail
+
+mkdir -p /opt/viral-ngs/viral-ngs-etc
+ln -s /opt/viral-ngs/source /opt/viral-ngs/viral-ngs-etc/viral-ngs
+ln -s /opt/miniconda /opt/viral-ngs/viral-ngs-etc/conda-env
+export VIRAL_CONDA_ENV_PATH=/opt/miniconda
+mv /opt/docker/easy-deploy-viral-ngs.sh /opt/viral-ngs
+mv /opt/docker/env_wrapper.sh /opt/viral-ngs
```
tomkinsc: Why not copy the env_wrapper.sh to `/opt/viral-ngs` during the docker build since the copy is in the derivative image anyway and not the more generic base image?

requirements-conda.txt
```diff
@@ -32,6 +32,5 @@ future==0.16.0
 matplotlib=1.5.3
 pysam=0.12.0.1
 pybedtools=0.7.10
-Jinja2=2.8
```
tomkinsc: This was moved because we are really only using it rendering the conda recipe?

.travis.yml
dpark01: Yeah.. I've found that it doesn't seem to add that much to the checkout time or `.git` size, since the really big stuff is probably hundreds of commits old at this point. And I've found that my own typical work on a dev branch can approach 50 commits prior to a PR unless I squash it, and if it goes beyond that, it fails to deploy the conda and docker packages because it can't figure out the version tag.

.travis.yml
dpark01: Oops.. yeah. Could be separate script too. I wonder if I can `set -e` in a travis.yml line? Otherwise I might want to replace the semicolon with a `|| exit 1`.

docker/install-viral-ngs.sh
dpark01: Good point, all of this stuff can/should go straight in.

requirements-conda.txt
dpark01: AFAICT, yeah, it isn't used for anything but building conda (which is why I moved it to that conditional line with the superfluous semicolon).

PR #712, opened 2017-11-02
Change thread defaults throughout

dpark01:
The CLI should default to fully utilize available resources, and threads should be limited in only three scenarios:
 - snakemake parallel execution on a single machine
 - shared HPC cluster execution (also snakemake)
 - unit tests when py.test xdist is enabled

In all other scenarios (cloud executions, local workstation executions, single command invocations), we should default to use all cores. This is one area where CLI defaults should differ from snakemake defaults.

This PR:
 - converts all CLI argparse parameters to default threads=`util.misc.available_cpu_count` (snakemake defaults unchanged)
 - ensures that argparse options are `type=int`
 - ensures that unit tests are always setting it to `_CPU`
 - converts some `--numThreads` argparse options to `--threads` for consistency
 - fixes some erroneous situations where `--threads` was offered, but not actually passed downstream
 - and has some topically unrelated optimizations to the Travis build for sphinx documentation


assembly.py
```diff
@@ -333,7 +333,7 @@ def parser_assemble_trinity(parser=argparse.ArgumentParser()):
         default=tools.trinity.TrinityTool.jvm_mem_default,
         help='JVM virtual memory size (default: %(default)s)'
     )
-    parser.add_argument('--threads', default=1, help='Number of threads (default: %(default)s)')
+    parser.add_argument('--threads', type=int, default=None, help='Number of threads (default: all available cores)')
```
yesimon: Why don't we set the default cores == ALL_CORES at the parser level?

interhost.py
```diff
@@ -384,7 +384,7 @@ def parser_general_mafft(parser=argparse.ArgumentParser()):
                 Note: if "--localpair" or "--globalpair" is specified this defaults to 1000.""")
     parser.add_argument(
         '--threads',
-        default=-1,
+        default=--1,
```
yesimon: ?

taxon_filter.py
```diff
@@ -287,7 +286,6 @@ def parser_deplete_bam_bmtagger(parser=argparse.ArgumentParser()):
                 and db.srprism.idx, db.srprism.map, etc. by srprism mkindex.'''
     )
     parser.add_argument('outBam', help='Output BAM file.')
-    parser.add_argument('--threads', type=int, default=4, help='The number of threads to use in running blastn.')
```
yesimon: Was there a good reason this was 4 previously?

test/integration/test_kraken.py
```diff
@@ -97,7 +99,9 @@ def test_kraken(kraken_db, input_bam):
 
     assert os.path.getsize(out_report) > 0
 
-    if 'TestMetaegenomicsSimple' in kraken_db:
+    '''
+    # not sure what to do with this failing test for now that never seemed to work in the past
```
yesimon: Was this spelling error always there?

tools/kraken.py
```diff
@@ -69,10 +72,10 @@ def pipeline(self, inBam, db, outReport=None, outReads=None, filterThreshold=Non
                            JVMmemory=picard.jvmMemDefault,
                            background=True)
 
-            if numThreads is None:
+            if not numThreads:
```
yesimon: If we're leaving `None` as the sentinel value (instead of 0), then I would prefer the`is None` conditional.

tools/trinity.py
```diff
@@ -62,12 +63,13 @@ def execute(self,
...
-                threads=1):    # pylint: disable=W0221
+                threads=None):    # pylint: disable=W0221
         if JVMmemory is None:
             JVMmemory = self.jvm_mem_default
         outdir = tempfile.mkdtemp(prefix='trinity-')
-        if int(threads) < 1:
-            threads = 1
+        if not threads:
+            threads = 10000000
+        threads = min(threads, util.misc.available_cpu_count())
```
yesimon: Should this be a util function?

assembly.py
dpark01: I think I like the idea of using `None`, partly because some of these tools already do some form of autodetect on their own when a parameter isn't specified... but some don't.. so I could go either way.

interhost.py
dpark01: Oops! How does that even pass the python parser?

taxon_filter.py
dpark01: I'm not sure. The one place I didn't change the hardcoded default of 4 was for IlluminaBasecallsToSam, which scales memory usage linearly with the number of cores, and can easily exceed JVMmemory on a large flowcell if we're not careful. But I don't think we ever had that issue with blastn...

test/integration/test_kraken.py
dpark01: we could git blame, but either way, feel free to take a look and see how this test branch ought to go.. I simply commented it out because that was effectively the same as misspelling the conditional, so it's no worse than before.

tools/kraken.py
dpark01: true.. I think I was defending against zero, but that was before I worked my way into changing all the argparse defaults more thoroughly..

tools/trinity.py
dpark01: I mostly copy-pasted this all over, but now that it's everywhere, I agree, it ought to be consolidated into a util.misc function.

dpark01:
@yesimon can you take a peek at my edits and let me know if this looks okay?

assembly.py
```diff
@@ -392,20 +375,20 @@ def assemble_spades(
...
-                        help='Contigs of high quality previously assembled from the same sample')
+                        help='Optional input contigs of high quality, previously assembled from the same sample')
     parser.add_argument('--contigsUntrusted', dest='contigs_untrusted', 
-                        help='Contigs of high medium quality previously assembled from the same sample')
+                        help='Optional input contigs of high medium quality, previously assembled from the same sample')
     parser.add_argument('--nReads', dest='n_reads', type=int, default=10000000, 
                         help='Before assembly, subsample the reads to at most this many')
     parser.add_argument('--filterContigs', dest='filter_contigs', default=False, action='store_true', 
-                        help='only output contigs SPAdes is sure of')
+                        help='only output contigs SPAdes is sure of (dorp lesser-quality contigs from output)')
```
yesimon: s/dorp/drop

test/unit/test_assembly.py
```diff
@@ -103,7 +103,7 @@ def test_empty_input_succeed(self):
 
         # run refine_assembly
         args = [inFasta, inBam, outFasta, "--chr_names", 'G5012.3', "--min_coverage", '3', "--novo_params",
-                "-r Random -l 30 -g 40 -x 20 -t 502"]
+                "-r Random -l 30 -g 40 -x 20 -t 502 -c {}".format(_CPUS)]
         args = assembly.parser_refine_assembly(argparse.ArgumentParser()).parse_args(args)
         print(args)
```
yesimon: Remove

util/misc.py
```diff
@@ -428,6 +428,40 @@ def available_cpu_count():
...
+    if 'PYTEST_XDIST_WORKER_COUNT' in os.environ:
+        threads = 1
+
+    max_cores = available_cpu_count()
+
+    if threads is None:
+        threads = max_cores
+
+    if type(threads) != int:
+        raise Exception()
```
yesimon: Maybe just an assert here instead of bare exception

util/misc.py
```diff
@@ -428,6 +428,40 @@ def available_cpu_count():
...
 
+def sanitize_thread_count(threads=None, tool_max_cores_value=available_cpu_count):
+    ''' Given a user specified thread count, this function will:
+        - ensure that 1 <= threads <= available_cpu_count()
+        - interpret None values to mean max available cpus
+            unless PYTEST_XDIST_WORKER_COUNT is defined as an environment
+            variable, in which case we always return 1
+        - allow for various, tool-specific ways of specifying
+            max available cores (tool_max_value)
+        tool_max_cores_value can be one of:
```
yesimon: Wouldn't this just be a num_cores override? Aka if defined set threads = the parameter. The naming and below logic is more complicated than necessary.

dpark01: The idea is that this is an override only if the user wanted max cores, but doesn't override if someone wanted a specific value. For example usages, see tools/gap2seq (where it is set to 0) or tools/mafft (where it is set to -1). The Novoalign tool doesn't really have a --threads param yet (it probably should) but there it should be set to None.

test/unit/test_assembly.py
dpark01: So, if a novoalign.lic file exists (and it does in Travis), I think Novoalign defaults to use all cores. For unit tests, we want it to default to _CPUS (which varies depending on whether xdist is enabled).

util/misc.py
dpark01: My understanding is that python `assert` works well in unit tests and in most scenarios, but if anyone ever calls our code with `python -O` it's all blindly ignored.

Though I guess this is specific example is one of the more defensive coding scenarios where it's not strictly necessary (because everyone should be passing in ints in theory), so maybe it's okay for this to be skipped in some cases.

util/misc.py
yesimon: It's a bit weird to see `tool_max_cores_value=available_cpu_count` in the function signature since `available_cpu_count` is a function. Normally we would set the default to `None` and call the default function within the function? Especially because `tool_max_cores_value` is supposed to be an int and not a function if it's actually specified.

dpark01: Yeah I agree.. I dislike that too. None would be the obvious default, I think the only reason I didn’t use it is in the case that the user passed None in this param, it could return None (for example if we added a threads param to Novoalign, it would simply not specify anything to let the tool choose max cores by itself). I actually intended the default return value to be available_cpu_count() (ie the int return value from calling the function) but python (or lint, I forget) hates when you do that kind of thing.

I agree we should do something better here..

PR #714, opened 2017-11-07
New kraken execution strategy to process multiple in one run.

yesimon:
Hacky input -> output determination to put all inputs/outputs in a single rule.

By default now memory lock the kraken database which will require modification of ulimits on most
systems.


pipes/rules/metagenomics.rules
```diff
@@ -74,65 +75,104 @@ rule diamond:
...
         mem     = 36
-    params: 
+    params:
         UGER = config.get('UGER_queues', {}).get('long', '-l h_rt=36:00:00')
     run:
         rna_bwa_path_prefix = strip_protocol(config["align_rna_db"], relative=True)
         taxonomy_db_prefix = strip_protocol(config["taxonomy_db"], relative=True)
         shell("{config[bin_dir]}/metagenomics.py align_rna {input.bam} "+rna_bwa_path_prefix+" "+taxonomy_db_prefix+" {output.nodupes_report} --dupeReport {output.report} --outBam {output.bam} --outReads {output.nodupes_lca} --dupeReads {output.lca} --JVMmemory {resources.mem}g --threads {resources.threads}")
 
+if config.get('kraken_execution', 'single') == 'multiple':
```
haydenm: What happens if config['kraken_execution'] is set to 'multiple' but the target is `.cleaned.kraken.report`? I'm not sure whether a rule can generate it in that case. Also, you might want to put 'kraken_execution' into config.yaml so people don't forget about it.

tools/kraken.py
```diff
@@ -55,51 +55,67 @@ def build(self, db, options=None, option_string=None):
...
+            fastq_pipes = pipes[:n_bams * 2]
+            kraken_output_pipes = pipes[n_bams * 2:]
 
             kraken_bin = os.path.join(self.libexec, 'kraken')
-            cmd = '''export KRAKEN_DEFAULT_DB={kraken_db}; {{ {kraken} --paired --fastq-input --threads {threads} {fastq1} {fastq2} 2>&1 1>&3 3>&- | sed '/Processed [0-9]* sequences/d'; }} \
-            3>&1 1>&2'''.format(
-                kraken_db=db,
+            env = os.environ.copy()
+            env['KRAKEN_DEFAULT_DB'] = db
+            cmd = '''{kraken} --lock-memory --paired --fastq-input --threads {threads} {outputs} {fastqs}'''.format(
```
haydenm: If `kraken` errors with an unhelpful message related to `--lock-memory`, you might want to put a message about ulimits (especially if this happens even if `kraken_execution` is `single` by default).

tools/kraken.py
```diff
@@ -55,51 +55,67 @@ def build(self, db, options=None, option_string=None):
...
-            }
+    def pipeline(self, db, inBams, outReports=None, outReads=None, filterThreshold=None, numThreads=None):
+        assert outReads is not None or outReports is not None
+
+        n_bams = len(inBams)
+        # 2n for paired fastq, 1n for kraken output
+        n_pipes = n_bams * 3
+
+        threads = util.misc.sanitize_thread_count(numThreads)
+
```
haydenm: Would it make sense to `kraken --preload` before processing all the samples -- e.g., if one sequential reading is better than all the random reads? This is what I've been doing on AWS before calling `snakemake ... all_kraken`, as I think you had suggested.

tools/kraken.py
yesimon: Yeah I put it in the kraken fork source code. Most systems the hard + soft limit are 64 pages of 4kb. It requires changing a config file to set it to be `unlimited` and reboot. This limit is also consistent on broad cluster and AWS, so I might make the lock optional rather than default.

tools/kraken.py
yesimon: My plan is to read/load the kraken database from pipe if the config is an S3 location, which is equivalent to `--preload`. Otherwise, it's kind of heuristic because not preloading allows you to go faster (only partially load db) if you only hit a small subset of kmers, especially if the backing disk is SSD. But if the backing disk is HDD and you have large inputs (like with `st1` EBS instances, then `--preload` will be a huge win.

pipes/rules/metagenomics.rules
yesimon: Yeah I need to add that part.

yesimon:
Yes - there will only be 1 single kraken process call that will process each input sequentially. Along with `--lock-memory`, this guarantees that there will be no page cache eviction - a source of "mysterious" slowdowns.

tools/kraken.py
haydenm: Ah, yeah, I have the database on an `st1` type so it makes sense for me, but I see your point if it's on SSD.

yesimon:
1. I believe it's tested through integration, but it's failing right now

2. I like the idea of making `kraken_multi` be the default, but perhaps the `lock-memory` part is disabled by default because that part requires OS changes.

test/unit/test_taxon_filter.py
```diff
@@ -37,7 +37,7 @@ def setUp(self):
         TestCaseWithTmp.setUp(self)
         self.polio_fasta = os.path.join(
             util.file.get_test_input_path(),
-            'TestMetagenomicsViralMix', 'db', 'library', 'Viruses', 'Poliovirus_uid15288', 'NC_002058.ffn'
```
dpark01: I just noticed this, why did you change the input to a unit test of an unrelated piece of code?

util/cmd.py
```diff
@@ -51,7 +51,7 @@ def common_args(parser, arglist=(('tmp_dir', None), ('loglevel', None))):
     for k, v in arglist:
         if k == 'loglevel':
             if not v:
-                v = 'DEBUG'
+                v = 'WARNING'
```
dpark01: why did this change?

tools/kraken.py
```diff
@@ -55,51 +55,118 @@ def build(self, db, options=None, option_string=None):
...
+    def db_opts(self, db, threads):
+        '''Determine kraken command line options based on db string'''
+
+        env = os.environ.copy()
+        def s3_psub(path):
+            cmd = 'aws s3 cp {} -'.format(path)
+            if path.endswith('.bz2'):
+                cmd += ' | lbzip2 -n {threads} -d'.format(threads=threads)
+            elif path.endswith('.gz'):
+                cmd += ' | pigz -p {threads} -d'.format(threads=threads)
```
dpark01: pigz doesn't exist in the conda package, though it does exist in the docker image.. we can make an effort later to harmonize some of this code later on

dpark01: s/lbzip2/bzip2/

test/unit/test_taxon_filter.py
dpark01: Oh I see, github diff doesn't show the relationship between the old .ffn and the new .fna.

util/cmd.py
yesimon: It was a bit loud - I'm not sure `DEBUG` was intentional or not.

tools/kraken.py
yesimon: `lbzip2` is the parallelized version of `bzip2`

dpark01: ok I'll add it to requirements-conda.txt in my pr in a minute

tomkinsc: There's also `pbzip2`, but I've seen some benchmarks that show `lbzip2` is a bit faster. My benchmark are still running.

PR #725, opened 2017-11-15
experiment with travis build stages

tomkinsc:
see: https://docs.travis-ci.com/user/build-stages/

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -15,7 +15,7 @@ matrix:
           os: linux
           python: 3.5
           env: BUILD_DOCS=true
-        - stage: conda
+        - stage: deployment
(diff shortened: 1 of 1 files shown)


dpark01:
I missed the build in progress, but does this serialize the stages, or does it run everything in parallel? I like the organization, but if it's significantly increasing the end-to-end time...

tomkinsc:
Yeah it serializes the stages, which in retrospect is probably not something we want. The thought was: "Why run the deploy jobs if the unit/integration tests fail?", but with considering we have `fast_finish: true` and it extends total build time it doesn't make sense. Closing this PR.

PR #728, opened 2017-11-17
WDL, Cromwell, and DNAnexus auto-build

dpark01:
This PR adds WDL pipelines.
- The WDL pipelines are mostly inspired by the previous DNAnexus implementations of viral-ngs.
- These pipelines all execute successfully from start to finish on Cromwell via Google. I have not checked for correctness of output, but spot checking looks sane so far.
- These pipelines all compile properly to DNAnexus workflows via dxWDL (except for one, which I disabled for now, and am not sure how necessary it is to have anyway).
- None of the resulting DNAnexus workflows execute properly yet, but we'll work on that one later.
- The Travis build process now adds a few more actions in the Docker portion of the build matrix. After deploying the image to DockerHub, it rewrites the runtime sections of the WDL files to point to the versioned image, it runs a WDL validator tool on all WDL workflows, it compiles all WDL workflows to a new DNAnexus CI project, and there is a placeholder for future Cromwell tests that do not yet exist.
- The above build actions happen on every branch commit and PR, so the long-form version string tags for docker images have been re-instated.


pipes/WDL/workflows/tasks/metagenomics.wdl
```diff
@@ -0,0 +1,111 @@
...
+    DB_DIR=$(mktemp -d)
+
+    # decompress DB to $DB_DIR
+    cat ${kraken_db_tar_lz4} |
+      read_utils.py extract_tarball \
+        - $DB_DIR \
+        --pipe_hint=${kraken_db_tar_lz4} \
+        --loglevel=DEBUG
+
+    ## prep input and output file names
```
yesimon: Can we try bringing the multiple processing back?

dpark01: Oh yeah, let me give that a try. I'd have to test it manually, since this current framework doesn't yet execute any of the WDL pipelines, but will do so.

dpark01:
@yesimon re: runtime resource requirements: there seems to be a lot of latitude given to the various execution engines and backends on how to deal with these. Cromwell-on-Google uses them and tries best to match them. Cromwell-on-local seems to ignore them entirely. dxWDL does its best in the same way Cromwell-on-Google does, but as you can see, sometimes you can just put an explicit hint in there as to what type of instance to choose. So it certainly isn't hard and fast...

dpark01:
Ok the multiple input kraken works!

PR #734, opened 2017-12-08
[WIP] break out snakemake tests

tomkinsc:
begin process of separating out all tests of snakemake pipelines from
unit and integration into a separate space & Travis job.


.travis.yml
```diff
@@ -199,6 +199,28 @@ jobs:
       before_cache:
         - conda clean --all --yes
 
+    - language: python
```
yesimon: Is there really no way to DRY this?

tomkinsc: Build matrix expansion and build stages could help, but with build stages we'd lose some concurrency and things are arguably less explicit with matrix expansion. Refactoring of the travis config can happen in a follow-on PR.

dpark01:
Hm... I kind of was expecting the Py3.6 unit+integration test time to go down more when you removed snakemake, but oh well.. I think it went down a little.

There's a more fundamental issue for another day, which is that some of the metagenomics tools (e.g. diamond I think, perhaps others) have no test coverage in unit or integration (no tests that actually execute the binary and confirm that it works)--the only real test of that was in the snake tests.

Yeah I do agree that this particular change makes more sense in the context of a future one that utilizes build stages, but no reason not to do it now.

PR #735, opened 2017-12-08
also login and push to mirror docker registry (docker.io)

tomkinsc:


.travis.yml
```diff
@@ -11,12 +11,15 @@ env:
...
-  #- DOCKER_REPO_PROD="broadinstitute/viral-ngs"
-  #- DOCKER_REPO_DEV="broadinstitute/viral-ngs-dev"
+  
   - DOCKER_REGISTRY="quay.io"
   - DOCKER_REPO_PROD="quay.io/broadinstitute/viral-ngs"
   - DOCKER_REPO_DEV="quay.io/broadinstitute/viral-ngs-dev"
+  
+  - DOCKER_REGISTRY_MIRROR="docker.io"
+  - DOCKER_REPO_PROD_MIRROR="broadinstitute/viral-ngs"
+  #- DOCKER_REPO_DEV_MIRROR="broadinstitute/viral-ngs-dev"
```
dpark01: `docker.io/broadinstitute/viral-ngs-dev` no longer exists, and it looks like your shell script modifications below don't do anything with $DOCKER_REPO_DEV_MIRROR anyway?

travis/deploy-docker.sh
```diff
@@ -6,6 +6,7 @@ if [[ -n "$DOCKER_PASS" && -n "$DOCKER_USER" && -n "$DOCKER_REGISTRY" ]]; then
 
 	# log in to Docker registry
 	echo "$DOCKER_PASS" | docker login --password-stdin --username "$DOCKER_USER" "$DOCKER_REGISTRY"
+	echo "$DOCKER_PASS_MIRROR" | docker login --password-stdin --username "$DOCKER_USER_MIRROR" "$DOCKER_REGISTRY_MIRROR"
```
dpark01: Can you add something like this?
```
if [ -n "$DOCKER_USER_MIRROR" -a -n "$DOCKER_PASS_MIRROR" -a -n "$DOCKER_REGISTRY_MIRROR" ]; then
```

I'm thinking of the scenario where someone forks our github repo and would only need to tweak their .travis.yml accordingly to work with their deploy repos (and maybe they don't have or want a mirror).

PR #741, opened 2017-12-14
refsel - permit failure of scaffolding to some references as long as at least one scaffolding succeeds

notestaff:


dpark01:
That build failure is weird and looks unrelated to your changes.  Can you change the `travis_wait` in [this line](https://github.com/broadinstitute/viral-ngs/blob/master/.travis.yml#L156) to `travis_wait 30` and see if that helps? I wonder if something is timing out on anaconda.org...

notestaff:
travis_wait doesn't seem to be the issue.  Its defaullt is 20 minutes, and the build fails before that.

dpark01:
I bet then it’s the thing where we have to throw in a sleep 4 or something like that after the conda build so that the failed process has enough time to write to stdout/stderr before Travis closes them... there are a few other examples of this in the travis/* scripts.

PR #742, opened 2017-12-18
Set kraken_execution default to 'single'

yesimon:


dpark01:
@yesimon @haydenm can you remind me what the motivation for this was, and whether this means we should also switch to single-bam execution on the WDL/Cromwell/DNAnexus side of things (even if it seems to be okay there so far)?

yesimon:
It seems like it causes problems with unrelated snakemake rules.

PR #750, opened 2018-01-06
dxWDL update to 0.57

dpark01:
Update dxWDL from 0.51 to 0.57. This enables the setting of File default parameters during compile and utilizes this feature for a number of depletion, metagenomic, and trim/clip databases.


test/input/WDL/test_inputs-assemble_denovo_with_deplete-dnanexus.dx.json
```diff
@@ -1,17 +1,16 @@
 {
-	"stage_1.raw_reads_unmapped_bam": { "$dnanexus_link": { "project": "project-F8PQ6380xf5bK0Qk0YPjB17P", "id": "file-F8F2kVQ09y3Q9Qj14fF806q2" } },
+	"stage-1.raw_reads_unmapped_bam": { "$dnanexus_link": { "project": "project-F8PQ6380xf5bK0Qk0YPjB17P", "id": "file-F8F2kVQ09y3Q9Qj14fF806q2" } },
```
tomkinsc: The inputs can be referred to as `stage-${stageNum}` rather than by explicit task name? That is a dxWDL [nuance](https://github.com/dnanexus/dxWDL/blob/0.57/src/main/scala/dxWDL/CompilerIR.scala#L37)?  It looks like there is a case statement for explicitly-named stages--a change for the future?

dpark01: I think I was unaware that you could also refer to input parameters for dxWDL-compiled DNAnexus workflows using the WDL task name (e.g. the defaults json). I've just been formatting the parameter names in the input.json according to whatever `dx run --help` or `dx describe` say about the compiled workflow--these have all called them `stage-N` for 0.57.

Generally speaking, the naming of the parameters for the compiled workflows has been a bit volatile with each new release of dxWDL and isn't guaranteed to be stable until their 1.0 release I think, so I've just been running with whatever naming convention du jour has been.

PR #751, opened 2018-01-08
option in depletion to revert tags in aligned files

tomkinsc:
Depletion calls RevertSam on already-aligned input. This adds an option to clear tags in the bam file as well, with a default tag list an an optional argument to explicitly specify the tags. These options are passed through to the WDL. Also bugfix blastdb prefix finding for large multipart blastdbs; parallelize mvicuna rmdup by library in response to #754


taxon_filter.py
```diff
@@ -65,6 +65,11 @@ def parser_deplete_human(parser=argparse.ArgumentParser()):
     )
     parser.add_argument('--srprismMemory', dest="srprism_memory", type=int, default=7168, help='Memory for srprism.')
     parser.add_argument("--chunkSize", type=int, default=1000000, help='blastn chunk size (default: %(default)s)')
+    parser.add_argument('--clearTags', dest='clear_tags', default=False, action='store_true', 
+                        help='When supplying an aligned input file, clear the per-read attribute tags')
+    parser.add_argument("--tagsToClear", type=str, nargs='+', dest="tags_to_clear", default=["XT", "X0", "X1", "XA", 
+                        "AM", "SM", "BQ", "CT", "XN", "OC", "OP"], 
+                        help='blastn chunk size (default: %(default)s)')
```
dpark01: this argparse help string looks incorrect!

pipes/WDL/workflows/tasks/taxon_filter.wdl
```diff
@@ -7,6 +7,8 @@ task deplete_taxa {
   Array[File]? bmtaggerDbs  # .tar.gz, .tgz, .tar.bz2, .tar.lz4, .fasta, or .fasta.gz
   Array[File]? blastDbs  # .tar.gz, .tgz, .tar.bz2, .tar.lz4, .fasta, or .fasta.gz
   Int?         query_chunk_size
+  Boolean?     clear_tags = false
+  String? tags_to_clear_space_separated
```
dpark01: `Array[String]?` perhaps? Then further below in the task command section you can use the `${sep=' ' tags_to_clear}` WDL string join notation.

dpark01: Also, because of the behavior of this pair of parameters, maybe it would be clearer to the user of this workflow if `tags_to_clear` were actually set to a default list that matches the argparse default in python, otherwise it's not clear that leaving it unspecified actually implies a decent default list (or why this requires two parameters instead of just one).

tomkinsc: I initially had it as `Array[String]`, but DNAnexus rendered a simple text box in the optional settings, and the expected input format was not clear (comma-separated, or space-separated?). Setting a default makes sense though, thanks!

dpark01: Ok, sounds like something to ask the dxWDL folks about later on.

tomkinsc:
I'll want to run this in DNAnexus prior to merging.

taxon_filter.py
```diff
@@ -371,6 +371,7 @@ def multi_db_deplete_bam(inBam, refDbs, deplete_method, outBam, **kwargs):
     for db in refDbs:
         if not samtools.isEmpty(tmpBamIn):
             tmpBamOut = mkstempfname('.bam')
+            print("db", db)
```
dpark01: What's this print line for?

PR #752, opened 2018-01-10
Update DNAnexus default files

dpark01:
A few fixes and optimizations:
1. Add default ERCC spike in database for the demux_plus workflow (forgot this earlier)
2. Change all bmtagger default DBs from fasta files to pre-indexed tarballs (since bmtagger on-the-fly indexing is very slow).
3. Move some of the very small depletion databasees (GRC ncRNA and metagenomics_v3) to blastn instead of bmtagger (and fasta instead of pre-indexed tarball). This adds minimal runtime to blast (since blast only runs once on a concatenated fasta) and reduces the number of bmtagger invocations to one (hg19 only).
4. Remove the hard coded test inputs for blast in the test dnanexus executions during build in order to test proper defaulting of blastn dbs (tests will still set the bmtagger db to something small in order to speed up runtime vs. the hg19 default).
5. Add some commented-out WiP WDL code for pre-demux tarball merging which may be required to get demux_plus to invoke automatically after a streaming upload in the future.


tomkinsc:
Do we want to consider uploading and using our Kraken database that includes insects? It's 98.45GB vs 75.38GB without.

dpark01:
Yeah I think that’s a good idea—should at least put the tar.lz4 in resources/production-defaults as an option

tomkinsc:
I kicked off a job to transfer `kraken-full-and-insects-20171106.db.tar.lz4` to `resources/production-defaults`.

PR #755, opened 2018-01-16
add align_and_plot workflow

dpark01:
per request of @kjsiddle and UCAD, a simple WDL/DNAnexus workflow to just align and plot coverage of raw reads against any genome.


dpark01:
@kjsiddle can you comment on whether this worked the way you wanted it to?

dpark01:
Let's just run with this for now, since it's a simple add.

kjsiddle:
Hi Danny,
Apologies for this. Actually I just tried and was not able to get the
workflow to work. Probably I'm doing something stupid, but I've tried on 4
samples (2xCMV, 2xLASV) that do contain viral reads but all finish with an
error. I've had a look through the errors but haven't yet figured out where
this is coming from. I'll share you on the project now.
On a related note, I was looking for the older version of demux-plus to try
on the rapid run. The earliest I can see (in OLD Broad viral-ngs dxWDL) is
v18.2. I recalled you mentioned going I might need to go back further than
that, do you know where I can find those?
Katie

PR #760, opened 2018-01-23
order_and_orient: if some contigs align ambiguously, warn but still process the rest

notestaff:


dpark01:
I'll try to look at this, but with a few other things on the plate, it'd help to know if this PR was motivated by a bug that a user was running into or whether it was self-motivated cleanup.

notestaff:
I ran into this on one of kjsiddle's HIV samples when experimenting with SPAdes.

dpark01:
Ok, and just to be clear, the previous behavior was to fail/error hard, and the new behavior is to drop such contigs (with a warning) and carry on?

notestaff:
Yes, i.e. to assemble rest of reference.  On the example I saw, the hole left by the missing contig is then filled by gapfilling.   Root cause, I think, was that SPAdes output a misassembled contig which glued together two non-adjacent pieces of reference, so the contig aligned to both; then it ends up with two alignments to reference.  Could add a step to break up such misassemblies, as in SHIVER and QUAST.

dpark01:
That sounds pretty good -- I think historically the AmbiguousAlignmentException was pretty rare, but it makes sense that it would pop up more with something like HIV, and that, in some of those scenarios, you'd have reason to suspect misassembly anyway.

test/unit/test_assembly.py
```diff
@@ -15,6 +15,7 @@
 import tempfile
 import argparse
 import itertools
+import gzip
```
dpark01: I think this import can be avoided... `test_assembly.py` already imports util.file which provides `open_or_gzopen` which is the more common pattern for opening gzip files.

notestaff:
this was actually LASV not HIV.

PR #761, opened 2018-01-24
Fix tar extraction bug under Docker

### LINKED ISSUE #758: possible demux or extract_tarball bug
At least as manifested on DNAnexus (dxWDL-based workflows), there appear to be some independent errors with running demux on a few flowcells:

- https://platform.dnanexus.com/projects/F5z8Jpj0Yqp6fFpXGfJVBg3b/monitor/analysis/F9GZ4XQ0YqpJVG0GB5Yx9FKY
- https://platform.dnanexus.com/projects/F95x4j80Z3ygZgQB0v92Gfyz/monitor/analysis/F9Fk8PQ0Z3yqbJ76JYZZbx5Y
- https://platform.dnanexus.com/projects/F95x4j80Z3ygZgQB0v92Gfyz/monitor/analysis/F9GJG200Z3yX1vv1JZ3V05zP

The first two are on master-branch dxWDL-compiled workflows, and the third one was on a dev branch that [disabled](https://github.com/broadinstitute/viral-ngs/commit/13129a48fdbe1d64fe0ab43e3017b5bed2252b59) the streaming download feature just in case that was confusing things. These are notably on independently created tarballs of two different flowcells.

I haven't yet followed the trail enough to pinpoint where this problem might lie -- whether in the extract_tarball code, the WDL wrappers, or I still haven't fully ruled out the possibility of user error in creating the tarballs.

dpark01:
Fixes #758 

When running `read_utils.py extract_tarball` as root in a system with GNU tar (e.g. our Ubuntu Docker container), the `tar` extraction will often error when trying to chown files to UIDs that don't exist. This is because GNU tar behaves differently when called as root, and needs the `--no-same-owner` flag to suppress that behavior (and do the same thing as when called as a normal user). We only add this flag when running as root and when tar is GNU.

Unrelated: also specifies `dx_instance_type` for a few small WDL tasks that didn't have it already defined. Fixes a spikein count error where it was running out of disk space, even for small inputs.


dpark01:
This also no longer suppresses `tar`'s stderr output, which might be important in failures. This also now requires a successful return code from the decompressor. In general we should assume that the user might put entirely malformed inputs at this stage, and that either `tar` or the decompressor might fail.

util/file.py
```diff
@@ -241,6 +241,9 @@ def extract_tarball(tarfile, out_dir=None, threads=None, compression='auto', pip
         elif compression == 'none':
             decompressor = ['cat']
         untar_cmd = ['tar', '-C', out_dir, '-x']
+        if os.getuid() == 0:
+            log.debug("tar is running as root. Assuming that this is GNU tar, appending --no-same-owner")
+            untar_cmd.append('--no-same-owner')
```
tomkinsc: Should we have a [conditional](https://github.com/broadinstitute/viral-ngs/blob/f8a8a699748a0628721799d00867503a72544098/tools/megan.py#L84) to only append this to `tar` in the non-Mac case?

dpark01: Ok, added a new conditional check. I didn't want to check for MacOSX though, I decided to check for GNU tar, since that's what the behavior is specific to (just in case someone wants to run viral-ngs as root on a FreeBSD system).

PR #762, opened 2018-01-24
add bwa depletion

tomkinsc:


tools/bwa.py
```diff
@@ -206,39 +206,35 @@ def align_mem_one_rg(self, inBam, refDb, outBam, rgid=None, options=None,
...
+        with util.file.tempfname('.aligned.sam') as aln_sam:
+            with samtools.bam2fq_tmp(inReads) as (fq1, fq2):
+                self.execute('mem', options + [refDb, fq1, fq2], stdout=aln_sam)
 
-        self.execute('mem', options + [refDb, fq1, fq2], stdout=aln_sam)
-        os.unlink(fq1)
-        os.unlink(fq2)
+            if min_score_to_filter:
+                # Filter reads in the alignment based on on their alignment score
+                aln_sam_filtered = util.file.mkstempfname('.sam')
```
yesimon: Output `.bam` by default as much as possible? Saving on disk IO typically makes up for compression/decompression CPU time.

tomkinsc: Great point. It should really be piped to `samtools` for output of a `.bam` intermediary, but this is older code we haven't gotten to yet. Maybe in a future PR where we made the change more broadly across the codebase. For now I just want to get bwa merged in...

taxon_filter.py
```diff
@@ -554,9 +529,89 @@ def wrapper(inBam, db, outBam, threads, JVMmemory=None):
 
     multi_db_deplete_bam(args.inBam, args.refDbs, wrapper, args.outBam, threads=args.threads, JVMmemory=args.JVMmemory)
     return 0
+__commands__.append(('deplete_blastn_bam', parser_deplete_blastn_bam))
 
 
-__commands__.append(('deplete_blastn_bam', parser_deplete_blastn_bam))
+@contextlib.contextmanager
```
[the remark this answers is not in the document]
dpark01: One day, we can add support things like bz2 and other fasta extensions (fna, faa), but let's leave that for later improvement.

taxon_filter.py
```diff
@@ -554,9 +529,89 @@ def wrapper(inBam, db, outBam, threads, JVMmemory=None):
...
+def deplete_bwa_bam(inBam, db, outBam, threads=None, JVMmemory=None):
+    'Use blastn to remove reads that match at least one of the databases.'
+
+    threads = util.misc.sanitize_thread_count(threads)
+
+    with extract_build_or_use_database(db, bwa_build_db, 'bwt', tmp_suffix="-bwa_db_unpack", db_prefix="bwa") as (db_prefix,tempDbDir):
+        with util.file.tempfname('.aligned.sam') as aligned_sam:
+            tools.bwa.Bwa().align_mem_bam(inBam, db_prefix, aligned_sam, threads=threads)
+            tools.samtools.SamtoolsTool().view(['-f0x4'], aligned_sam, outBam)
+            # TODO: consider using Bwa().mem() so the input bam is not broken out by read group
```
dpark01: All good TODO thoughts. As for the first one, the question is whether `bwa mem` mangles the RG headers (I think it does.. I think that's why we have the custom function. Since `deplete_bwa_bam` actually emits output that was, at some point, funneled through BWA itself (vs all the other depletion methods that just take inBam->outBam via Picard FilterSamReads), it's actually important to make sure that BWA mem doesn't mangle the headers.

taxon_filter.py
```diff
@@ -554,9 +529,89 @@ def wrapper(inBam, db, outBam, threads, JVMmemory=None):
...
+            # TODO: pipe bwa input directly to samtools process (need to use Bwa().mem() directly, )
+            #       with Popen to background bwa process
+
+def parser_deplete_bwa_bam(parser=argparse.ArgumentParser()):
+    parser.add_argument('inBam', help='Input BAM file.')
+    parser.add_argument('refDbs', nargs='+', help='One or more reference databases for bwa. '
+                         'An ephemeral database will be created if a fasta file is provided.')
+    parser.add_argument('outBam', help='Ouput BAM file with matching reads removed.')
+    #parser.add_argument("--chunkSize", type=int, default=1000000, help='FASTA chunk size (default: %(default)s)')
```
dpark01: The commented out `--chunkSize` and the actual argument `--JVMmemory` can probably both be removed here. This is a nice, Java-free command thus far.

taxon_filter.py
```diff
@@ -621,6 +676,53 @@ def merge_compressed_files(inFiles, outFile, sep=''):
...
+        inputFasta = new_fasta
+
+    if outputFilePrefix:
+        outPrefix = outputFilePrefix
+    else:
+        baseName = os.path.basename(inputFasta)
+        fileNameSansExtension = os.path.splitext(baseName)[0]
+        outPrefix = fileNameSansExtension
+
+    blastdb_path = tools.bwa.Bwa().index(inputFasta, output=os.path.join(outputDirectory, outPrefix))
```
dpark01: variable assignment seems misnamed and also unnecessary?

taxon_filter.py
tomkinsc: Thanks!

taxon_filter.py
```diff
@@ -76,15 +84,26 @@ def parser_deplete_human(parser=argparse.ArgumentParser()):
...
+    util.cmd.attach_main(parser, main_deplete)
     util.cmd.attach_main(parser, main_deplete_human)
     return parser
 
+def parser_deplete_human(parser=argparse.ArgumentParser()):
+    parser = parser_deplete(parser)
+    return parser
 
 def main_deplete_human(args):
+    ''' A wrapper around 'deplete', deprecated but preserved for legacy compatibility.
```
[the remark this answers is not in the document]
tomkinsc: Just hadn't gotten there yet, thanks for pointing it out.

tools/bwa.py
```diff
@@ -206,39 +206,35 @@ def align_mem_one_rg(self, inBam, refDb, outBam, rgid=None, options=None,
             #os.unlink(reheadered_bam)
 
     def mem(self, inReads, refDb, outAlign, options=None, min_score_to_filter=None,
-            threads=None):
+            threads=None, invert_filter=False):
         options = [] if not options else options
 
+        if '-t' not in options:
```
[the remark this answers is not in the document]
tomkinsc: This was a mistake, thanks.

tools/bwa.py
```diff
@@ -52,7 +52,7 @@ def index(self, inFasta, output=None, algorithm=None):
         self.execute('index', cmd)
 
     def align_mem_bam(self, inBam, refDb, outBam, options=None,
-                      min_score_to_filter=None, threads=None, JVMmemory=None):
+                      min_score_to_filter=None, threads=None, JVMmemory=None, invert_filter=False):
```
tomkinsc: The `invert_filter` argument is for the future (speculative), should we desire to do depletion based on alignment score rather than the `-f0x4` flag.

tools/bwa.py
```diff
@@ -210,9 +210,8 @@ def mem(self, inReads, refDb, outAlign, options=None, min_score_to_filter=None,
         options = [] if not options else options
 
         if '-t' not in options:
-            options.extend(('-t', str(threads)))
+            options.extend(('-t', str(util.misc.sanitize_thread_count(threads))))
 
-        threads = threads or util.misc.available_cpu_count()
```
dpark01: Actually, watch out about removing this line entirely, down below you're actually using the `threads` variable again for `samtools.sort`. Perhaps just precede the `if` block with `threads = util.misc.sanitize_thread_count(threads)` and then you can use that in both places.

tools/bwa.py
```diff
@@ -194,15 +195,7 @@ def align_mem_one_rg(self, inBam, refDb, outBam, rgid=None, options=None,
...
-            # sort
-            sorter = tools.picard.SortSamTool()
-            sorter.execute(
-                tmp_bam_aligned,
-                outBam,
-                sort_order='coordinate',
-                picardOptions=['CREATE_INDEX=true', 'VALIDATION_STRINGENCY=SILENT'],
-                JVMmemory=JVMmemory
-            )
+            shutil.move(tmp_bam_aligned, outBam)
```
dpark01: Because `tmp_bam_aligned` was created with `mkstempfname`,  you might need to actually `shutil.copy` instead, or maybe just delete the `else` block and have the `self.mem` call just write directly to `outBam` instead of the temp file.

tools/bwa.py
```diff
@@ -168,20 +170,44 @@ def align_mem_one_rg(self, inBam, refDb, outBam, rgid=None, options=None,
...
-        # if there was more than one RG in the input, we had to create a temporary file with the one RG specified
-        # and we can safely delete it this file
-        # if there was only one RG in the input, we used it directly and should not delete it
-        if removeInput:
-            os.unlink(one_rg_inBam)
+        with util.file.tempfname('.aligned.bam') as tmp_bam_aligned:
+            # rather than reheader the alignment bam file later so it has the readgroup information
+            # from the original bam file, we'll pass the RG line to bwa to write out
+            self.mem(one_rg_inBam, refDb, tmp_bam_aligned, options=options+['-R',
+                     readgroup_line.rstrip("\r\n")],
```
yesimon: Replace tabs with with `\t` or else samtools will warn about duplicate IDs even though they are in the `@PG` header line later.

tools/bwa.py
yesimon: `samtools sort` works on piped input - it keeps memory usage down by doing an external sort on disk, so it can be sped up by option `-m` for memory usage

dpark01:
@tomkinsc if you feel good about this, I think this is ready to go -- I'll let you merge it.

Issue #759, opened 2018-01-22
Allow "demux" on single-sample runs without samplesheet


tomkinsc:
Currently we make assumptions in various places that a sample sheet will be provided for basecalling (and demultiplexing). Our Picard wrapper code for "demux" should be able to run without a samplesheet, if Picard can run without one. Related to this, verify Picard will run on iSeq output.

Should we rename `illumina_demux` to something more generic?

dpark01:
I'm not totally sure, but I think for GitHub to automatically close an issue referenced by a merged PR, you need to precede it with magic words like "fixes" or something like that (I guess "addresses" isn't good enough).

dpark01:
Oh yes, here it is: https://help.github.com/articles/closing-issues-using-keywords/

PR #776, opened 2018-01-31
modify illumina_demux to support simplex sequencing runs

tomkinsc:
This modifies `illumina_demux` (which should perhaps be renamed) to read Illumina runs & produce BAM files, resulting in one bam per sample for multiplexed runs, or a single bam file bearing the flowcell ID for simplex runs.

This will be helpful for creating an output `.bam` file from raw basecall data where a single sample is sequenced in a run without any barcodes (so no samplesheet is provided).

This addresses #759.

### DIFF
--- a/illumina.py
+++ b/illumina.py
@@ -80,8 +80,8 @@ def parser_illumina_demux(parser=argparse.ArgumentParser()):
 
 
 def main_illumina_demux(args):
-    ''' Read Illumina runs & produce BAM files, demultiplexing to one bam per sample.
-        For simplex samples, a single bam will be produced bearing the flowcell ID.
+    ''' Read Illumina runs & produce BAM files, demultiplexing to one bam per sample, or 
+        for simplex runs, a single bam will be produced bearing the flowcell ID.
         Wraps together Picard's ExtractBarcodes (for multiplexed samples) and IlluminaBasecallsToSam
         while handling the various required input formats. Also can
         read Illumina BCL directories, tar.gz BCL directories.
--- a/illumina.py
+++ b/illumina.py
@@ -155,7 +155,9 @@ def main_illumina_demux(args):
         assert samples is not None, "This looks like a multiplexed run since 'B' is in the read_structure: a SampleSheet must be given."
     else:
         assert samples==None, "A SampleSheet may not be provided unless 'B' is present in the read_structure"
-        assert not args.commonBarcodes, "--commonBarcodes may not be specified unless 'B' is present in the read_structure"
+        if args.commonBarcodes:
+            log.warn("--commonBarcodes was set but 'B' is not present in the read_structure; emitting an empty file.")
+            util.file.touch(args.commonBarcodes)
 
     # B in read structure indicates barcoded multiplexed samples
     if multiplexed_samples:
@@ -209,7 +211,7 @@ def main_illumina_demux(args):
         tools.picard.IlluminaBasecallsToSamTool().execute_single_sample(
             illumina.get_BCLdir(),
             os.path.join(args.outDir,flowcell+".bam"),
-            "single_sample_"+flowcell,
+            flowcell,
             args.lane,
             flowcell,
             picardOptions=picardOpts,
@@ -520,10 +522,14 @@ class RunInfo(object):
 
     def get_flowcell(self):
         fc = self.root[0].find('Flowcell').text
-        if '-' in fc:
-            # miseq often adds a bunch of leading zeros and a dash in front
-            fc = fc.split('-')[1]
-        assert 4 <= len(fc) <= 9
+        # do not slice in the iSeq/Firefly case
+        if not self.get_machine().startswith("FF"):
+            if '-' in fc:
+                # miseq often adds a bunch of leading zeros and a dash in front
+                fc = fc.split('-')[1]
+        # >=5 to avoid an exception here: https://github.com/broadinstitute/picard/blob/2.17.6/src/main/java/picard/illumina/IlluminaBasecallsToSam.java#L510
+        # <= 15 to limit the bytes added to each bam record
+        assert 5 <= len(fc) <= 15
         return fc
 
     def get_rundate_american(self):
--- a/illumina.py
+++ b/illumina.py
@@ -522,14 +522,16 @@ class RunInfo(object):
 
     def get_flowcell(self):
         fc = self.root[0].find('Flowcell').text
-        # do not slice in the iSeq/Firefly case
-        if not self.get_machine().startswith("FF"):
+        # slice in the case where the ID has a prefix of zeros
+        if re.match(r"^0+-", fc):
             if '-' in fc:
                 # miseq often adds a bunch of leading zeros and a dash in front
-                fc = fc.split('-')[1]
+                fc = "-".join(fc.split('-')[1:])
         # >=5 to avoid an exception here: https://github.com/broadinstitute/picard/blob/2.17.6/src/main/java/picard/illumina/IlluminaBasecallsToSam.java#L510
         # <= 15 to limit the bytes added to each bam record
-        assert 5 <= len(fc) <= 15
+        assert len(fc) >= 5,"The flowcell ID must be five or more characters in length"
+        if len(fc) > 15:
+            log.warn("The provided flowcell ID is longer than 15 characters. Is that correct?")
         return fc
 
     def get_rundate_american(self):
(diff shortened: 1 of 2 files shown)


illumina.py
```diff
@@ -182,22 +195,34 @@ def main_illumina_demux(args):
...
+            barcodes_tmpdir,
+            flowcell,
+            args.lane,
+            basecalls_input,
+            picardOptions=picardOpts,
+            JVMmemory=args.JVMmemory)
+    else:
+        tools.picard.IlluminaBasecallsToSamTool().execute_single_sample(
+            illumina.get_BCLdir(),
+            os.path.join(args.outDir,flowcell+".bam"),
```
tomkinsc: If we break out a separate parser for simplex bcl-to-bam, it should allow the user to specify an output file rather than an output directory.

illumina.py
```diff
@@ -182,22 +195,34 @@ def main_illumina_demux(args):
...
+            "single_sample_"+flowcell,
```
tomkinsc: Open to suggestions of what to use here. The value must have a length >=5 chars per https://github.com/broadinstitute/picard/blob/2.17.6/src/main/java/picard/illumina/IlluminaBasecallsToSam.java#L510

tools/picard.py
```diff
@@ -536,4 +536,35 @@ def execute(
             'RUN_BARCODE=' + run_barcode, 'LIBRARY_PARAMS=' + library_params
         ]
         PicardTools.execute(self, self.subtoolName, opts, JVMmemory)
+
+    def execute_single_sample(self, 
+        basecalls_dir,
+        output_file,
+        run_barcode, # this is a run-specific ID, ex. the flowcell ID
```
dpark01: Maybe rename this to `read_group_id_precursor` because that's basically what it's being used for in this use case? I mean ... it might even be safe to just always use the flowcell ID, I don't really see this being something the user would want to fiddle with.

tools/picard.py
```diff
@@ -536,4 +536,35 @@ def execute(
...
+            opts_dict[k] = v
+        opts = []
+        for k, v in opts_dict.items():
+            if v is not None:
+                if type(v) in (list, tuple):
+                    for x in v:
+                        opts.append('='.join((k.upper(), str(x))))
+                else:
+                    opts.append('='.join((k.upper(), str(v))))
+        opts += [
```
dpark01: Thanks for setting `SAMPLE_ALIAS`. Can you also set `LIBRARY_NAME` to the same value? Our downstream iSNV code would appreciate it.

tools/picard.py
```diff
@@ -536,4 +536,35 @@ def execute(
...
+            'OUTPUT=' + output_file,
+            'BASECALLS_DIR=' + basecalls_dir, 
+            'LANE=' + str(lane),
+            'RUN_BARCODE=' + run_barcode, #must be >=5 chars per https://github.com/broadinstitute/picard/blob/2.17.6/src/main/java/picard/illumina/IlluminaBasecallsToSam.java#L510
```
dpark01: We could `assert len(run_barcode)>=5`

illumina.py
tomkinsc: Right, the issue is that (as currently parsed), iSeq flowcell IDs end up as shorter than 5 chars, resulting in an exception in picard. Unfortunately there seems to be some irregularity to how the iSeq flowcells are described in `RunInfo.xml`; here are two difference schemes we've seen so far:
- `BNS417-14-15-08`
- `BNS41706-1716`

The right solution is to probably pass `flowcell` as you mention and alter our flowcell ID-parsing code by machine type, but `RunInfo.xml` does not explicitly the instrument model. We could infer it based on the `<FlowcellLayout>` attributes. So far it looks like the iSeq `<Instrument>` text is prefixed with "`FF`" (presumably for "Firefly"), so maybe we can at least remove the MiSeq-specifc check that slices based on the presence of '`-`'. Is there a historical reason for the specific bounds set on the flowcell length assertion [here](https://github.com/broadinstitute/viral-ngs/blob/v1.19.1/illumina.py#L501)?

illumina.py
```diff
@@ -147,33 +149,44 @@ def main_illumina_demux(args):
...
-        #count_and_sort_barcodes(barcodes_tmpdir, args.commonBarcodes)
-        executor = concurrent.futures.ProcessPoolExecutor()
-        executor.submit(count_and_sort_barcodes, barcodes_tmpdir, args.commonBarcodes)
+    multiplexed_samples = True if 'B' in read_structure else False            
+    
+    if multiplexed_samples:
+        assert samples is not None, "This looks like a multiplexed run since 'B' is in the read_structure: a SampleSheet must be given."
+    else:
+        assert samples==None, "A SampleSheet may not be provided unless 'B' is present in the read_structure"
+        assert not args.commonBarcodes, "--commonBarcodes may not be specified unless 'B' is present in the read_structure"
```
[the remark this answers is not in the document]
tomkinsc: Good point. I'll change it touch the file if `args.commonBarcodes` is specified in the no 'B' case, and emit a log warning.

tools/picard.py
tomkinsc: Sounds good.

tools/picard.py
tomkinsc: Will do.

tomkinsc: `SAMPLE_ALIAS` and `LIBRARY_NAME ` are supposedly deprecated, but from what [the picard docs](https://broadinstitute.github.io/picard/command-line-overview.html#IlluminaBasecallsToSam) describe it sounds like picard expects barcode columns in the `LIBRARY_PARAMS` file (though I haven't yet confirmed that).

illumina.py
dpark01: Hm.. great question... maybe that length check doesn't need to be quite like that, or maybe historically we discovered other Picard problems if it was over 9 characters? Certainly the lower bound should be upped to 5 based on what we're seeing in Picard.

Yeah that dash-splitting code for MiSeqs should certainly be not used in the iSeq scenario. I wonder if MiSeq machine names always start with 'M'? I wouldn't count on the iSeq always starting with 'FF' until I see data from a post-release standard machine... ours doesn't even have the model name on front (like the promo pictures do) so I could see the 'FF' changing to something else w/o notice. Or maybe they'll keep it.

I have no idea what a normal Illumina HiSeq's machine name looks like, since the only output from HiSeqs we've seen are from the Broad GP's machines and I don't know if they have their own custom naming convention. I'm pretty sure when we've run a walkup flowcell at GP we didn't get an "M###" machine ID back (maybe we can check.. I think we might have a Zika ampseq walkup MiSeq run somewhere).

How about this.. according to the [comment](https://github.com/broadinstitute/viral-ngs/blob/v1.19.1/illumina.py#L499), that string splitting is just looking for leading zeros followed by a dash in front.. maybe we should just do a proper regex on `/^0+-/` and it won't match the iSeq/FireFly dashes?

tools/picard.py
tomkinsc: Though maybe setting `BARCODE=N` would work?

dpark01: You've already set `SAMPLE_ALIAS` here, and it seems like Picard hasn't turned off these options yet.. given that it seems semantically wrong to send in a `LIBRARY_PARAMS` file in this scenario, hopefully they won't drop support for it. You could do `BARCODE=N` but you'd have to check whether Picard gets confused when there's no B in the `read_structure`.

illumina.py
```diff
@@ -520,10 +522,14 @@ def get_fname(self):
...
-            # miseq often adds a bunch of leading zeros and a dash in front
-            fc = fc.split('-')[1]
-        assert 4 <= len(fc) <= 9
+        # do not slice in the iSeq/Firefly case
+        if not self.get_machine().startswith("FF"):
+            if '-' in fc:
+                # miseq often adds a bunch of leading zeros and a dash in front
+                fc = fc.split('-')[1]
+        # >=5 to avoid an exception here: https://github.com/broadinstitute/picard/blob/2.17.6/src/main/java/picard/illumina/IlluminaBasecallsToSam.java#L510
+        # <= 15 to limit the bytes added to each bam record
```
dpark01: I'm not sure that we have to enforce an upper bound for the sake of reducing BAM file size (Picard already truncates this anyway), but it's probably good to limit it anyway, or at least bring it to our attention if they end up longer than we thought.

tomkinsc:
This is indeed an incremental PR prior to making adjustments to the pipeline code. It does seem to work for non-barcoded iSeq runs.

PR #781, opened 2018-02-02
WDL: provide bmtagger more RAM and parallelize krona

### LINKED ISSUE #780: bmtagger OOM errors more often than it should
On DNAnexus / dxWDL workflows. Possible fix in progress on `dp-bmtagger` branch.

dpark01:
This PR:
1. Increases bmtagger (srprism) RAM limit in the WDL invocation to 90% RAM, up from 50% (since it never runs simultaneously with Java or anything else). [Fixes #780] Bug was non-deterministic, but recent assemblies using this branch seem to have cleared it up.
2. Uses GNU Parallel to parallelize up to `nproc` invocations of single-threaded Krona in the kraken WDL task. Addresses the observation that the Krona for loop was sometimes taking several times longer than Kraken itself. Also parallelizes the single-threaded tar/gzip calls after Krona. This is currently being [tested here](https://platform.dnanexus.com/projects/F95x4j80Z3ygZgQB0v92Gfyz/monitor/analysis/F9p7yV00Z3ypbB5pF5xzjY08), please wait a couple hours for this to succeed before merging the PR.
3. Adds `--quiet` back to `conda create` and `conda install` invocations in Travis to deal with log length limits.

### DIFF
--- a/pipes/WDL/workflows/tasks/metagenomics.wdl
+++ b/pipes/WDL/workflows/tasks/metagenomics.wdl
@@ -52,16 +52,16 @@ task kraken {
     wait # for krona_taxonomy_db_tgz to download and extract
 
     # run single-threaded krona on up to nproc samples at once
-    parallel \
+    parallel -I ,, \
       "metagenomics.py krona \
-        {}.txt.gz \
+        ,,.txt.gz \
         taxonomy \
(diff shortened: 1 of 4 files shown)


dpark01:
Okay this test run looks good.

PR #783, opened 2018-02-05
WDL assembly workflow updates and bugfixes

dpark01:
1. Bugfix: the `refine_2x_and_plot` combined task (introduced in the last month as an optimization to reduce instance spin up and staging time) was computing the `assembly_length` and `assembly_length_unambiguous` numbers incorrectly: it was computing on the input fasta (from scaffolding), not the output / final assembly.
2. Filename cleanups: more assembly WDL tasks are now stripping off suffixes left by previous steps (e.g. "taxfilt" or "assembly1-trinity") before appending their new suffixes, resulting in cleaner filenames throughout the assembly workflow. The final `refine_2x_and_plot` step now creates its final output filename based on the input reads bam instead of the input fasta (as in the pre-WDL dx workflows) and calls the final assembly *samplename*.fasta (instead of *samplename*.taxfilt.assembly1-trinity.scaffold.refine2.fasta).
3. Final assembly output file variable is now `final_assembly_fasta` instead of `refine2_assembly_fasta`.
3. `refine_2x_and_plot` was accidentally invoking `plot_coverage` twice--remove the first invocation.

### DIFF
--- a/pipes/WDL/workflows/tasks/assembly.wdl
+++ b/pipes/WDL/workflows/tasks/assembly.wdl
@@ -10,7 +10,8 @@ task assemble {
   String? assembler="trinity"  # trinity, spades, or trinity-spades
   String  cleaned_assembler = select_first([assembler, ""]) # workaround for https://gatkforums.broadinstitute.org/wdl/discussion/10462/string-type-in-output-section
 
-  String  sample_name = basename(reads_unmapped_bam, ".bam")
+  # do this in two steps in case the input doesn't actually have "taxfilt" in the name
+  String  sample_name = basename(basename(reads_unmapped_bam, ".bam"), ".taxfilt")
 
   command {
     set -ex -o pipefail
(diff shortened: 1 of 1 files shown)


pipes/WDL/workflows/tasks/assembly.wdl
```diff
@@ -95,7 +96,8 @@ task scaffold {
   Int?    nucmer_min_cluster
   Int?    scaffold_min_pct_contig_aligned
 
-  String  sample_name = basename(contigs_fasta, ".fasta")
+  # do this in multiple steps in case the input doesn't actually have "assembly1-x" in the name
+  String  sample_name = basename(basename(basename(contigs_fasta, ".fasta"), ".assembly1-trinity"), ".assembly1-spades")
```
tomkinsc: Not really 'cleaner,' but [`sub()`](https://software.broadinstitute.org/wdl/documentation/spec#string-substring-string-string) can probably be used here with a regex. Something like:

`String  sample_name = sub(contigs_fasta, "(?:\.assembly1\-spades)?(?:\.assembly1\-trinity)?\.fasta$", "")`
(not sure if the escaping is right for the WDL regex engine)

dpark01: Ooh good to know. As you note it'd probably require a little experimenting to make sure we get the syntax right with the escaping & string substitution.

PR #788, opened 2018-02-07
Add sample name to Krona report

### LINKED ISSUE #785: Add sample name to Krona output
Many users have noted how confusing it is to open up a bunch of krona html files as separate tabs in a browser when none of them provide any way to describe what sample they are talking about or input file they ran on.

The basename of the input file name should appear somewhere within the html file that Krona emits. There are several ways to do this:
1. Use the `-n` parameter of ktImportTaxonomy to rename the "Root" level to the basename of the input file. This puts the sample name at the center of the pie, and has the added advantage of showing up in the `<title>` of the html and therefore in the name of the tab in the browser.
2. Use the multiple input file mode of ktImportTaxonomy (see the [granola](http://marbl.github.io/Krona/examples/xml.krona.html) example) which defaults to giving the report a nice name corresponding to the basename of the input file. But if we only have one input file, we'd have to fake an empty second input file so that Krona uses dataset names in the first place.

I lean towards option 1.

Orthogonally, we could consider adding a multiple-input-file invocation of ktImportTaxonomy for the WDL-based tasks that already run kraken on a bunch of bams at once, but I wouldn't want to lose the per-sample html outputs.

While we're at it, we should also double check that `metagenomics.py krona` succeeds on empty input.

dpark01:
This PR alters Krona reports by changing the name of "root" to the input filename given to Krona. This allows a user to identify which sample a Krona report is talking about by seeing it at the center of the pie chart and in the title of the browser tab.

Test output is [here](https://platform.dnanexus.com/projects/F8PQ6380xf5bK0Qk0YPjB17P/data/tests/quay.io/broadinstitute/viral-ngs-dev/1.19.2-6-gc516b72-dp-krona/demux_plus) in the CI project. This also validates that Krona succeeds on empty input.

This fixes #785.


yesimon:
This looks good but for the blank input the krona output html is blank. It contains GUI but no pie chart and no root name as well. This might be somewhat confusing.

dpark01:
Yeah.. at least it doesn't error. If we think we don't like how krona deals with empty input, we could [create our own output](https://github.com/broadinstitute/viral-ngs/blob/master/tools/fastqc.py#L36).

PR #790, opened 2018-02-15
update bwa depletion db URI in snakemake config

tomkinsc:
Hold on merging for now, until the database file has been uploaded to s3


tomkinsc:
Ok, the file should be live now.

tomkinsc:
hg19 is still in bmtagger in the Snakemake config, but since bwa happens first the size of the input to bmtagger is significantly reduced. We can remove hg19 from bmtagger, but maybe let's wait for #791.

tomkinsc:
Closing; included in #791

PR #791, opened 2018-02-16
refactoring and optimizations related to bwa-based depletion

tomkinsc:
add RevertSam to the input side of each flavor of depletion function in taxon_filter.py (bwa, bmtagger, blastn). Add RevertSam after bwa depletion to strip alignment info from depleted bam file. refactor revertsam functionality to be a context manager in read_utils.py. add USE_THREADING=true as default for MergeSamFilesTool. bwa-based depletion now splits out into separate threads by rg. Removed redundant indexing of bwa temp files for multi-rg case. Pipe from samtools view for bwa. WIP(?) Todo: additional piping, filter bwa output of proper pairs only (-F0x2 rather than -f0x4; likely requires adjustments to unit test expected output)


tools/samtools.py
```diff
@@ -56,23 +56,30 @@ def install_and_get_path(self):
...
-        subprocess.check_call(tool_cmd, stdout=stdout, stderr=stderr)
+
+        env = os.environ.copy()
+        env.pop('JAVA_HOME', None)
+        if background:
+            subprocess.Popen(tool_cmd, stdout=stdout, stderr=stderr, env=env)
+        else:
+            subprocess.check_call(tool_cmd, stdout=stdout, stderr=stderr, env=env)
+
         if stdout:
```
dpark01: we should potentially avoid stdout/stderr.close() if background is true?

tools/bwa.py
```diff
@@ -195,8 +215,9 @@ def mem(self, inReads, refDb, outAlign, options=None, min_score_to_filter=None,
         samtools = tools.samtools.SamtoolsTool()
 
         aln_sam = util.file.mkstempfname('.aligned.sam')
-        with samtools.bam2fq_tmp(inReads) as (fq1, fq2):
-            self.execute('mem', options + [refDb, fq1, fq2], stdout=aln_sam)
+        fastq_pipe = samtools.bam2fq_pipe(inReads)
```
dpark01: nice. Is this basically streaming to an interleaved fastq output (if paired)?

tools/bwa.py
```diff
@@ -74,41 +79,55 @@ def align_mem_bam(self, inBam, refDb, outBam, options=None,
...
-                    options=options,
-                    min_score_to_filter=min_score_to_filter,
-                    threads=threads, 
-                    invert_filter=invert_filter
-                )
-                if os.path.getsize(tmp_bam) > 0:
-                    align_bams.append(tmp_bam)
-                else:
-                    log.warning("No alignment output for RG %s in file %s against %s", rg, inBam, refDb)
+
```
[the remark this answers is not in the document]
tomkinsc: This PR was a while ago, but I recall that being the case.

tools/bwa.py
tomkinsc: Yup! Setting the`-p` flag for bwa mem allows for interleaved input.

tools/samtools.py
tomkinsc: Yes, thanks

tomkinsc:
Yeah, the background flag is not used anywhere yet. Maybe I should just remove it?
The bam2fq_pipe() call is mirroring prior usage in [`last.py`](https://github.com/broadinstitute/viral-ngs/blame/master/tools/last.py#L49), which I notice does not have guardrails against samtools failure. I'll add a `.poll()` check for there too.

PR #792, opened 2018-02-16
kraken summary parser

tomkinsc:
This aggregates stats at a taxonomic level for lines in a kraken summary file below (or at the same level as) a named taxon. WIP


metagenomics.py
```diff
@@ -1143,6 +1144,162 @@ class KrakenBuildError(Exception):
...
+                # 0.00  16  0   D   10239     Viruses
+                #
+                # row[0] Percentage of reads covered by the clade rooted at this taxon
+                # row[1] Number of reads covered by the clade rooted at this taxon
+                # row[2] Number of reads assigned directly to this taxon
+                # row[3] A rank code, indicating (U)nclassified, (D)omain, (K)ingdom, (P)hylum, (C)lass, (O)rder, (F)amily, (G)enus, or (S)pecies. All other ranks are simply '-'.
+                # row[4] NCBI taxonomy ID
+                # row[5] indented scientific name
+
+                if indent_of_line == indent_of_selection:
```
yesimon: After finishing all the children selection, there's no guarantee the next line will have the same indent as the selection indent, so this should be `<=`

metagenomics.py
```diff
@@ -1143,6 +1144,162 @@ class KrakenBuildError(Exception):
...
+
+        with util.file.open_or_gzopen(f, 'rU') as inf:
+            should_process = False
+            indent_of_selection = -1
+            currently_being_processed = ""
+            for line in inf:
+                if len(line.rstrip('\r\n').strip()) == 0:
+                    continue
+                csv.register_dialect('kraken_report', quoting=csv.QUOTE_MINIMAL, delimiter="\t")
+                row = next(csv.reader([line.strip().rstrip('\n')], dialect="kraken_report"))
```
yesimon: Using a `DictReader` would make the below easier to understand.

metagenomics.py
```diff
@@ -1143,6 +1144,162 @@ class KrakenBuildError(Exception):
...
+                        if row[3] == rank_code(taxlevel_focus):
+                            same_level = True
+                        if row[3] == "-":
+                            log.warning("Non-taxonomic parent level selected")
+                
+                if should_process:
+                    if (rank_code(taxlevel_focus) == row[3]):
+                        
+                        if int(row[1])>=count_threshold:
+                            sample_summary[currently_being_processed][row[5]] = [float(row[0]), int(row[1])]
```
yesimon: Make values NamedTuples here, or just store the entire row in `sample_summary` ?

metagenomics.py
```diff
@@ -1143,6 +1144,162 @@ class KrakenBuildError(Exception):
...
+
+        
+        for k,taxa in sample_summary.items():
+            sample_summary[k] = collections.OrderedDict(sorted(taxa.items(), key=lambda item: (item[1][1]) , reverse=True)[:top_n_entries])
+
+            if len(list(sample_summary[k].items()))>0:
+                log.info("{f}: most abundant among {heading} at the {level} level: \"{name}\" with {reads} reads ({percent:.2%} of total); included since >{threshold} read{plural}".format(
```
yesimon: line length

metagenomics.py
```diff
@@ -1143,6 +1144,162 @@ class KrakenBuildError(Exception):
...
+                writer.writerow(sample_dict)
+                
+
+            csv_out.close()
+
+        # otherwise write out a histogram
+        else:
+            count = 0
+            summary_counts = collections.defaultdict(dict)
+            for sample, totals in samples.items():
```
yesimon: Using pandas DataFrame might make these axis sum operations easier (and perhaps write out) - and also condense with above?

tomkinsc:
Thanks, yeah this was quick and dirty for a call last week and maybe should just be trashed entirely.

metagenomics.py
dpark01: @tomkinsc would you like to revive this PR for merging in soon? If so, it looks like of @yesimon 's comments last month, this is the only one about correctness that ought to be address, while the rest are stylistic.

tomkinsc: Potentially. I delivered some output to Anne for her work, and may incorporate changes based on her suggestions, if any.

metagenomics.py
tomkinsc: Using pandas would help, but we don't currently have it as a dependency (and it seems a little silly to include just for this).

metagenomics.py
tomkinsc: Thanks

metagenomics.py
tomkinsc: Thanks

metagenomics.py
notestaff: "seems a little silly to include just for this" -- problem is, I think this every time I miss pandas, and there've been several over time.

dpark01: Many years ago I did a similar hesitant but eventual embrace of numpy/scipy for viral-ngs.. back before our conda-based installs, packages like this incurred a huge install-time cost on Travis, because `pip install` did a ton of compiling (about 15 mins worth of gcc/make). So I rewrote a bunch of stuff to avoid using it. Conda greatly reduces that install time now, and eventually numpy (but not scipy) was included as a conda dependency by something else (not sure exactly what), so I re-embraced numpy in the code. But I do think we need to carefully consider what the additional payload is for pandas+dependencies before dropping it in. We're already pretty heavyweight...

tomkinsc:
Change in reply to @yesimon's feedback have been incorporated, and this is ready for review.

metagenomics.py
```diff
@@ -1143,6 +1144,172 @@ class KrakenBuildError(Exception):
...
+    parser.add_argument('--countThreshold', type=int, dest="count_threshold", help='Minimum number of reads to be included (default: %(default)s)', default=1)
+    parser.add_argument('--noHist', action='store_true', dest="no_hist", help='When topN==1, write out a report by-sample rather than a histogram.')
+    parser.add_argument('--zeroFill', action='store_true', dest="zero_fill", help='When absent from a sample, write zeroes (rather than leaving blank).')
+    util.cmd.common_args(parser, (('loglevel', None), ('version', None), ('tmp_dir', None)))
+    util.cmd.attach_main(parser, taxlevel_summary, split_args=True)
+    return parser
+
+def taxlevel_summary(summary_files_in, json_out, csv_out, tax_headings, taxlevel_focus, top_n_entries, count_threshold, no_hist, zero_fill):
+    """
+        By default, when --taxHeading is at the same level as --taxlevelFocus
```
notestaff: add summary sentence of what the command does?

tomkinsc: Done, thanks!

PR #799, opened 2018-03-23
pass sample name in WDL call to plot_coverage

tomkinsc:
pass sample name in WDL call to plot_coverage so the same name shows up in the coverage plot PDF


tomkinsc:
Thanks! I missed that call. Added it to `refine_2x_and_plot`.

PR #801, opened 2018-03-23
metadata recording

notestaff:
This PR adds an infrastructure for centralized recording of metadata for CLI commands.  The infrastructure is agnostic with respect to where analyses are run (locally or on one of the clouds), how commands are organized into workflows, and where the data and recorded metadata is stored.   The goal is to gather into one place and in uniform format the parameters and metrics of all/most analyses run in the lab (and maybe by other users if they opt in), to better understand which workflows need improvement, and to simplify comparisons of alternate workflows run on same/similar data.


dpark01:
Okay thanks Ilya. Just FYI, since this is a pretty heavy diff, there'll be a lot to look through, and with all the Lassa activity recently I can't guarantee a speedy review, so I'd appreciate your patience on this. One thing that might help navigate through this is an in-person presentation/code review with a few folks at some point later on.

A few quick questions, not having actually read through the details yet:
1. Why write the stats to object storage vs something that is more amenable to logging stats, e.g. Amazon DynamoDB, Google's BigTable, StackDriver (both AWS+GCP), etc? You could more easily restrict permissions to just the "C" of the "CRUD", and make it much simpler to query the results.
2. Related to that, does this instrument the `logging` module or does it do its own thing?
3. What is the behavior with either total or partial network outage? Are communications asynchronus or blocking? Is runtime of any analysis affected by total or partial connectivity? (these are all asking the same question in different ways)

Again, hope you don't mind letting this one take a little time.

notestaff:
No rush.  Re: questions, destination is easy to change; I just started with logging to a filesystem because it's easy to debug, and through pyfilesystem2 supports S3 and other destinations.  @tomkinsc made a helpful list of possible destinations; I'll add Google Stackdriver option.  Re: network connectivity, any errors/exceptions in metadata recording code get converted to warnings; I'll also add a timeout so it never takes more than a set time, and make it run concurrently with the command as much as possible.  But I haven't seen much slowdown so far.

PR #809, opened 2018-04-04
add dnanexus demux_launcher with travis test

mlin:
@dpark01 @tomkinsc comments on this? (the commits are meant to be squashed obviously)

The `demux_launcher` native DNAnexus applet is invoked by the streaming upload tool to consolidate the uploaded tarballs for the instrument run, and then invoke the demux WDL workflow on each lane.


dpark01:
@tomkinsc any comments? Resulting run is [here](https://platform.dnanexus.com/projects/F8PQ6380xf5bK0Qk0YPjB17P/monitor/job/FF2Yx8Q0xf5j9Qjx9yZ442j3).

@mlin this looks quite nice and seems to do just what we'd want at first glance. From what I can tell it ... is a single native dx applet that launches-and-blocks on a WDL workflow (demux_plus) and then tallies the workflow's outputs as its own. Does that seem about right?

A few questions:
1. Looking at the example job linked above, the name of the applet seems a bit ugly.. given that we're compiling it to a subdirectory that contains the full versioned docker tag name, can we just name the applet "demux_launcher"?
2. We can separately test this out a bit on some multi-lane flowcells just to see if that works (non-travis). Also curious to see what the tarball-repack performance looks like with pigz on an _x8. Looks like you're minimizing redundant disk I/O by piping download/upload as much as possible and skipping the repack when it's not necessary, awesome.
3. Are the RunInfo.xml and SampleSheet.csv files uploaded by the uploader scripts? I had thought they were unpacked and outputted by the old demux_plus (and would need to be emitted by this applet) but maybe I was wrong about that.

tomkinsc:
This is great, thanks for putting the time in on this, @mlin. Re: @dpark01's points:

1. I second @dpark01 on the naming, though it may be nice to report the viral-ngs version as an output field.

2. It'll be straightforward to merge in instance scaling, perhaps as we had before, based on [tile count](https://github.com/dnanexus/viral-ngs/blob/dnanexus/demux/viral-ngs-demux-wrapper/src/viral-ngs-demux-wrapper.sh#L17-L57). I'm not sure I understand the reason for re-packing the tarballs--that's only because our current demux workflow accepts only a single tarball (stream) as input? Can we accept a stream of multiple tarballs and bylass the consolidation if we pass `-i` to `tar` to ignore EOF? (This would depend on a change to [`read_utils.py::extract_tarball()`](https://github.com/broadinstitute/viral-ngs/blob/master/util/file.py#L201)). I assume the advantage of spinning up separate demux instances per-lane is greater than the added instance startup overhead since the tarball transfers can happen in parallel? I wonder if we should think about changing the uploader to send [`zstd`](https://facebook.github.io/zstd/)-compressed tarballs. Faster than gzip with the same or better compression ratio. With Ubuntu adopting it, zstd may become mainstream soon. 

3. @dpark01: [Looks like they're uploaded separately](https://github.com/dnanexus-rnd/dx-streaming-upload/blob/365348ed53dcd76086fd885923a1582231dc171d/files/incremental_upload.py#L395-L399)?

mlin:
The long name actually comes from travis/tests-dx.sh where it launches the test job with the lengthy version tag. The other test jobs (for demux_plus, assemble_denovo_with_deplete, etc.) are named similarly. The applet itself is just named demux_launcher, which is what a job would be named by default unless the invoker (incremental uploader) specifies `dx run ... --name xxx`

Testing on any existing run from the incremental upload tool should be as simple as `dx run demux_launcher -i upload_sentinel_record=xxx`. The applet has the demux_plus workflow ID as a default embedded with each build.

Re compression algos...I'd second Zstandard which is what the cool kids are using these days. In addition to dominating gzip as @tomkinsc mentioned, it also has a new fast mode that approaches lz4 (see the 'Faster compression levels' table [here](https://github.com/facebook/zstd/releases/tag/v1.3.4))

dpark01:
Well if the cool kids on El Camino are all doing it, it's just a matter of time before we adopt it too. But that's for a future PR. I will note the observation that the pigz piped upload/download for large flowcells in this iteration is definitely not saturating the network pipe (something like 30-60MB/sec, see the below example runs) so there's definitely gains to be had.

Back to this PR. Yeah no worries about the naming, I now realize that's not the applet's name, just the job name, so no big deal.

I don't entirely get how the mechanics of these native applets that call sub jobs/analyses work, but I noticed that when I kick of a job where it consolidates tars, when it's done with the consolidate job and in the middle of executing demux_plus, it seems that the consolidated tarball lives on DNAnexus as a file-xxxx object that the downstream jobs can consume (and I can dx describe and such) but it doesn't yet appear in the output directory of demux_launcher until the whole thing has completed running. Is that something we can change, or is that just how these types of applets work?

Also, again a bit of a structural mechanical question, but when there is a failure somewhere in some sub-step of demux-plus, how do I determine which subjob of which analysis was the root cause of the failure (since I think everything terminates once one of them fails, it's hard to tell which one was the cause). I know how to sort this out for normal scattered workflows, but there's multiple layers of nesting going on here. Here's an example failure (8 lane HiSeq 2500): https://platform.dnanexus.com/projects/FB2y4500QZ7547qG1V9Y048x/monitor/job/FF3JB380QZ79jvF36B5fBk56

Here's an example success (2 lane HiSeq 2500): https://platform.dnanexus.com/projects/FB2y4500QZ7547qG1V9Y048x/monitor/job/FF3JBqj0QZ78125BGf9qb5GX

mlin:
Yeah, the default behavior of individual applets is to deliver results "atomically" to the parent project on success, and nothing on failure. There are a few ways to override that which I can see being useful in this case where we might not want a failure in any one sample to hold up delivering all the other results. I will look into this next week. Before being copied into the project, the results are held in a 'workspace' which is a kind of temporary project that exists during the job run, which is normally cleaned up promptly but you can tell the system to keep it around with `dx run ... --delay-worksace-destruction` and go back later to inspect/debug it or copy stuff out.

There is indeed an issue currently where the information about the specific failing sub-job isn't bubbled all the way to the top when a workflow is called from another job or workflow, as in this case. The team is tracking that cc @comandlinegirl @geetduggal @orodeh

mlin:
Agreed- looking into it

dpark01:
@mlin I think we're going to close this PR in favor of #838 (which is a branch off this branch), does that sound right?

mlin:
SGTM!

PR #812, opened 2018-04-30
bugfix: filename errors in reports.plot_coverage

dpark01:
the samtools flagstat invocation in reports.plot_coverage was looking for bam files that didn't exist.


dpark01:
merging in so it's ready for tomorrow's training

PR #826, opened 2018-06-22
rewrite of feature table remapping function

tomkinsc:
This PR includes a class-based rewrite of the sequin tsv feature table parser to better handle the various cases that can occur in the file format. multi-location features (exons, etc.) are now allowed, as are fuzzy locations. Feature keys are now checked against an object representing the spec. feature locations are now sorted within each feature (though features themselves can occur in any order). bugfix to delimit note key from value with tab instead of space. add unit tests for the feature table remapping functionality


test/input/TestFeatureTransfer/synthetic/input/ref.tbl
```diff
@@ -0,0 +1,25 @@
+>Feature gb|AY772699.1|
+<1	99	CDS
+			gene	should-be-output-if-not-ignore_ambig_feature_edge
```
dpark01: I know this is just an artificial feature (the `<1` - `99` CDS) for testing purposes, but it seems weird to have a "gene" tag within a CDS feature. can we turn this into a "product" or something?

dpark01: otherwise, nice battery of edge cases

PR #828, opened 2018-06-27
Add WDL for iSNVs/v-phaser

tomkinsc:
This PR adds WDLs for running v-phaser on one sample to output iSNVs. It probably makes sense to wait to do multi-sample via scatter-gather until we update everything to [WDL 1.0](https://github.com/openwdl/wdl/blob/master/versions/1.0/SPEC.md) which has a `Pair` datatype for gathering multiple files (we're using WDL draft-2 currently).
Also, this should probably be merged via rebase due to all of the dx debugging...


dpark01:
Okay this is a big one... I think I'll need some deconstructed explanation on each piece. So far, I see a few things in this PR:

1. A ton of diffs that have to do with updating to the latest dxWDL + Cromwell and all the various changes that imposed.
2. A new WDL workflow for single sample iSNVs.
3. Some decent sized changes to `intrahost.merge_to_vcf`

A few questions and thoughts:
- How interconnected are those three changes? Are they somewhat orthogonal/independent or did **2** necessitate both **1** and **3**?
- I've been watching some of the dxWDL updates from a distance and haven't yet given it a spin myself, but I was concerned about how they dropped support for implicit workflow inputs (percolated up from unlinked task inputs). But it looked like the various new releases of dxWDL have gone back and forth on supporting unlocked workflows.. not sure how related that was. I dunno, if the Cromwell/OpenWDL community more broadly has generally moved away from implicit workflow inputs then I guess we can follow, but I'm not terribly happy with that direction, but if this was only dxWDL we can inquire about it.
- What motivated change **3**? That particular method is a pretty hairy beast to dive into.. lint has complained about how overly complex it is but I've struggled to find ways to simplify and break it down because it really is managing a ton of edge cases at once (and I'm not absolutely sure that all those cases are well tested).

Anyway, breaking it down a bit will help navigate the actual diffs.

tomkinsc:
- The three changes are fairly orthogonal, however w/r/t changes 1&2, in the course of writing the new WDL workflows I wanted to target the newer dxWDL + Cromwell (at least under the draft-2 spec). They're interdependent as included here. 
- The trend seems to be toward requiring inputs to be explicitly specified. With the combination of versions of dxWDL and Cromwell used in this PR with the draft-2 format WDL files, explicit declaration appears to extend only to required inputs. With the move to [WDL 1.0](https://github.com/openwdl/wdl/blob/master/versions/1.0/SPEC.md) we may need to also explicitly list the optional inputs from tasks in workflows (not sure yet). It will become apparent when we start updating from WDL draft-2 to 1.0 format. I haven't come across a migration guide, and it is a bit difficult to tell when comparing the specs. In general, we will need to make fairly significant changes to support WDL 1.0... 
- Change 3 is two things: 
  - a block that infers sample names from the v-phaser output files in the case where a list of sample names is not passed in. This is to reduce the number of inputs required by `intrahost.merge_to_vcf()`, with the WDL/DNAnexus implementation in mind.
  - a bugfix to encode+report ALT alleles in the [VCF](https://samtools.github.io/hts-specs/VCFv4.2.pdf) GT fields (not sure if anyone was bitten by the bug?). This should have been caught by our tests, except the sample names of the tests were not changed to the `sampleName-chr#` format when we changed the rest of the code to use the`-chr#` suffix format for dealing with multiple chromosomes.

intrahost.py
```diff
@@ -531,17 +546,17 @@ def merge_to_vcf(
...
-                # get two independent iterators into the alignment file
-                number_of_aligned_sequences = count_iter_items(Bio.SeqIO.parse(inf, 'fasta'))
-                # -1 is to account for inclusion of reference in the alignement in addition
-                # to the assemblies
-                if not (number_of_aligned_sequences - 1) == len(isnvs) == len(samples):
-                    raise LookupError(
-                        """samples, isnvs, and alignments must have the same number of elements
-                        (plus an extra reference record in the alignment).
-                        %s does not have the right number of sequences""" % fileName)
+        # for fileName in alignments:
```
dpark01: This commented out section of `intrahost.merge_to_vcf` appears to have been simply a pre-screening sanity check / safety check to make sure the inputs were self consistent. Is this check no longer necessary or is it just that it doesn't function for guessed sample names?

intrahost.py
```diff
@@ -580,7 +595,10 @@ def merge_to_vcf(
                         alignmentFile)
 
                 for s in samplesToUse:
-                    for row in util.file.read_tabfile(samp_to_isnv[sampleIDMatch(s)]):
+                    isnv_filepath = samp_to_isnv[sampleIDMatch(s)]
+                    if os.path.getsize(isnv_filepath) == 0:
```
dpark01: Why the explicit check for `os.path.getsize() == 0`? Are there special scenarios where we pass in empty files that semantically mean something we should skip, or is this just a normal situation where vphaser detects no isnvs for that particular file (and thus produces empty output)? `util.file.read_tabfile` should simply produce an empty iterator, so this check seems unnecessary?

intrahost.py
```diff
@@ -460,6 +460,21 @@ def merge_to_vcf(
...
 
+    guessed_samples = []
+    isnvs_to_use = []
+    if not samples:
+        for isnvs_file in isnvs:
+            if os.path.getsize(isnvs_file) > 0:
+                isnvs_to_use.append(isnvs_file)
+                continue
+            for row in util.file.read_tabfile(isnvs_file):
+                guessed_sample_ID = sampleIDMatch(row[0])
```
dpark01: If I read this correctly, this code is unreachable, because anytime the `isnvs_file` is non-empty, the above `if` statement will `continue` and skip this `for` loop, but any time it's empty, the `for` loop won't do anything.

intrahost.py
```diff
@@ -800,7 +817,7 @@ def merge_to_vcf(
                         continue
                     alleleMap = dict((a, i) for i, a in enumerate(alleles))
                     # GT col emitted below
-                    genos = [str(alleleMap.get(consAlleles.get(s), '.')) for s in samples]
+                    genos = [str(alleleMap.get(consAlleles.get(s), '.')) for s in samplesToUse]
```
dpark01: oops, so this was a bug before?

test/unit/test_intrahost.py
```diff
@@ -152,7 +152,7 @@ def test_single_strand_bias_hard_filter(self):
         self.assertEqual(output[0][7:], expected[7:])
 
 
-@unittest.skipIf(tools.is_osx(), "vphaser2 osx binary from bioconda has issues")
+#@unittest.skipIf(tools.is_osx(), "vphaser2 osx binary from bioconda has issues")
```
dpark01: did the osx version get fixed?

test/unit/test_intrahost.py
```diff
@@ -224,8 +224,10 @@ def add_indel(self, sample, chrom, pos, acounts, libinfo=None):
 
     def dump_isnv_tmp_file(self, sample):
         fn = util.file.mkstempfname('.txt')
+        print("self.isnvs[sample]",sample,list(map(str,self.isnvs[sample])))
```
dpark01: are we sure we want to `print` all this or was this leftover from a debugging stage?

tomkinsc: leftover, whoops!

test/unit/test_intrahost.py
tomkinsc: It seems to be? The test passes on OSX and I can get help text when I call it in isolation. Do you recall what the issue was?

intrahost.py
tomkinsc: Yes. :(

intrahost.py
tomkinsc: I'll adapt it for the both cases, where sample names are provided, and when they are not.

intrahost.py
tomkinsc: You're right, this should not be necessary.

intrahost.py
tomkinsc: Thanks, that was sloppy of me. I'll fix this and add a unit test where sample names are guessed so this has some test coverage.

travis/install-wdl.sh
```diff
@@ -17,9 +17,9 @@ cached_fetch_jar_from_github () {
 	ln -s $CACHE_DIR/$_jar_fname $_tool_name.jar
 }
 
-cached_fetch_jar_from_github broadinstitute cromwell womtool 30.2
-cached_fetch_jar_from_github broadinstitute cromwell cromwell 30.2
-cached_fetch_jar_from_github dnanexus dxWDL dxWDL 0.60.2
+cached_fetch_jar_from_github broadinstitute cromwell womtool 32
+cached_fetch_jar_from_github broadinstitute cromwell cromwell 32
+cached_fetch_jar_from_github dnanexus dxWDL dxWDL 0.66.2
```
dpark01: If it doesn't cause much trouble, since you're updating from dxWDL 0.60.2 -> 0.66.2 anyway, want to just try going all the way to 0.68.1 (since some time has elapsed in the course of the development of this branch)?

tomkinsc: Let's see what happens. I'll try updating Cromwell too.

dpark01:
Actually it might be weird if you update Cromwell because then dxWDL and Cromwell will be out of sync (dxWDL 0.68.1 is based on Cromwell 32, I think).

Issue #839, opened 2018-07-06
Non-deterministic failure of test_kraken_multi teardown


dpark01:
Sometimes Travis fails, sometimes it succeeds, on this particular part of the test suite. Probably some race condition as xdist parallelizes the tests.

Mysteries include:
 - Why is the existence of `libgkl_compression2251932993111688424.so` non-deterministic?
 - Why is there a weird number in that filename?
 - Why is `os.unlink` trying to use a compression library to delete files?
 - Will this problem go away by itself? If so, will that make us feel better or worse about it?

```
_______ ERROR at teardown of test_kraken_multi[TestMetagenomicsViralMix] _______
[gw1] linux -- Python 3.6.5 /home/travis/build/broadinstitute/viral-ngs/tools/conda-tools/default/bin/python
def reset():
>       shutil.rmtree(new_tempdir)
conftest.py:64: 
_ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ 
tools/conda-tools/default/lib/python3.6/shutil.py:480: in rmtree
    _rmtree_safe_fd(fd, path, onerror)
tools/conda-tools/default/lib/python3.6/shutil.py:438: in _rmtree_safe_fd
    onerror(os.unlink, fullname, sys.exc_info())
_ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ 
topfd = 11, path = '/tmp/pytest-of-travis/pytest-0/popen-gw1/test-function0'
onerror = <function rmtree.<locals>.onerror at 0x7f004cc5f1e0>
    def _rmtree_safe_fd(topfd, path, onerror):
        names = []
        try:
            names = os.listdir(topfd)
        except OSError as err:
            err.filename = path
            onerror(os.listdir, path, sys.exc_info())
        for name in names:
            fullname = os.path.join(path, name)
            try:
                orig_st = os.stat(name, dir_fd=topfd, follow_symlinks=False)
                mode = orig_st.st_mode
            except OSError:
                mode = 0
            if stat.S_ISDIR(mode):
                try:
                    dirfd = os.open(name, os.O_RDONLY, dir_fd=topfd)
                except OSError:
                    onerror(os.open, fullname, sys.exc_info())
                else:
                    try:
[...]

PR #840, opened 2018-07-06
conftest.py fixes

notestaff:
Changed tmpdir_{session,module,function} to use tempfile.mkdtemp(), and to check if the tempdir still exists before trying to erase it.
In tmpdir_function() switch to using monkeypatch; this fixes a bug
where if TMPDIR was originally unset it will be left set to the
now-deleted tempdir.

### DIFF
--- a/conftest.py
+++ b/conftest.py
@@ -1,11 +1,13 @@
 import operator
 import os
-import pytest
 import shutil
 import sys
 import tempfile
 import time
+import contextlib
+import string
 
+import pytest
 
 def timer():
     if sys.version_info < (3, 3):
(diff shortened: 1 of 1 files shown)


conftest.py
```diff
@@ -29,46 +31,44 @@ def pytest_configure(config):
     reporter = FixtureReporter(config)
     config.pluginmanager.register(reporter, 'fixturereporter')
 
+def _max_fname_len(file_system_path, default_max_len=80):
+    name_max_str = [s for s in os.pathconf_names if s.endswith('_NAME_MAX')]
```
dpark01: _PATH_MAX?

conftest.py
```diff
@@ -29,46 +31,44 @@ def pytest_configure(config):
...
+    if len(name_max_str) == 1:
+        try:
+            return os.pathconf(file_system_path, name_max_str[0])
+        except OSError:
+            pass
+    return default_max_len
+
+def _make_fname_valid(file_system_path, fname, len_margin):
```
dpark01: What's invalid about non-ascii filenames? Is there any reason this needs to be enforced?

conftest.py
```diff
@@ -29,46 +31,44 @@ def pytest_configure(config):
...
+    max_len = _max_fname_len(file_system_path)-len_margin
+    return ''.join((c if c in valid_fname_chars else '_') for c in fname[:max_len])
+
+@contextlib.contextmanager
+def _tmpdir_aux(request, tmpdir_factory, scope, name):
+    basetemp = str(tmpdir_factory.getbasetemp())
+    name = _make_fname_valid(file_system_path=basetemp, fname=name, len_margin=50)
+    tmpdir = tempfile.mkdtemp(dir=basetemp, prefix='test-{}-{}-'.format(scope, name))
+    yield tmpdir
+    if os.path.isdir(tmpdir): shutil.rmtree(tmpdir)
```
dpark01: Don't we still think we should `shutil.rmtree(ignore_errors=True)`? Our error before in #839 was not about the tmpdir not existing, it was about a file within the tmpdir disappearing while rmtree was in the middle of walking the tree.

And if we `ignore_errors=True` then the `if os.path.isdir` becomes unnecessary as a side effect.

conftest.py
tomkinsc: (if it does need to be enforced, we have [`util.file.string_to_file_name()`](https://github.com/broadinstitute/viral-ngs/blob/master/util/file.py#L571-L619))

dpark01: Actually the commented out portions of that function show that we made a point of allowing unicode filenames, we just disallowed certain pathsep or escape characters.

conftest.py
notestaff: this is for a (prefix of) a path component, not full path

notestaff: turning into iterators is for specifying cleanup actions after fixture goes out of scope

PR #844, opened 2018-07-10
allow sequencing_center to be passed as an optional input to demux_launcher

tomkinsc:
With this PR, sequencing_center can be passed as an optional input to the demux_launcher DNAnexus applet; this allows each job spawning a demux (uploader laptops, etc.) to set the sequencing center at upload time

### DIFF
--- a/pipes/WDL/dx-launcher/demux_launcher.yml
+++ b/pipes/WDL/dx-launcher/demux_launcher.yml
@@ -22,7 +22,7 @@ inputSpec:
 - name: sequencing_center
   class: string
   optional: true
-  help: Optional- the name of the sequencing center specified in the demultiplexing output (bam files)
+  help: Optional- the name of the sequencing center to be included in the demultiplexed output (bam files). Please avoid spaces and special characters.
 - name: api_token
   class: file
   optional: true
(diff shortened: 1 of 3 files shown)


pipes/WDL/dx-launcher/demux_launcher.yml
```diff
@@ -19,6 +19,10 @@ inputSpec:
   class: string
   default: /
   help: Output folder in the root project. For this applet, the --destination should be set to / and this input should be used to control the output folder.
+- name: sequencing_center
+  class: string
+  optional: true
+  help: Optional- the name of the sequencing center specified in the demultiplexing output (bam files)
```
dpark01: Note in the help string here that the value should avoid any spaces or special characters? Beats having to escape and support all those scenarios.

tomkinsc: We could process the value passed for sequencing center via `util.file.string_to_file_name()` in illumina.py?

dpark01: I'm not assuming that this yml has access to our python `util.file`. In fact the runSpec looks a bit more bash-like. I'm not worried about the downstream stuff, I'm more worried about your actual `runcmd` just below if someone entered a sequencing center like "My Sequencing Center".

tomkinsc: That makes sense. I suppose I'll escape the input in illumina.py just to be safe, and also mention in the help text to avoid spaces and special characters.

PR #847, opened 2018-07-16
change fuzzy position operators for RHS clipped features

tomkinsc:
this reverses changes from:
 https://github.com/broadinstitute/viral-ngs/commit/1271307b5a5abbe96ad5ff74f1f5913b6f9ce488#diff-1fff82e70c2795767d937ffd93ccc616

### DIFF
--- a/ncbi.py
+++ b/ncbi.py
@@ -107,7 +107,7 @@ def tbl_transfer_common(cmap, ref_tbl, out_tbl, alt_chrlens, oob_clip=False, ign
                                 # less than a codon remains, drop it
                                 return (None, None)
                         start.position = '{}'.format(r)
-                        start.location_operator = '>'
+                        start.location_operator = '<'
                     if end.position == None:
                         # clip the end (left side)
                         end.position = '1'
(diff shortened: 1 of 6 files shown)


dpark01:
[testfiles.zip](https://github.com/broadinstitute/viral-ngs/files/2198197/testfiles.zip)

To reproduce: unpack this zip file into an empty directory, copy in all of the `tbl` files from `viral-ngs/test/input/TestFeatureTransfer/lasv/expected` (from the `ct-feature-clipping-fix` branch), and run `tbl2asn -t authors-nga_lasv.sbt -p . -X C -j "organism=Lassa mammarenavirus mol_type=cRNA"`

dpark01:
Okay according to the fine print in the [documentation](https://www.ncbi.nlm.nih.gov/Sequin/table.html):

> Locations of partial (incomplete) features are indicated with a ">" or "<" next to the number. In this example, the first gene, CDS, and mRNA all begin upstream of the start of the nucleotide sequence. The "<" symbol indicates that they are 5\' partial features. Furthermore, for the protein to translate correctly, the correct reading frame must be indicated with the qualifier "codon_start" on the first CDS. There is no need to indicate the codon_start on complete CDSs, as it is assumed that the translation starts at the first nucleotide of the interval if no codon_start is provided.

So this implies that `<` does *not* mean "to the left of" (ie coordinates that are lower integers than this one). It means "5\' partial feature" (even if it's the right end of the feature). So the original encoding was correct.

PR #853, opened 2018-07-19
add tarball merger

tomkinsc:
add a new utility function that merges separate tarballs into one tarball; data can be piped in and/or out, and the contents can optionally be extracted to disk during the repack


dpark01:
I'll try to look at this more tomorrow... is this pre-requisite for breaking out the tar-repack post-upload step into its own dx applet?

Part of me wonders if we should bite the bullet soon and switch our compression/decompression stuff to blosc. Maybe a separate PR. But we could shed all the binaries for pigz, bzip2, lz4 and just add the python library and gain a few other formats like zstd.

I also wonder whether we could implement a streaming tarcat standalone method that avoids the unpacking to disk by withholding the EOF zero markers for the first `N-1` input streams... boggles my mind that no one has implemented it yet (GNU tar's --concatenate option modifies an input posix file; it doesn't work on pipes like cat does).

BTW if you add a new top-level python file, there's various places it needs to be added including coveralls/py.test invocation scripts, readthedocs/sphinx code, and possibly some other places I'm forgetting at the moment.

tomkinsc:
The intent was to have tar repack functionality as part of viral-ngs, to support either a separate dx applet or repack-capable demux. @mlin now [has a branch](https://github.com/broadinstitute/viral-ngs/compare/mlin-dx-launcher-consolidate) with a WIP yml-specified applet to perform the merge operation, so this may be a bit redundant, but including it as part of demux would save an extra download of the packed tarball onto the demux instance.

Switching (de)compression to blosc should probably be a separate PR since we call the various binaries in various places. Something I'm not sure of is whether some of the magic of blosc relies on compile-time optimizations targeting the instruction set extensions/CPU cache available, or if it compiles to include multiple code paths. We may lose some optimizations if it is installed from a source like conda. The best thing is to probably just try it and see if real-world performance is improved over what we have now.

The function included in this PR has a few different code paths. By default it acts as a streaming tarcat and file data never touches the disk; it is read from the untarred stream, buffered by python's [tarfile](https://github.com/python/cpython/blob/master/Lib/tarfile.py#L234), and written directly to the tarfile output stream (`TarFile.extractfile()` can return a stream file object). That's for the simple case of a repack. A flag, `--extractToDiskPath`, is available for cases where we want to read the input tarballs once and do two things: extract the files to disk, and repack them. That's intended for use on a demux instance. In that case, the default read function is wrapped by a custom class that provides file-like read,  and writes out bytes to a file before returning them to be written to the output stream (the wrapper sets file attributes on close). The sequential nature of tarfiles is helpful here, and it's easy because the stream object of Python's `tarfile` pads 512-byte blocks and 20-block records as necessary when writing, so we don't need to do the bookkeeping. In something of a hacky shortcut, non-file tarball members (directories, symlinks, etc.) are written to disk and then added from disk in the case where we are writing the output to disk anyway (so we do not need a complicated `read()` diverter for non-file data to create directories, etc. on disk.) It adds a disk roundtrip, but the members are small so it's quick, and it avoids duplicating the (internal, private) logic within Python's `tarfile` to inspect and create the various member types with appropriate attributes. For the simple case where extracted output on disk is not desired, the small members are streamed along with everything else.

I had the same thought as you about concatenating tarballs in the old-school tape drive way, by stripping the final two 512-byte zero blocks off each (`N-1`) tarball. That would be fairly straightforward if the files were not compressed, since we could know the size in advance and pipe to `dd` (or even `head` with a negative offset) to truncate the tarfile content. Compression complicates it because the original size of the tarball can only be reliably determined after it is decompressed. The gz format stores the length at the end of the file as four bytes encoding an unsigned `long`, making it impossible to get the uncompressed size for streamed files without a second pass, or for files >4GB in size. We can use `pigz -l` to get the decompressed size, but it seems to rely on the same `long` trick so it fails for files >4GB and needs to read the entire file for streams.

We can, of course, simply `cat` the `tar.gz` files together if the resulting file is always read with `tar -i` to ignore zero blocks, but that seems dangerous since the consumer would have to know to apply `-i` or else the output would be truncated. The function in this PR tries to be a reasonable compromise that will work with streamed data as input and/or output.

Something I'm not sure of yet is how dxWDL handles an `Array[File]+` input with `parameter_meta` set to `stream`. Does it allow the selection of multiple files and stream all of them?

dpark01:
Fascinating. A few thoughts.

I think some of the speed magic of blosc isn't really about blosc, but the underlying algorithms that it implements.. a lot of the newer ones are computationally simple enough to saturate the memory bandwidth of any machine. Agreed though that we should keep it separate.

dxWDL does not currently handle `Array[File]` streaming at all. Ohad had said there would be some implementation difficulty around that.. personally I think a non-localized / NIO approach that WDL 1.0 seems to advocate would require some greater rethinking of this anyway.

Anyway the short of it is that you're saying that the default execution behavior of this is to be able to repack tarballs while both streaming the inputs and outputs and avoiding any disk I/O. That's great, and quite ideal actually! That should significantly speed up repacking of large tarballs (avoiding the disk I/O). In my mind, a hacked together tarcat would have always been streaming: instead of calculating file sizes and offsets, I would've just introduced a 1kB buffer and dropped any EOF blocks from the stream except for the last one. But better to use an established python library.

Of the different code paths, how many of them are tested? Can you test both the `--extractToDiskPath` and default diskless behavior?

The advantage to @mlin 's standalone applet for repack prior to demux is that the repacked tarball always emits first, regardless of whether demux fails, and the standalone applet doesn't need to waste 5 mins pulling the viral-ngs docker image. The downside to it though is that it is hitting disk. Since DNAnexus seems to rely exclusively on non-EBS-backed EC2 instances, the speed hit isn't too big of a deal, but it does mean that the instance sizes have to scale based on the data size (since local disk has to be big enough for the uncompressed flowcell, which wouldn't be true with your repacker). I guess the best of both worlds would be to separate out your python `tarfile` based repacker into a minimalist docker that pulls faster...

tomkinsc:
Ok, so it sounds like we'll be using @mlin's implementation for several reasons. I can add additional tests for this PR if you think it is useful enough to be included.

dpark01:
Yeah I think we'd want this anyway, it does look quite useful

util/file.py
```diff
@@ -830,3 +831,129 @@ def join_paired_fastq(input_fastqs, output_format='fastq', num_n=None):
...
+            return buf
+
+    if extract_to_disk_path and not os.path.isdir(extract_to_disk_path):
+        os.makedirs(extract_to_disk_path)
+
+    with open(out_compressed_tarball, "w") as outfile:
+        if out_compressed_tarball != "-":
+            out_compress_ps = subprocess.Popen(choose_compressor(out_compressed_tarball)["compress_cmd"], stdout=None if out_compressed_tarball == "-" else outfile, stdin=subprocess.PIPE)
+        else:
+            assert out_compressed_tarball != '-' or pipe_hint, "cannot autodetect compression for stdout output unless pipeHint provided"
```
yesimon: Isn't `out_compressed_tarball != '-'` just checked?

util/file.py
```diff
@@ -830,3 +831,129 @@ def join_paired_fastq(input_fastqs, output_format='fastq', num_n=None):
...
+        return return_obj
+
+    def create_containing_dirs(path):
+        sub_path = os.path.dirname(path)
+        if not os.path.exists(sub_path) and sub_path != path:
+            create_containing_dirs(sub_path)
+        if not os.path.exists(path) and len(path):
+            os.mkdir(path)
+
+    class FileDiverter(object):
```
yesimon: I'm not a huge fan of class inside the function. This function is already very long

util/file.py
```diff
@@ -830,3 +831,129 @@ def join_paired_fastq(input_fastqs, output_format='fastq', num_n=None):
...
+        """
+        def __init__(self, fileinfo, fileobj, written_mirror_file=None, extract_numeric_owner=False):
+            self.written_mirror_file = open(written_mirror_file,"wb")
+            self.fileinfo = fileinfo
+            self.fileobj = fileobj
+            self.extract_numeric_owner = extract_numeric_owner
+
+        def __del__(self):
+            self.written_mirror_file.flush()
+            self.written_mirror_file.close()
```
yesimon: `close` calls `flush`

util/file.py
```diff
@@ -830,3 +831,129 @@ def join_paired_fastq(input_fastqs, output_format='fastq', num_n=None):
...
+            if in_compressed_tarball != "-":
+                pigz_ps = subprocess.Popen(choose_compressor(in_compressed_tarball)["decompress_cmd"] + [in_compressed_tarball], stdout=subprocess.PIPE)
+            else:
+                assert in_compressed_tarball != '-' or pipe_hint, "cannot autodetect compression on stdin input unless pipeHint provided"
+                pigz_ps = subprocess.Popen(choose_compressor(pipe_hint)["decompress_cmd"] + [in_compressed_tarball], stdout=subprocess.PIPE, stdin=sys.stdin)
+            tar_in = tarfile.open(fileobj=pigz_ps.stdout, mode="r|", ignore_zeros=True)
+
+            fileinfo = tar_in.next()
+            while fileinfo is not None:
+                if extract_to_disk_path:
```
yesimon: Break out into subfunction?

util/file.py
```diff
@@ -830,3 +831,129 @@ def join_paired_fastq(input_fastqs, output_format='fastq', num_n=None):
...
+                            outfile = tar_in.extract(fileinfo, path=extract_to_disk_path)
+                            with pushd_popd(extract_to_disk_path):
+                                tar_out.add(fileinfo.name)
+                        
+                    else:
+                        # alternative implementation:
+                        outfile = tar_in.extract(fileinfo, path=extract_to_disk_path)
+                        with pushd_popd(extract_to_disk_path):
+                            tar_out.add(fileinfo.name)
+                else:
```
yesimon: Comment because the if branch is a bit far

util/file.py
```diff
@@ -830,3 +831,129 @@ def join_paired_fastq(input_fastqs, output_format='fastq', num_n=None):
...
+                fileinfo = tar_in.next()
+            pigz_ps.stdout.flush()
+            pigz_ps.wait()
+            tar_in.close()
+            if pigz_ps.poll():
+                raise subprocess.CalledProcessError(pigz_ps.returncode, "Call error %s" % pigz_ps.returncode)
+
+        tar_out.close()
+        out_compress_ps.stdin.flush()
+        out_compress_ps.stdin.close()
```
yesimon: `close` calls `flush`

util/file.py
```diff
@@ -830,3 +831,129 @@ def join_paired_fastq(input_fastqs, output_format='fastq', num_n=None):
...
+        out_compress_ps.wait()
+        if out_compress_ps.poll():
```
yesimon: Should this be `if ps.returncode != 0`?

util/file.py
```diff
@@ -830,3 +831,129 @@ def join_paired_fastq(input_fastqs, output_format='fastq', num_n=None):
...
+            fileinfo = tar_in.next()
+            while fileinfo is not None:
+                if extract_to_disk_path:
+                    target_path = os.path.normpath(os.path.join(extract_to_disk_path, fileinfo.name).rstrip("/"))
+                    containing_path = os.path.dirname(target_path)
+                    create_containing_dirs(containing_path)
+
+                    if avoid_disk_roundtrip:
+                        fileobj = tar_in.extractfile(fileinfo)
+                        #tar_out.addfile(fileinfo)
```
yesimon: delete

util/file.py
```diff
@@ -830,3 +831,129 @@ def join_paired_fastq(input_fastqs, output_format='fastq', num_n=None):
...
+
+    if extract_to_disk_path and not os.path.isdir(extract_to_disk_path):
+        os.makedirs(extract_to_disk_path)
+
+    with open(out_compressed_tarball, "w") as outfile:
+        if out_compressed_tarball != "-":
+            out_compress_ps = subprocess.Popen(choose_compressor(out_compressed_tarball)["compress_cmd"], stdout=None if out_compressed_tarball == "-" else outfile, stdin=subprocess.PIPE)
+        else:
+            assert out_compressed_tarball != '-' or pipe_hint, "cannot autodetect compression for stdout output unless pipeHint provided"
+            out_compress_ps = subprocess.Popen(choose_compressor(pipe_hint)["compress_cmd"], stdout=None if out_compressed_tarball == "-" else outfile, stdin=subprocess.PIPE)
```
yesimon: DRY this because only the `choose_compressor` call is different

file_utils.py
```diff
@@ -0,0 +1,57 @@
...
+    parser.add_argument(
+        'in_tarballs', nargs='+',
+        help=('input tarballs (*.tar.gz|*.tar.lz4|*.tar.bz2)')
+    )
+    parser.add_argument('--extractToDiskPath',
+                        dest="extract_to_disk_path",
+                        help='If specified, the tar contents will also be extracted to a local directory.')
+    parser.add_argument('--pipeHint',
+                        dest="pipe_hint",
+                        default=".gz",
```
yesimon: Maybe a better UI is not to have the leading `.`

file_utils.py
```diff
@@ -0,0 +1,57 @@
...
+        data can be piped in and/or out
+    '''
+    util.file.repack_tarballs(out_tarball, in_tarballs, threads=threads, extract_to_disk_path=extract_to_disk_path, pipe_hint=pipe_hint)
+    return 0
+def parser_merge_tarballs(parser=argparse.ArgumentParser()):
+    parser.add_argument(
+        'out_tarball', 
+        help='''output tarball (*.tar.gz|*.tar.lz4|*.tar.bz2|-) 
+        Note: if "-" is used, a gzip-compressed tarball 
+        will be written to stdout''')
```
yesimon: Add some comment about how output compression is inferred by the file extension.

test/unit/test_file_utils.py
```diff
@@ -0,0 +1,88 @@
...
+        tb = tarfile.open(out_tarball_file)
+        tb.extractall(path=temp_dir)
+
+        for i in range(len(raw_files)):
+            inf = os.path.join(self.input_dir,"raw-input",raw_files[i])
+            outf = os.path.join(temp_dir,raw_files[i])
+
+            assert_equal_contents(self, inf, outf)
+
+    def test_merge_with_extract(self):
```
yesimon: Test `avoid_roundtrip = False` as well?

util/file.py
dpark01: Also may want to avoid using [assert](https://docs.python.org/3/reference/simple_stmts.html#the-assert-statement) outside of unit test code.

util/file.py
dpark01: True, but it works though... `Popen.poll()` returns either the returncode (if it's done) or `None` (if it's not done, but it's guaranteed to be done after the `.wait()` anyway). I think we have a few other places like this throughout our code that could be converted for clarity, but would likely not change behavior.

file_utils.py
dpark01: To be clear, the `pipeHint` used in other places is actually meant to be any kind of file path or URI even, and the logic that matches on it isn't looking for strings that match `.gz`, but rather it's doing `string.endswith()` calls on it. The idea is that you could just supply a hint or you could be lazy and supply the whole filename, bucket path, or whatever.

util/file.py
tomkinsc: Yes, I agree, feels ugly but it's also not used beyond here and it relies on the attributes possessing TarInfo attributes so its function is quite internal to this function.

tomkinsc:
Thanks for the code review!

PR #854, opened 2018-07-19
Adding commands for working with kmer sets using the KMC tool.

notestaff:
Adding commands for working with kmer sets using the KMC tool.  Some planned uses: recovering currently depleted but relevant reads; evaluating optimality of assemblies and finding potentially fillable gaps; generating fine-grained reports on the occurrence in the known reference sequences of raw read kmers.


dpark01:
If I don't get a good look at it before the weekend, let's chat early next week

notestaff:
Periodically, coverall erroneously reports big drops in coverage.  Is there a known workaround?

dpark01:
I think the exact reason for that has varied over time... in the past we used to do more test skipping in certain build scenarios (which I think we don't do as much now.. though there is still some, for example, most of the tool install tests are skipped unless it's a tagged release). Sometimes it might be because the PR's merge structure is funny and coveralls might not be comparing to the right comparison point (but this particular instance looks like a clean merge). In any case, this is why I've changed the coveralls test to be a non-blocking hook in github--it doesn't prevent a merge of a PR, but it's a useful data point to have.

notestaff:
It's definitely a useful data point in general.   In this specific case, the reason I think the data is incorrect is that a tiny change ( https://github.com/broadinstitute/viral-ngs/pull/854/commits/58ce922b24a69dc75b3b7c3de1c0ec185fc6c7a7 ) caused a 41% drop in coverage.

yesimon:
I think it's because the long and short tests are run separately and have different amounts of coverage. If only one side is updated and rerun it overwrites the old coveralls report and I don't think it has the ability to merge reports

notestaff:
But pushing a commit runs all parts of the build matrix, and they all have --cov-append, no?

PR #864, opened 2018-07-23
add workflows for isnvs_merge_to_vcf

tomkinsc:


dpark01:
Hm.. just realized looking at this: what’s the difference between the two tasks besides the boolean input? If there’s no difference perhaps they should be consolidated to one task? The Boolean input could even be optional if we want it to be.

Separately, should we rename all our tasks/xxx.wdl import files to tasks_xxx.wdl ?

tomkinsc:
Nice catch; the two had carried over from the Snakemake rules. I'll consolidate them.

We could rename the task files to avoid collisions. I can submit that as a separate PR.

pipes/WDL/workflows/tasks/tasks_intrahost.wdl
```diff
@@ -92,18 +66,16 @@ task isnvs_vcf_filtered {
         --isnvs ${sep=' ' vphaser2Calls} \
         --alignments ${sep=' ' perSegmentMultiAlignments} \
         --strip_chr_version \
-        ${'--naive_filter' + naiveFilter} \
-        --parse_accession
-
+        $naive_filter \
+        --parse_accession && \
```
dpark01: Actually, instead of stringing everything in the WDL `command` section together with `&&`, I've found it easier just to `set -ex -o pipefail` in any WDL task that's more than a one-liner.

tomkinsc:
`set -ex -o pipefail` works within WDL bash, that's great. 
I'm still kicking the tires on this one. I'm getting VCFs with headers but not rows, and still checking whether the cause is the data (v-phaser does emit some variants) or something with snpEff/our process.

pipes/WDL/workflows/tasks/tasks_intrahost.wdl
```diff
@@ -29,61 +29,37 @@ task isnvs_per_sample {
...
-    intrahost.py merge_to_vcf  \
-        ${reference_fasta} \
-        isnvs.vcf.gz \
-        $SAMPLES \
-        --isnvs ${sep=' ' vphaser2Calls} \
-        --alignments ${sep=' ' perSegmentMultiAlignments} \
-        --strip_chr_version \
-        --parse_accession
+    if [[ "${naiveFilter}" == "true" ]]; then
+      naive_filter="--naive_filter"
```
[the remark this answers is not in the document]
tomkinsc: 👍 It's even in the docs! 
https://github.com/openwdl/wdl/blob/master/versions/draft-2/SPEC.md#true-and-false

docker/rundocker.sh
```diff
@@ -3,7 +3,7 @@
 # A wrapper script to run viral-ngs docker images
 # The following paths have to be modified according to end-user environment
 NOVOALIGN_PATH="/opt/novocraft" # Directory where novoalign.lic license file can befound
-GATK_PATH="/opt/GenomeAnalysisTK-3.6" # Directory where the correct GATK jar file can be found
+GATK_PATH="/opt/GenomeAnalysisTK-3.8" # Directory where the correct GATK jar file can be found
```
notestaff: @tomkinsc add this to https://github.com/broadinstitute/viral-ngs/blob/master/DEVELOPMENT_NOTES.md#upgrading-gatk
?

tomkinsc: done, thanks

dpark01: Has anyone used rundocker.sh recently? It doesn’t get used for anything more pipelined, but for ad hoc command line work, is it useful beyond what you get from the usual “docker run -it —rm”?

tomkinsc: I haven't used it recently.

tools/snpeff.py
```diff
@@ -59,7 +59,7 @@ def execute(self, command, args, JVMmemory=None, stdin=None, stdout=None):    #
             ] + args
 
         _log.debug(' '.join(tool_cmd))
-        return util.misc.run_and_print(tool_cmd, stdin=stdin, buffered=True, silent=("databases" in command), check=True)
+        return util.misc.run_and_print(tool_cmd, stdin=stdin, stderr=stderr, buffered=True, silent=("databases" in command), check=True)
```
dpark01: Any reason we can just revert this to subprocess.check_call?

tomkinsc: We probably should revert. Let's save it for a separate PR? This one is already larger than it should be.

PR #866, opened 2018-07-24
testing: fixed an issue with the handling  of tempdirs

notestaff:
Fixed an issue with the handling of temporary directories during testing, caused by an interaction of a legacy unittest-based class with the newer pytest-based fixtures.

### DIFF
--- a/test/__init__.py
+++ b/test/__init__.py
@@ -80,22 +80,9 @@ def assert_md5_equal_to_line_in_file(testCase, filename, checksum_file, msg=None
 
     testCase.assertEqual(hash_md5.hexdigest(), expected_checksum, msg=msg)
 
+@pytest.mark.usefixtures('tmpdir_class')
 class TestCaseWithTmp(unittest.TestCase):
     'Base class for tests that use tempDir'
-
-    @classmethod
-    def setUpClass(cls):
-        cls._class_tempdir = util.file.set_tmp_dir(cls.__name__)
-
-    def setUp(self):
-        util.file.set_tmp_dir(type(self).__name__)
-
-    @classmethod
-    def tearDownClass(cls):
-        util.file.destroy_tmp_dir(cls._class_tempdir)
-
-    def tearDown(self):
-        util.file.destroy_tmp_dir()
 
     def assertEqualContents(self, f1, f2):
         assert_equal_contents(self, f1, f2)
--- a/test/__init__.py
+++ b/test/__init__.py
@@ -12,6 +12,7 @@ import logging
 
 # third-party
 import Bio.SeqIO
+import pytest
 
 # intra-project
 import util.file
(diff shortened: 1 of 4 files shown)


notestaff:
Btw, integration/test_ncbi.py uses class-scope tempdirs rather than function-scope ones.   Is it necessary for tests in each test class there to share a tempdir?

notestaff:
@tomkinsc calling tempfile.gettempdir() still works:
https://github.com/broadinstitute/viral-ngs/blob/master/test/integration/test_ncbi.py#L38

tomkinsc:
The tempdirs would ideally be function-scoped. With the changes from this PR applied that appears to be true for `tempfile.gettempdir()`?

notestaff:
That's true.  I just want to check that it was not important for them to be class-scoped, as they were originally.

test/integration/test_ncbi.py
```diff
@@ -20,7 +20,7 @@
 log = logging.getLogger(__name__)
 
 
-skip_test = True
+skip_test = False
```
tomkinsc: These tests were mostly disabled because instability in the NCBI API due to load frequently causes the tests to fail, even with the retries in `genbank.py`. Enabling them was mostly to illustrate the tempdir issue, so we can probably switch them off again.

test/__init__.py
```diff
@@ -80,23 +81,10 @@ def assert_md5_equal_to_line_in_file(testCase, filename, checksum_file, msg=None
 
     testCase.assertEqual(hash_md5.hexdigest(), expected_checksum, msg=msg)
 
+@pytest.mark.usefixtures('tmpdir_class')
 class TestCaseWithTmp(unittest.TestCase):
     'Base class for tests that use tempDir'
 
-    @classmethod
```
tomkinsc: This removal fixes the issue I was having, but does it break `unittest`?

notestaff: @tomkinsc What mode of breakage are you thinking about?   Can you maybe add a test to test for it?   Current test suite doesn't seem to break.

tomkinsc: Incompatibility when testing under the built-in `unittest` module, which I suppose we don't need to worry about if we're going all-in on pytest.

notestaff: @tomkinsc I've been adding tests to the test suite that are pytest-based only, which would not be run if pytest is not used.  Is there a goal of keeping the test suite runnable without pytest?

tomkinsc: I don't think we've talked about it(?). @dpark01 @yesimon @haydenm, what are your thoughts?

yesimon: I'm for moving to pytest

dpark01: Originally I think we had it all set up to be backwards compatible with `unittest`. But I think we've already crossed the bridge to being pytest-dependent. A fair amount of the metagenomics tests are not `unittest` compatible. Anyway I wouldn't worry about it at this point...

PR #867, opened 2018-07-24
fixes related to GATK bump to 3.8

notestaff:
Removed GATK command-line args deprecated since 3.7 .  Changed default path to GATK in pipes/config.yaml to point to a 3.8 version.


pipes/config.yaml
```diff
@@ -87,7 +87,7 @@ env_vars:
   # The directory path to the location of the GATK jar file.
   # It must be explicitly given since GATK has to be licensed and downloaded
   # manually out of band.
-  GATK_PATH: "/humgen/gsa-hpprojects/GATK/bin/GenomeAnalysisTK-3.6-0-g89b7209"
+  GATK_PATH: "/idi/sabeti-scratch/shared-resources/software/gatk/GenomeAnalysisTK-3.8-0-ge9d806836"
```
tomkinsc: Somewhat annoying there isn't an extracted directory for v3.8 within `/humgen/gsa-hpprojects/GATK/bin`. Should we ping someone upstairs?

tomkinsc:
Thanks for catching these!

notestaff:
@tomkinsc I wonder why the standard travis-ci testing didn't catch these?

tomkinsc:
The easy-deploy script is no longer tested; I think it was taking too long. Otherwise, [we set](https://github.com/broadinstitute/viral-ngs/blob/master/.travis.yml#L12) the `GATK_PATH` on Travis.

dpark01:
But the call to unified genotyper... that should've failed. Oh no, I know why... @tomkinsc when you updated GATK versions, did you also update the encrypted Travis tarball? Or is Travis still testing 3.6?

dpark01:
Actually instead of pulling the GATK jar from an encrypted Travis secret, I wonder if we can just pull it now from the no-longer-auth-protected GATK public URL... at least for Travis...

tomkinsc:
Shoot, I haven't updated it yet.

dpark01:
Yeah so... since the GATK team no longer protects their 3.x download URLs (and in fact, it looks like the gatk-protected github repo is now public?), we could replace [this section of our install script](https://github.com/broadinstitute/viral-ngs/blob/master/travis/install-gatk.sh#L14) with just `wget https://software.broadinstitute.org/gatk/download/auth?package=GATK-archive&version=3.8-1-0-gf15c1c3ef` (which, despite the "auth" in the URI, doesn't actually auth anymore).

notestaff:
@tomkinsc   Is https://github.com/broadinstitute/viral-ngs-deploy now obsolete?

dpark01:
Yeah we don't use that repo anymore... not since last fall.

tomkinsc:
@notestaff: Could you incorporate the Travis GATK `wget` call into this PR?

notestaff:
Maybe, delete the viral-ngs-deploy repo, to avoid confusion?   Or check in a final commit pointing to the new location.  @tomkinsc @dpark01

dpark01:
So actually... it looks like our travis setup no longer does anything with novoalign (we just test on the free/public version) so if we can get rid of the encrypted secret (like make the changes I suggested above and then also delete the three instances of [BUNDLE_SECRET](https://github.com/broadinstitute/viral-ngs/blob/master/.travis.yml#L60) from .travis.yml) then people can actually run our python test suites on forks of our repo. Which would be nice.

PR #868, opened 2018-07-25
demux changes to support novaseq

tomkinsc:
autoscale some novoseq demux params, allowing override by WDL args, decrease the demux memory footprint, expose a few more demux params, remove `--force_gc=false` from the options viral-ngs passes to picard by default


dpark01:
Ok if the host is killing the process because it's exceeding the available system RAM, but Java XmX is set to 90% RAM, my guess is that we should also be keeping a careful eye on the memory footprint of [count_and_sort_barcodes](https://github.com/broadinstitute/viral-ngs/blob/master/illumina.py#L401). For better and for worse, Java is quite predictable in its RAM usage, the JVM generally uses exactly the amount proscribed by XmX regardless of what it's doing inside the JVM. The way we invoke illumina_demux, we're kicking off count_and_sort_barcodes in the background, which has a subprocess count equal to the tile count and maintains an in-memory map of observed barcodes.

Maybe there's a way to profile things to verify if my guess is correct, but if this is actually the memory problem we're encountering, there's two ways we can approach this. We can either avoid kicking it to the background and run it sequentially against Picard (which, in the case of smaller flowcells, I think doubled the runtime, but I think that was before you parallelized & optimized the function). Or we can try to keep the memory footprint of `count_and_sort_barcodes` to be sublinear with input size. One approach would be to utilize an indexed sqlite (on-scratch-disk) table for counting observed barcodes instead of a python map/dict. Again, assuming this is the source of the problem.

Orthogonally to that, I agree that optimizing our in-JVM memory usage in the NovaSeq context (with Picard GC's and messing with Picard thread counts and memory settings) is also a good thing, but those OOM errors should manifest differently. I also lean towards keeping whatever parameter tuning we do in that Picard space consistent across flowcell types unless there's a drastic change from what we normally do (or if it really negatively impacts performance on smaller flowcells)--ie, I don't think we need to always autodetect and custom fit params around the NovaSeq scenario unless they're actually bad for the smaller scenarios.

tomkinsc:
There's certainly room to optimize `count_and_sort_barcodes`, but at least from [this failure](https://platform.dnanexus.com/projects/FJ4PZ2j0VkG0Vz2Z74bX0x7Q/monitor/analysis/FJFk91Q0VkG54Q0Z6BVzQ93K), it looks like we're hitting the OOM error more than an hour and a half after the `count_and_sort_barcodes` sub-job finishes (search for ["writing output"](https://github.com/broadinstitute/viral-ngs/blob/master/illumina.py#L431) in the log), so I don't think it is entirely to blame. SQLite would be one option, though [`shelve`](https://docs.python.org/3.6/library/shelve.html) could be another and would not require an external dependency.

The motivation for tuning (some of) the params for NovaSeq runs within the demux task was the same as selecting params in the demux_launcher applet, for cases where the user wishes to (re)invoke demultiplexing on already uploaded/consolidated data, with sane defaults for each. We certainly need a sane default for `--minimum_base_quality` due to RTA3's quirky four-value Q-score output. Other than that, tuning the RAM and thread count is probably enough if we leave all other Picard settings the same across flowcell types.

PR #877, opened 2018-08-07
misc small fixes

notestaff:
Misc small fixes.   Factoring out parts of #854 not specific to kmer operations.

* added gitignore for emacs backup files
* added printing of defaults to help msgs. fixed log message that always reported an exception when keeping tmp files, even when there was none. added parse_cmd() and run_cmd() utils to cmd.py .
* added developer docs on adding top-level scripts
* fixed comments; added uncompressed_file_type() to util/file.py; improved preservation of tmp files on request
* added misc utils: as_type, subdict, chk
* reduced code duplication in .travis.yml
* added pytest option to mark tests as slow and run them only when --runslow is specified


util/cmd.py
```diff
@@ -257,4 +260,28 @@ def check_input(condition, error_msg):
...
+    assert inspect.ismodule(module)
+    parser_fn = dict(getattr(module, '__commands__'))[cmd]
+    return parser_fn(argparse.ArgumentParser()).parse_args(list(map(str, args)))
+
+CmdRunInfo = collections.namedtuple('CmdRunInfo', ['result', 'args_parsed'])
+
+def run_cmd(module, cmd, args):
+    """Run command after parsing its arguments with the command's parser.
     
+    Args:
```
tomkinsc: What led to these changes, or what do they fix?

.travis.yml
```diff
@@ -11,6 +11,7 @@ env:
   - PIP_DIR="$HOME/virtualenv"
   - GATK_PATH="$CACHE_DIR"
   - PYTHONIOENCODING=UTF8
+  - PYTEST_ADDOPTS="-rsxX -n 2 --durations=25 --fixture-durations=10 --junit-xml=pytest.xml --cov-report= --cov broad_utils --cov illumina --cov assembly --cov interhost --cov intrahost --cov metagenomics --cov ncbi --cov read_utils --cov reports --cov taxon_filter --cov tools --cov util --cov file_utils"
```
dpark01: I had moved these travis env variables to the parts of the build matrix that actually use py.test... there are a bunch of other points in the build matrix that have nothing to do with pytest.

util/cmd.py
notestaff: These are useful for testing, if you want to run a command but also see the results of parsing its arguments, e.g. https://github.com/broadinstitute/viral-ngs/pull/854/files#diff-771ac3c0ea113ca9a6634523f4fb6893R273
Also calling run_cmd() is clearer than the current pattern of calling the parser manually then calling its func_main .

.travis.yml
notestaff: Right; on the other hand, there are parts that don't use GATK_PATH either.   There's a downside to having copies of the same complex line in several places -- the copies have to be kept in sync.
But if having PYTEST_ADDOPTS defined in jobs that don't use it might cause problems, I'll revert this change.

Issue #874, opened 2018-07-31
add length threshold to metagenomics.py::assemble_spades


tomkinsc:
Spades produces many short contigs. It may be helpful to have an optional parameter for `metagenomics.py::assemble_spades`, perhaps `--minContigLength`, to only write contigs greater in length than a specified value.

dpark01:
Closing as part of monorepo migration. This legacy repository is being rebuilt. If this issue is still relevant, please re-open it in the appropriate repository (viral-core, viral-classify, viral-assemble, viral-phylo, or viral-pipelines).

PR #889, opened 2018-10-18
assembly.assemble_spades: added option to drop contigs shorter than given length.

notestaff:
assembly.assemble_spades: added option to drop contigs shorter than given length (for #874)


notestaff:
travis-ci seems in odd state -- usually there's a pointer to the build in progress.   anyone seen this before / know the fix?

tomkinsc:
From their status page, it looks like there are problems between TravisCI and GitHub:
https://www.traviscistatus.com/

PR #897, opened 2018-11-09
respec snakemake per-rule mem requirement: s/mem/mem_mb/g

tomkinsc:
For Snakemake, `mem_mb` is now the preferred way to specify rule-specific memory; this param is recognized by certain execution engines such as kubernetes. See: https://snakemake.readthedocs.io/en/v5.3.0/executable.html#kubernetes

This PR updates the Snakemake rules files as well as the cluster-submitter scripts


pipes/run-pipe_local.sh
```diff
@@ -0,0 +1,50 @@
+#!/bin/bash
+# Wrappers around Snakemake for use on the Broad UGER cluster
```
dpark01: Can we move this to `pipes/Broad_UGER/` instead of `pipes/`? Or is the comment here on line 2 incorrect--is this actually for running on a single VM, not on UGER?

tomkinsc: Sorry about that; the comment was incorrect. It is indeed for invoking Snakemake on a single instance with similar params. I'll fix the comment.

PR #900, opened 2018-12-04
Fix issues with Trinity

notestaff:
Fix some quirks of Trinity, that were breaking the build.  Trinity needs env var 'OSTYPE' to be set, and the output of 'java -version' to start with the java version.


dpark01:
Thanks for fixing and merging this quickly!

Out of curiosity, what changed? (or how was this ever working before)

PR #906, opened 2018-12-21
conda fixes related to changed bioconda guidelines

notestaff:
Updates to reflect current bioconda guidelines:

bioconda channel priorities now put conda-forge before bioconda, and do not use r: https://bioconda.github.io/#set-up-channels
bioconda now uses openjdk: https://bioconda.github.io/guidelines.html#java

Also updated requirements-minimal.txt to match versions in requirements-conda.txt .

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -38,7 +38,7 @@ jobs:
   include:
 
     - language: java
-      jdk: oraclejdk8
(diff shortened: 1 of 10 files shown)


notestaff:
I did clear all caches before building.  The coveralls failure isn’ a blocker here?

PR #909, opened 2019-01-04
update PyYAML to fix CVE-2017-18342

tomkinsc:
CVE-2017-18342
Vulnerable versions: < 4.2b1
Patched version: 4.2b1
In PyYAML before 4.1, the yaml.load() API could execute arbitrary code. In other words, yaml.safe_load is not used.


tomkinsc:
So this is failing because the old version is still the current release.
pyyaml >3.13 is not yet available anywhere in non-beta form, and release of 4.2 was rolled back. See: https://github.com/yaml/pyyaml/issues/193
A 4.x package for PyYaml is notably absent from anaconda/conda-forge which includes only finalized versions.

Issue #911, opened 2019-01-17
everything is single threaded now?


dpark01:
When executing command line viral-ngs in google cloud, either within a conda environment on a manually created VM, or within dsub/PAPIv2, I'm noticing that `--threads 1` is being passed to pretty much every subprocess call.

I note two possible bugs here:
1. Should [this](https://github.com/broadinstitute/viral-ngs/blob/master/util/misc.py#L422) say `cfs_period_us` instead of `cfs_quota_us`?
2. In my environments, `cfs_quota_us` is reporting `-1`, perhaps using that value to denote unlimited quota. This means `available_cpu_count()` is always reporting `1` instead of the number of processors available on the machine. [This line](https://github.com/broadinstitute/viral-ngs/blob/master/util/misc.py#L424) should change to treat `-1` as infinity.

tomkinsc:
Can we close this now that #912 has been merged?

dpark01:
Yes, similar to the other PR, I think @notestaff 's description of PR #912 needed to say "fixes" or "closes" #911 before it merged, in order for merging to close this issue.

PR #912, opened 2019-01-18
fix util.misc.available_cpu_count()

notestaff:
Fixes util.misc.available_cpu_count() .  Also adds some generally useful pytest utils.

### DIFF
--- a/conftest.py
+++ b/conftest.py
@@ -6,7 +6,10 @@ import tempfile
 import time
 import contextlib
 import string
+import inspect
+import functools
 
+import util.misc
 import util.file
 
 import pytest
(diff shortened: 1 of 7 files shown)


conftest.py
```diff
@@ -83,6 +87,44 @@ def tmpdir_function(request, tmpdir_class, monkeypatch):
         monkeypatch.setenv('TMPDIR', tmpdir)
         yield tmpdir
 
+@pytest.fixture
```
yesimon: I don't think this should be a fixture and contextmanager. The pytest philosophy doesn't want to use `with mock(obj) as m...` lines in the test cases because it breaks up the flow. https://pypi.org/project/pytest-mock/. For your actual test on the CPU counts the simple `mock.return_value = '1'` would suffice. You can perhaps monkeypatch the `MockFixture` module to add this functionality if necessary.

notestaff: The available_cpu_count tests needs to monkeypatch different return values of slurp_file for different arguments.  Also, a contextmanager is needed if you want to test more than one thing within one test function.  monkeypatch with context managers is built into pytest, so pytest philosophy does allow for that.

notestaff:
The situation in #911, where cfs_quota_us is reporting -1, doesn’t exist on travis, so the moneypatching simulates it.

dpark01:
Hm, yes worth mocking all the possible values, I'm just not used to this way of mocking it. Let's just go with this.

I do see `-1` from that cgroup file in the usual PAPIv2 setup, my guess is that `test-docker.sh` is actually getting `-1` if `--cpus` was not passed to `docker run` (but then you wouldn't know what the expected return value of the function would be).

PR #915, opened 2019-01-24
Replace remote S3 paths with local paths for test_snake dryruns.

yesimon:
Save time querying S3 remote file presence during dryruns. Since the files are not specified
directly in config (they get operations to add suffixes in the rules), need to create a fake tree
of files (or hardcode the paths to rewrite).

### DIFF
--- a/test/unit/test_snake.py
+++ b/test/unit/test_snake.py
@@ -6,17 +6,23 @@ import sys
 import os
 import subprocess
 import shutil
-import tempfile
-import argparse
 import itertools
-import unittest
+
+import pytest
+import yaml
 
 import util.cmd
 import util.file
-from test import TestCaseWithTmp
+
+
+needs_snakemake = pytest.mark.skipif(
+    sys.version_info < (3, 5),
+    reason='python version is too old for snakemake')
+
 
 if sys.version_info >= (3, 5):
     import snakemake
+
 
 def add_to_sample_list(workdir, sample_list_name, sample_names):
     with open(os.path.join(workdir, 'samples-{}.txt'.format(sample_list_name)), 'a') as outf:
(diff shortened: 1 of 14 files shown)


util/file.py
```diff
@@ -321,6 +321,11 @@ def mkdir_p(dirpath):
         else:
             raise
 
+def touch_p(path):
+    '''Touch file, making parent directories if they don't exist.'''
+    mkdir_p(os.path.dirname(path))
+    open(path, 'a').close()
```
tomkinsc: For creating files this is fine, but simply opening and closing the file will not update the file times as `touch` would. The function `touch(fname, times=None)` within this same file will do so. I don't see `touch_p` used anywhere else in this PR. Is it needed elsewhere?

test/unit/test_snake.py
```diff
@@ -45,73 +50,75 @@ def setup_dummy_simple(sample_names=('G1234', 'G5678', 'G3671.1_r1', 'G3680-1_4'
...
     shutil.copy(os.path.join(util.file.get_project_path(), 'pipes', 'Snakefile'), workdir)
-    shutil.copy(os.path.join(util.file.get_project_path(), 'pipes', 'config.yaml'), workdir)
 
-    os.symlink(util.file.get_project_path(), os.path.join(workdir, 'bin'))
+    with open(os.path.join(util.file.get_project_path(), 'pipes', 'config.yaml')) as f:
+        config = yaml.load(f)
+
+    def translate_remote_s3(uri):
+      remote_path = uri[5:]
+      fake_s3_root = os.path.join(util.file.get_project_path(), 'test', 'input', 's3')
```
tomkinsc: Maybe all of the empty placeholder files should be made dynamically as part of the tests?

yesimon: It's hard to determine what the filenames need to be dynamically. Either need to store a list of filenames to then create at runtime or create empty files - I'm ambivalent about these two.

PR #921, opened 2019-02-08
retry conda environment install on Travis

tomkinsc:
installation of conda packages on Travis often fails due to transient web request errors. This wraps the install command in a retry block to attempt the install up to 3 times.

### DIFF
--- a/travis/install-tools.sh
+++ b/travis/install-tools.sh
@@ -1,5 +1,5 @@
 #!/bin/bash
-set -e -o pipefail
+#set -e -o pipefail
 
 CONDA_CHANNEL_STRING="--override-channels -c broad-viral -c conda-forge -c bioconda -c defaults"
 CONDA_ENV="$(pwd)/tools/conda-tools/default"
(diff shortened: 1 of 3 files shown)


travis/install-tools.sh
```diff
@@ -1,5 +1,5 @@
 #!/bin/bash
-set -e -o pipefail
+#set -e -o pipefail
```
notestaff: maybe better to disable error-checking locally, for commands allowed to fail, rather than for the whole script

yesimon: It seems like conda network commands are incredibly flaky lately, like 1/3 gateway errors?

notestaff: yes, so @tomkinsc 's fix would be great.  though ideally conda itself would have a 'retries' setting.

notestaff:
it seems conda has a built-in config setting for retries:

$ conda config --describe remote_max_retries
# # remote_max_retries (int)
# #   The maximum number of retries each HTTP connection should attempt.
# #
# remote_max_retries: 3

tomkinsc:
Thanks, @notestaff. It looks like conda does indeed [retry 3 times by default](https://github.com/conda/conda/blob/0191812c2bcc531a5de72b6862a03221144fa808/conda/base/context.py#L168), and we're still seeing network errors. I guess I'll try raising the retry count...
Looking at the rest of the [config](https://docs.conda.io/projects/conda/en/latest/configuration.html), we can also try turning the knobs `remote_read_timeout_secs` and `remote_connect_timeout_secs` if it seems like we're seeing timeout failures.

tomkinsc:
Or maybe (some of) the network-related error messages are actually SSL/cert -elated?
- https://github.com/conda/conda/issues/7150
- https://github.com/conda/conda/issues/8046
I'll try adding an SSL/cert update to our conda install script.

PR #922, opened 2019-02-11
Bump CONDA_PERL 5.22 -> 5.26 to match conda-forge pins in CB3

yesimon:
https://github.com/conda-forge/conda-forge-pinning-feedstock/blob/master/recipe/conda_build_config.yaml


notestaff:
@yesimon are there packages on broad-viral, such as the customized kraken, that need to be rebuilt with the new setting?

yesimon:
Kraken will be moved to krakenuniq. Otherwise most other things work except for a downgrade of blast 2.7.1 -> 2.6.0.

dpark01:
Looks good.. why did BLAST have to downgrade? And to Ilya's question: even though our use of kraken will eventually change, did you rebuild our current one for this already?

yesimon:
I think it's because there's no valid build for `boost=1.6.8`

yesimon:
I did rebuild the kraken with cb3 pins so that's fine

dpark01:
Okay, since this partially reverts some of #863, @tomkinsc any objections?

tomkinsc:
No objections; the older BLAST should be fine. When possible we'll probably want to update to BLAST [>v2.8](https://www.ncbi.nlm.nih.gov/books/NBK131777/) to take advantage of the new feature to limit search by taxonomy.

PR #923, opened 2019-02-11
Update CondaPackage InstallMethod for overwritten envs.

yesimon:
Previously `install_package()` will first try to create the environment, which is a no-op if the
environment already exists. But newest conda4 will overwrite the existing environment and replace a
with an empty environment when this happens.


tomkinsc:
In general this looks good; any particular reason for preserving the commented-out bits?

yesimon:
Removed commented out bits

PR #933, opened 2019-03-01
Update illumina_indices.py

09aaronl:
Added additional index names and sequences for Nextera, IDT's xGen UMI adaptors, and small RNA prep


tomkinsc:
Thanks, @09aaronl!

PR #935, opened 2019-03-05
Split out TaxonomyDB into a separate python lib

yesimon:
New python lib ncbitax


dpark01:
Hm, similarly, in order for this PR to be reviewable, we'll need to see the github repo and CI setup for the new package (which is presumably mostly code snipped from here, but still)

PR #938, opened 2019-03-08
 Kraken2 depletion for human reads as alternative to bmtagger

yesimon:
* Add `kraken_unclassified` command that combines p filtering with depletion
* Less sensitive than bmtagger (tradeoffs) but much faster. Can be controlled with p. p=0.01-0.05 a
  close to bmtagger but not as aggressive.
* Custom kraken2 needed to avoid executable name clashes with other kraken* packages.


tomkinsc:
Does it make any sense to use this as a first pass to pare down data prior to running bmtagger?

yesimon:
Not really because bmtagger is already fast enough. It's not going to deplete as many reads as bmtagger.

PR #943, opened 2019-04-02
Revert "Remove merge_compressed_files because they are cattable"

yesimon:
Reverts broadinstitute/viral-ngs#936

The files may be compressed in different ways per file.


tomkinsc:
Closing this; the reversion included here has been included as part of #944.

PR #955, opened 2019-05-29
add aggregate_spike_count to reports.py

tomkinsc:
This adds a new function, `aggregate_spike_count`, to `reports.py` to produce a matrix of spike-in counts per-sample from a directory containing array of per-sample spike-in count files. This is in contrast to `consolidate_spike_count` which simply concatenates the count files (without extraneous columns).

### DIFF
--- a/reports.py
+++ b/reports.py
@@ -10,7 +10,7 @@ import logging
 import glob
 import os
 import time
-from collections import OrderedDict
+from collections import OrderedDict, defaultdict
 import csv
 import math
 import shutil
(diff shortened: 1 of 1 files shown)


reports.py
```diff
@@ -453,6 +453,50 @@ def parser_consolidate_spike_count(parser=argparse.ArgumentParser()):
...
+            if s not in samples_seen:
+                samples_seen.append(s)
+            with open(fn, 'rt') as inf:
+                for line in inf:
+                    if not line.startswith('Input bam') and not line.startswith('*'):
+                        spike, count = [line.strip().split('\t')[i] for i in [0,2]]
+                        spike_in_sample_counts[spike][s] = count
+                        #outf.write('\t'.join([s, spike, count]) + '\n')
+        outf.write("\t".join(["spike-in"]+samples_seen)+"\n")
+        for spike in spike_in_sample_counts.keys():
```
yesimon: Sorted IDs or OrderedDict?

reports.py
```diff
@@ -453,6 +453,50 @@ def parser_consolidate_spike_count(parser=argparse.ArgumentParser()):
...
+            for s in samples_seen:
+                if s in spike_in_sample_counts[spike]:
+                    row.append(spike_in_sample_counts[spike][s])
+                else:
+                    row.append("0")
+            outf.write("\t".join(row)+"\n")
+
+
+def parser_aggregate_spike_count(parser=argparse.ArgumentParser()):
+    parser.add_argument('inDir', help='Input spike count directory.')
```
yesimon: metavar these so variables can avoid camelcasing?

reports.py
tomkinsc: Thanks!

PR #957, opened 2019-06-03
binned coverage plot option in align_and_plot

lakras:
added binning for coverage plots with large reference genomes generated by align_and_plot:
- off by default, can be turned on with option bin_large_plots = True
- plots maximum or minimum read depth in each bin (option binning_summary_statistic = "max" or "min", default is max)
- bin size set to one bin per pixel (bin size displayed in y axis label--Plasmodium bin size is 27 kb)
- binning results in Plasmodium coverage plots with file size 20 KB, compared to 140 MB without binning (unopenable and indecipherable if it does open)

also:
- made sample name field in plot_coverage optional to allow running with new batch tool runner on DNA nexus (gets sample name from reads_unmapped_bam)
- fixed indentation, repeated code in plot_coverage

Here's a coverage plot for the first chromosome of _P. falciparum_ without binning (left; 5.2 MB) and with binning (right; 21 KB).

plasmodium_chr1_binning_comparison

And here's a coverage plot for the whole _P. falciparum_ genome without binning (left; 136.4 MB) and with binning (right; 21 KB).

plasmodium_full_genome_binning_comparison


reports.py
```diff
@@ -673,6 +680,22 @@ def plot_coverage(
             segment_depths.setdefault(row[0], []).append(float(row[2]))
             domain_max += 1
 
+    # Binning
+    bin_size = 1
+    if bin_large_plots:
+        # Number points plotted after binning is set to true width of plot (excluding
+        # axis titles and labels and whitespace around plot), calculated by trial and error
+        # There is one bin per pixel
+        preferred_domain = int(0.907634 * plot_width - 103.207)
```
tomkinsc: Are these constants a function of the DPI of the system? If so, can the width in pixels be calculated similar to how it's done below via `DPI = plot_dpi or fig.get_dpi()`? Matplotlib thinks more in relative dimensions than in pixels to suit its vector backends, so we generally have to take DPI into account as we're [doing below](https://github.com/broadinstitute/viral-ngs/blob/master/reports.py#L679-L680). Can the binning code block be moved within the `with plt.style.context(plot_style)` context below to take advantage of the sizing code there? Or if it's more a function of the data range than the dimensions of the plot (with axis labels, ticks, etc.), perhaps something like [`ax.get_xlim()`](https://matplotlib.org/2.0.0/api/_as_gen/matplotlib.axes.Axes.get_xlim.html) (using the `ax` below that references the subplot) can be used?

reports.py
```diff
@@ -694,19 +717,27 @@ def plot_coverage(
             colors = list(plt.rcParams['axes.prop_cycle'].by_key()['color'])    # get the colors for this style
             segment_color = colors[segment_num % len(colors)]    # pick a color, offset by the segment index
 
+            x_values = range(prior_domain_max, domain_max)
```
tomkinsc: Thanks for tidying up the repeated range statements.

reports.py
```diff
@@ -550,6 +550,12 @@ def parser_plot_coverage_common(parser=argparse.ArgumentParser()):    # parser n
         help="The max coverage depth (default: %(default)s)"
     )
     parser.add_argument('-l', dest="read_length_threshold", default=None, type=int, help="Read length threshold")
+    parser.add_argument(
+        '--binLargePlots',
+        dest="bin_large_plots",
+        action="store_true",
```
tomkinsc: Do you think we should turn on binned plots as the default option for genomes beyond a certain length (the axis `xlim`?)? It would be nice for users to simply get a plot regardless of genome size, and your y-axis labeling code makes it clear whether a given plot has been binned or not. If we do that we'd probably want the ability to override and turn off auto-binning (or set the bin size manually and auto-scale the width?).

reports.py
```diff
@@ -715,7 +746,11 @@ def plot_coverage(
 
         plt.title(plot_title, fontsize=font_size * 1.2)
         plt.xlabel("bp", fontsize=font_size * 1.1)
-        plt.ylabel("read depth", fontsize=font_size * 1.1)
+        
+        ylabel = "read depth"
+        if(bin_size > 1):
+            ylabel = "read depth (max in " + str(bin_size)+ "-bp bin)"
```
tomkinsc: Not important, but the `str` casting and concatenation can be avoided via `format()` (maybe a bit more readable):
`ylabel = "read depth (max in {size}-bp bin)".format(size=bin_size)`

reports.py
```diff
@@ -673,6 +680,22 @@ def plot_coverage(
...
+        # Number points plotted after binning is set to true width of plot (excluding
+        # axis titles and labels and whitespace around plot), calculated by trial and error
+        # There is one bin per pixel
+        preferred_domain = int(0.907634 * plot_width - 103.207)
+
+        # Bin locations and take maximum in each bin
+        bin_size = 1 + int(domain_max/preferred_domain)
+        binned_segment_depths = OrderedDict()
+        for segment_num, (segment_name, position_depths) in enumerate(segment_depths.items()):
+            max_depths_in_bins = [max(position_depths[i:i + bin_size]) for i in range(0, len(position_depths), bin_size)]
```
tomkinsc: This may be a question for others who routinely look at these plots, but since these plots are often used to spot regions of low coverage, I wonder if the `max` depth of positions within a bin is the stat we care about most, or if `min` may be more appropriate. Or maybe the value could be the number of positions within a bin exceeding some threshold value (the mean coverage within the bin?)? Perhaps it should be user-configurable?

pipes/WDL/workflows/tasks/tasks_reports.wdl
```diff
@@ -15,6 +14,9 @@ task plot_coverage {
 
   Boolean? skip_mark_dupes=false
   Boolean? plot_only_non_duplicates=false
+  Boolean? bin_large_plots=false
+  
+  String sample_name = basename(basename(basename(reads_unmapped_bam, ".bam"), ".taxfilt"), ".clean")
```
tomkinsc: Great, thanks for this.

lakras: Welcome. :)

reports.py
lakras: Sure. When I was playing around with different options, I found that the max best represented the actual shape of the plot, which is what I wanted to see. I have also been more interested in the presence rather than absence of coverage. I don't see any harm in giving people options—but I would want max to be one of those options, since it has served me quite well.

reports.py
lakras: Sure thing.

reports.py
lakras: Welcome. :)

reports.py
lakras: Beyond some point, plots without binning are useless. But there is a middle ground where you could bin or not bin. We could have binning turned on by default with an override option, and bin a bit less aggressively (something like two to 20 bins per pixel, rather than one bin per pixel). If a plot doesn't really need it, then it won't be binned. As things currently are, GB virus C gets 11-bp bins when binning is turned out, but the non-binned plot is quite readable and reasonably sized, which makes me think I was a bit too aggressive.

Setting the bin size manually—I would not make this the default, but I think that would make an excellent option—I can see people wanting to be able to do it. Auto-scaling the width would be a nice way to make sure that all coverage plots produced are actually readable, but I don't think it is absolutely necessary.

reports.py
lakras: > Are these constants a function of the DPI of the system?

No, I don't think so.

This is just the width in pixels of the actual plot part of the plot (not including the whitespace and text around the plot)—see attached. Not sure how to actually get that value in a less hacky way—if you happen to know, I would love to know too. I tried different plot_width values and checked what the inside plot width was and did a linear regression on the pairs—hence the constants (messy, I agree). If you prefer I can just use plot_width instead—it won't be the width in pixels of the actual plotted part of the plot (so we won't be able to say that there are _n_ bins per pixel), but it won't have ugly constants.

Plasmodium_coverage_plot-01

lakras: And on second look that variable name sucks. I'll rename it to inner_plot_width and google for a better way to get that width....

yesimon: I think getting the exact size in pixels is hard because the renderer doesn't know how big the axes are going to be and it depends on the text of the y axis labels for example.

pipes/WDL/workflows/tasks/tasks_reports.wdl
lakras: Ah, but sometimes the bam file is named `.clean.bam`.

reports.py
tomkinsc: [`ax.get_xlim()`](https://matplotlib.org/2.0.0/api/_as_gen/matplotlib.axes.Axes.get_xlim.html) should give the dimension of what you annotated in the image above, in arbitrary units.

lakras: I agree. And I haven't any solutions online. I suppose even if I added plot_height to my little linear regression to capture the effect of font size, I still couldn't capture the additional width added by the y axis tick labels in any non-absurd way.... Well. Let's use plot_width instead of trying to get the width of the inside plot. It is not as pleasing but I think it is okay.

lakras: Oops, sorry, didn't see your response @tomkinsc! ax.get_xlim() gives me (0.0, 1.0), but I do not yet know what that means will try to figure out what I can do with it.

lakras: Ah, I see. @tomkinsc, `get_xlim` gives us the limits of the x axis; in this particular case it gives us -1.2e7 to 24.8e7 (more precisely, -1,181,769 to 24,817,161). What I am looking for is the width of that space in pixels. The purpose of getting the width of that space in pixels is to then be able to set the bin size so that there is some known, predetermined number of bins per pixel (currently 1 bin per pixel).

tomkinsc: Matplotlib doesn't think in pixels though, but physical units (inches), which can be converted to pixels based on the DPI of the system (so with a desired pixel dimension and the DPI of the system, you can [set the size in pixels](https://github.com/broadinstitute/viral-ngs/blob/master/reports.py#L679-L680)). Should be able to interrogate the current size, get inches, and multiply by DPI to get the size in pixels.

lakras: Absolutely. I agree with you 100%. But what I would be looking for is not the width of the plot (`plot_width` in pixels or `plot_width / DPI` in inches), but the width of that smaller part of the plot I outlined in blue above. If you know how to interrogate the current size (in inches or pixels!) of that smaller part of the plot I outlined in blue above, then that would definitely be really cool and I would love to know how. As far as I can tell `get_xlim` does not do that, but I would not be sad to be proved wrong.

tomkinsc: If [this](https://stackoverflow.com/a/19306776) works, possibly:
```
bbox = ax.get_window_extent().transformed(fig.dpi_scale_trans.inverted())
width, height = bbox.width, bbox.height
width *= fig.dpi
height *= fig.dpi
```
The matplotlib API leaves much to be desired in terms of usability/memorability; half the time I end up inspecting objects with `dir()`...

lakras: It works!!!! Thou art sorcery! Thank you so so much!!!!

reports.py
```diff
@@ -549,7 +549,25 @@ def parser_plot_coverage_common(parser=argparse.ArgumentParser()):    # parser n
...
+        type=int,
+        help="Read length threshold"
+    )
+    parser.add_argument(
+        '--binLargePlots',
+        dest="bin_large_plots",
+        action="store_true",
+        help="Plot summary read depth in one-pixel-width bins for large plots."
+    )
+    parser.add_argument(
```
tomkinsc: The `--binningSummaryStatistic` argument can be constrained to valid options by passing a list to the [`choices` kwarg](https://docs.python.org/3/library/argparse.html#choices).

tomkinsc: Also, thanks for adding this!

reports.py
```diff
@@ -686,27 +705,60 @@ def plot_coverage(
...
+            
+        # Binning
+        bin_size = 1
+        if bin_large_plots:
+            inner_plot_width_inches = ax.get_window_extent().transformed(fig.dpi_scale_trans.inverted()).width
+            inner_plot_width_px = inner_plot_width_inches * fig.dpi # width of actual plot (sans whitespace and y axis text)
+            bins_per_pixel = 1 # increase to make smaller (but less visible) bins
+            
+            # Parse summary statistic
+            if(binning_summary_statistic.lower() == "min" or binning_summary_statistic.lower() == "minimum"):
```
tomkinsc: Constraining the param above with valid `choices` would simplify this a bit

reports.py
```diff
@@ -686,27 +705,60 @@ def plot_coverage(
...
+                binning_summary_statistic = "min" # for y axis label
+                binning_action = min
+            else:
+                binning_summary_statistic = "max"
```
tomkinsc: Using `binning_action.__name__` where you want a string version of the binning function name would obviate the need to store it separately

reports.py
lakras: Great idea! Thanks!

reports.py
lakras: Removed that whole block of code and replaced with

`binning_action = eval(binning_summary_statistic)`

since `binning_summary_statistic` is constrained by the `choices` now.

reports.py
lakras: Agreed.

reports.py
tomkinsc: Using `eval` was my first instinct as well, despite the fact that it always feels a little wrong, even with constraints. It may be nice to preserve the ability to do it by name in case we ever add an inline function for another bin stat (threshold, q score filter, etc.)

PR #958, opened 2019-06-06
emit log warnings for malformed kraken reports

tomkinsc:
emit log warnings for malformed kraken reports in summary report function

### DIFF
--- a/metagenomics.py
+++ b/metagenomics.py
@@ -1172,7 +1172,7 @@ def taxlevel_summary(summary_files_in, json_out, csv_out, tax_headings, taxlevel
             should_process = False
             indent_of_selection = -1
             currently_being_processed = ""
-            for line in inf:
+            for lineno, line in enumerate(inf):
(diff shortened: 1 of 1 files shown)


metagenomics.py
```diff
@@ -1210,7 +1210,12 @@ def indent_len(in_string):
 
                 row = next(csv.DictReader([line.strip().rstrip('\n')], fieldnames=fieldnames, dialect="kraken_report"))
 
-                indent_of_line = indent_len(row["sci_name"])
+                try:
+                    indent_of_line = indent_len(row["sci_name"])
+                except Exception as e:
```
dpark01: Should this be `KeyError` instead of `Exception` (or something more specific along those lines)?

Separately... do we know what scenarios make kraken behave this way?

tomkinsc: `row["sci_name"]` can be there but return `None` which results in an `AttributeError` when it hits `lstrip()`; I'll change it to that.

This is more of a file integrity check than anything; when I was downloading ~2500 kraken reports via wget, some of them ended up with broken lines for a reason I didn't really dig into.

PR #959, opened 2019-06-06
update default ERCC seq file for WDL

tomkinsc:
update pipeline default DNAnexus reference for ERCC seq file, from 32-seq file to 96-seq file


tomkinsc:
Yep!

PR #969, opened 2019-07-01
add tool wrapper for BBMap

notestaff:


branch-info/is-1808081312-add-bbmap.org
```diff
@@ -0,0 +1,3 @@
+Branch is-1808081312-add-bbmap:
```
tomkinsc: Should org-mode files be added to `.gitignore`?

notestaff: no; i'll just remove the branch-info/ file .  I use branch-info/branch-name.org to store descriptions of what's being done on a given branch.

PR #970, opened 2019-07-02
add clumpify-based dedup

tomkinsc:
add `bbmap.BBMapTool().dedup_clumpify()`, along with unit tests; pass `JVMmemory` to bbmap and clumpify; add `rmdup_clumpify_bam` to `read_utils.py`; change `TestRmdupUnaligned `unit tests for bbmap to use `read_utils.py::rmdup_clumpify_bam`; add `dedup_bam` WDL task to `tasks_read_utils.wdl`


pipes/WDL/workflows/tasks/tasks_read_utils.wdl
```diff
@@ -37,4 +37,32 @@ task downsample_bams {
   }
 }
 
+task dedup_bam {
+  File  in_bam
+  Int?  max_mismatches=5
```
yesimon: Seems kind of high?

pipes/WDL/workflows/tasks/tasks_read_utils.wdl
```diff
@@ -37,4 +37,32 @@ task downsample_bams {
...
+        in_bam \
+        ${sample_name}.dedup.bam \
+        ${'--maxMismatches=' + max_mismatches} \
+        --JVMmemory "8g"
+
+    reports.py fastqc ${sample_name}.dedup.bam ${sample_name}.deduped_fastqc.html
+  }
+
+  output {
+    File deduplicated_bam               = "${sample_name}.dedup.bam"
```
yesimon: Should settle on `dedup` vs `deduped`

read_utils.py
```diff
@@ -902,9 +903,49 @@ def parser_rmdup_cdhit_bam(parser=argparse.ArgumentParser()):
...
     util.cmd.attach_main(parser, rmdup_cdhit_bam, split_args=True)
     return parser
+__commands__.append(('rmdup_cdhit_bam', parser_rmdup_cdhit_bam))
 
 
-__commands__.append(('rmdup_cdhit_bam', parser_rmdup_cdhit_bam))
+def rmdup_clumpify_bam(inBam, outBam, max_mismatches=None, JVMmemory=None):
+    ''' Remove duplicate reads from BAM file using bbmap's clumpify tool.
+    '''
+    max_mismatches = max_mismatches or 4
```
yesimon: Make consistent with WDL. I would say 3-4 is a good default.

read_utils.py
```diff
@@ -902,9 +903,49 @@ def parser_rmdup_cdhit_bam(parser=argparse.ArgumentParser()):
...
+
+    bbmap = tools.bbmap.BBMapTool()
+    out_bams = []
+    for f in os.listdir(tmp_dir):
+        out_bam = mkstempfname('.bam')
+        out_bams.append(out_bam)
+        library_sam = os.path.join(tmp_dir, f)
+
+        log.info("executing BBMap clumpify on library " + library_sam)
+        # cd-hit-dup cannot operate on piped fastq input because it reads twice
```
yesimon: Remove

tools/bbmap.py
```diff
@@ -28,38 +28,94 @@ def __init__(self, install_methods=None):
...
+                                This is platform-specific; recommendations:
+                                   NextSeq      40  (and spany=t)
+                                   HiSeq 1T     40
+                                   HiSeq 2500   40
+                                   HiSeq 3k/4k  2500
+                                   Novaseq      12000
+                k=31            Use kmers of this length (1-31).  Shorter kmers may
+                                increase compression, but 31 is recommended for error
+                                correction.                
+    '''
```
yesimon: indent

yesimon: Possibly add comment that optical duplicates is weaker dedupe limited by xy-position. Meant for single-end sequencing like that commonly used in NextSeq

pipes/WDL/workflows/tasks/tasks_read_utils.wdl
tomkinsc: The [default for mvicuna](https://github.com/broadinstitute/mvicuna#running-vicuna) rmdup is 5 as well (`-drm_max_mismatch: default 5`). We can certainly reduce it.

pipes/WDL/workflows/tasks/tasks_read_utils.wdl
tomkinsc: Will do, thanks

read_utils.py
tomkinsc: I'll make it 3 across the board.

read_utils.py
tomkinsc: thanks

tomkinsc:
Let's hold off on merging this for now. While the tests pass here, the slow tests from DNAnexus (for which we do not have feedback on GitHub) are currently failing during dedup, with a NullPointerException in Picard. For example: https://platform.dnanexus.com/projects/F8PQ6380xf5bK0Qk0YPjB17P/monitor/job/FZZbVxj0xf5jZ07ZJ71gpFJ6
It's possible PR #977 may resolve the issue, so the changes there should be merged first.

tomkinsc:
We should either remove `containment=t` or wait to merge this PR until there's movement to address the apparent bug in bbmap identified by @yesimon: https://sourceforge.net/p/bbmap/tickets/18/

pipes/WDL/workflows/demux_metag.wdl
```diff
@@ -5,18 +5,26 @@ import "tasks_metagenomics.wdl" as metagenomics
...
   call demux.illumina_demux as illumina_demux
 
   scatter(raw_reads in illumina_demux.raw_reads_unaligned_bams) {
+      call reads.dedup_bam as dedup {
+          input:
+              in_bam = raw_reads        
+      }
+  }
+
+  scatter(reads_bam in dedup.dedup_bam) {
```
dpark01: I wouldn't double-scatter. You can just keep this as a single scatter block on the raw_reads and put all the task calls together in that single scatter. WDL interpreters/compilers are smart enough to figure out the DAG and parallelization opportunities within the scatter based on the dependencies between their inputs and outputs.

dpark01: So specifically:
```
  scatter(raw_reads in illumina_demux.raw_reads_unaligned_bams) {
    call reads.dedup_bam as dedup {
        input:
            in_bam = raw_reads        
    }
    call reports.spikein_report as spikein {
      input:
        reads_bam = dedup.dedup_bam
    }
    call taxon_filter.deplete_taxa as deplete {
      input:
        raw_reads_unmapped_bam = dedup.dedup_bam
    }
    call assembly.assemble as spades {
      input:
        assembler = "spades",
        reads_unmapped_bam = deplete.cleaned_bam
    }
  }
```

pipes/WDL/workflows/demux_metag.wdl
```diff
@@ -5,18 +5,26 @@ import "tasks_metagenomics.wdl" as metagenomics
 import "tasks_taxon_filter.wdl" as taxon_filter
```
dpark01: If you think this is ready and want to try it out, shouldn't you remove #DX_SKIP_WORKFLOW?

pipes/WDL/workflows/demux_metag.wdl
```diff
@@ -39,6 +47,6 @@ workflow demux_metag {
   }
   call metagenomics.kaiju as kaiju {
     input:
```
dpark01: We haven't really used kaiju regularly via WDL yet, but I'm betting that we may want to consider moving it to a scatter-on-single-sample execution mode (like everything else in our WDLs except kraken). Its database is about 4x smaller (I'm guessing the localization time is just a few minutes) and the execution time of the algorithm is much slower, so the cost efficiency (algorithmic compute time vs VM wall clock time) of kaiju on a single sample is much better than kraken... so we might as well move it within the same scatter block as well.

pipes/WDL/workflows/demux_metag.wdl
```diff
@@ -27,7 +35,7 @@ workflow demux_metag {
 
   call metagenomics.krakenuniq as kraken {
     input:
-      reads_unmapped_bam = illumina_demux.raw_reads_unaligned_bams,
+      reads_unmapped_bam = dedup.dedup_bam
   }
   call reports.aggregate_metagenomics_reports as metag_summary_report {
```
dpark01: Can we call aggregate_metageomics_reports a second time on the kaiju outputs as well?

pipes/WDL/workflows/demux_plus.wdl
```diff
@@ -13,15 +14,23 @@ workflow demux_plus {
...
     Array[File]? bwaDbs
+
     scatter(raw_reads in illumina_demux.raw_reads_unaligned_bams) {
+        call reads.dedup_bam as dedup {
+            input:
+                in_bam = raw_reads        
+        }
+    }
+    
+    scatter(reads_bam in dedup.dedup_bam) {
```
dpark01: See my comment in demux_metag about combining the scatter blocks.

taxon_filter.py
```diff
@@ -109,9 +109,12 @@ def main_deplete(args):
                                         tags_to_clear = args.tags_to_clear,
                                         picardOptions = ['MAX_DISCARD_FRACTION=0.5'],
                                         JVMmemory     = args.JVMmemory,
-                                        sanitize      = not args.do_not_sanitize) as bamToDeplete:
+                                        sanitize      = not args.do_not_sanitize) as bam_to_dedup:
+
+        read_utils.rmdup_mvicuna_bam(bam_to_dedup, args.rmdupBam, JVMmemory=args.JVMmemory)
```
dpark01: In this new world, should we consider:
1. dropping deduplication entirely from taxon_filter.deplete -- since you now include it in all the pipelines prior to depletion anyway, and since it never really fit in the scope of the name of the command, it was historically embedded in a funny place between depletion steps primary because of its performance profile: it was slower than bmtagger (so we ran it after that) but faster than blastn (so we ran it before that)
2. dropping mvicuna altogether if we think bbmap is better

pipes/WDL/workflows/demux_metag.wdl
dpark01: Two more thoughts on this topic:
1. Importantly, merging them would allow the execution platform to keep going on the linear DAG portions of each sample as they become ready without waiting for all samples to complete dedup before proceeding to the next steps.
2. I wonder if we should consider running the spikein counting step on raw / non-deduplicated reads.... ERCC's are so short that we might quickly hit an artificial upper bound on the counts if we do it on dedup output.

pipes/WDL/workflows/demux_metag.wdl
tomkinsc: `metagenomics.py::taxlevel_summary()` hasn't been adapted/tested to read kaiju summary files yet. I'd like that to be a separate PR (this one is already way beyond its initial scope).

pipes/WDL/workflows/demux_plus.wdl
```diff
@@ -13,38 +14,59 @@ workflow demux_plus {
...
                 trim_clip_db = trim_clip_db,
                 always_succeed = true
         }
-    }
 
-    call metagenomics.krakenuniq as krakenuniq {
-        input:
-            reads_unmapped_bam = illumina_demux.raw_reads_unaligned_bams
+        # classify de-duplicated reads to taxa via krakenuniq
+        call metagenomics.krakenuniq as krakenuniq {
```
dpark01: Oh, I don't think I'd bump this one inside the scatter. Unless maybe you want to take the opportunity to collect some data from this branch on cost efficiency changes (I've always been curious). The kraken and krakenuniq dbs have always burned about 10-15mins per job on db localization and unpacking -- which, when multiplied by the number of samples on highmem machines, adds up (which motivated the batched approach). But on principle I've always wanted to improve our staging time and move it within the scatter, but I guess I always assumed we'd never get there until we move to kraken2 (much smaller databases). That said, what's the dollars-per-demux-plus on our test/CI flowcell comparing scattered kraken vs batched kraken? If it's only a couple bucks extra, I might be fine with moving there now.

tomkinsc: I was curious of the cost delta, but CI builds don't show us costs since they're billed to the DNAnexus Science Team. I'll just move it back outside the scatter...

tools/bbmap.py
```diff
@@ -28,38 +31,223 @@ def __init__(self, install_methods=None):
...
+                            for line in inf:
+                                if (line_num % 4) == 0:
+                                    idVal = line.rstrip('\n')[1:]
+                                    if idVal.endswith('/1'):
+                                        outf.write(idVal[:-2] + '\n')
+                                line_num += 1
+
+                per_lb_read_lists.append(per_lb_read_list)
+        # merge per-library read lists together
+        util.file.concat(per_lb_read_lists, readListAll)
```
yesimon: Can't the read lists be written to a single file to start with instead of concat'ing later?

tomkinsc: The read lists are separate in prep for parallelizing across libraries.

tools/bbmap.py
```diff
@@ -28,38 +31,223 @@ def __init__(self, install_methods=None):
...
+                                    **kwargs)
+                # Make a list of reads to keep
+                with open(per_lb_read_list, 'at') as outf:
+                    for fq in (outFastq1, outFastq2):
+                        with util.file.open_or_gzopen(fq, 'rt') as inf:
+                            line_num = 0
+                            for line in inf:
+                                if (line_num % 4) == 0:
+                                    idVal = line.rstrip('\n')[1:]
+                                    if idVal.endswith('/1'):
```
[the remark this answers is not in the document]
tomkinsc: Do you have an example of what you mean? NovaSeq-like read IDs look the same from a few bam files I compared (NovaSeq vs HiSeq). Since we're doing the conversion to fastq files it should hopefully be fine? This is also the same code we've been using for mvicuna dedup, so the behavior should be the same (though that's not to say we missed a code path to update for NovaSeq support).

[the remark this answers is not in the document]
yesimon: Read IDs that identify the pairs with a space a `1:N:0:barcodes` instead of `/1` or `/2`
image

tomkinsc:
Here's another one, time to port over to viral-core?

PR #976, opened 2019-07-11
fix wdl type bug in tasks_assembly task scaffold

notestaff:


tomkinsc:
Thanks, @notestaff. Looks like we both opened a PR. Closing this one since #975 has been merged.

PR #977, opened 2019-07-11
update picard 2.18.11 -> 2.20.3

tomkinsc:
This PR does the following:
- updates `picard` 2.18.11 -> 2.20.3


dpark01:
Wait, regarding [this commit](https://github.com/broadinstitute/viral-ngs/pull/977/commits/eed5b082204b0ef46dde678d5d4345495039f284), maybe instead of "fixing" the test input file, should we be adding code that sanitizes input all the time now? Because apparently this wasn't an issue for old versions of Picard, but now it is, so these changes will break things that used to work?

tomkinsc:
Here's the (human-data-centric) discussion upstream about the breaking changes to htsjdk:
https://github.com/samtools/hts-specs/pull/333
Allowing some characters in the reference name seems to have put the resulting files out of spec with the SAM standard. We can sanitize if you think that's best. Or maybe if the reference sequences we include are from GenBank including the accession in addition to the sequence name would resolve the issue and be more 'correct.'

PR #978, opened 2019-07-12
add .git to .dockerignore

tomkinsc:
add .git to .dockerignore to reduce the size of the build context sent to the Docker daemon (all of the files in the pwd)

### DIFF
--- a/.dockerignore
+++ b/.dockerignore
@@ -2,3 +2,4 @@
 __pycache__
 tools/conda-cache
 tools/conda-tools
+.git


dpark01:
hm... I forget, do we save the git commit hash anywhere within the docker image? If not, this means we lose the ability to be sure of that.. an alternative could be to do a git checkout depth 1 or something?

tomkinsc:
Shoot, yes, we do use it internally during Docker build in the [call](https://github.com/broadinstitute/viral-ngs/blob/master/Dockerfile#L55) to write the `VERSION` file. Docker copies the entire Travis working directory, which currently does have a limit to its checkout depth (set to 150). I'll close this.

PR #985, opened 2019-09-12
skip build_conda stage in travis build matrix

tomkinsc:
Skip conda build to decrease test times since we are moving away from conda for distribution of viral-ngs. Only skip for now since we may want to re-use some of the build machinery for a refactored and modularized future version of viral-ngs


tomkinsc:
This PR has prompted users of the conda package to contact me via e-mail. Closing this for now, in favor of deprecation of the conda package prior to refactoring of viral-ngs.

PR #992, opened 2019-10-09
escape vcf path passed to snpeff

tomkinsc:
quote vcf path passed to snpeff so directories with spaces and other escapable characters are read correctly

### DIFF
--- a/tools/snpeff.py
+++ b/tools/snpeff.py
@@ -10,6 +10,7 @@ import tempfile
 import logging
 import subprocess
 import shutil
+import shlex
 
 # third-party
 import pysam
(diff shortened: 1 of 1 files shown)


dpark01:
Wait -- is there an example case where this code change fixes something? I would expect this kind of thing to only matter if we were using `os.system` calls like we did many years ago, but as far as I can tell, this code ought to be calling `subprocess.*call` or `Popen` or some other thing where we're not actually calling it via a shell invocation (quotations should not be necessary, whitespace should be fine).

That being said, some of our `util.misc.run*` code is a bit complex, and since we already decided to drop py2 support (and require py >=3.5) I think we could probably simplify or remove a ton of lines from there...

tomkinsc:
Yes, I encountered a failure when running snpeff on a vcf with a space in its file path (why Google Drive uses `Shared drives/` as a directory name name is beyond me). It appears to be passed correctly as a single argument by our code _to_ snpEff, so the failure appears to be downstream somewhere in snpEff. Quoting or escaping seems to resolve the issue.

dpark01:
Ah so, technically maybe this ought to be a PR in the upstream wrapper scripts or code of snpEff itself, but this is a workaround based on the assumption that snpEff's executable is a shell script.

tomkinsc:
True; just opened a PR [upstream](https://github.com/bioconda/bioconda-recipes/pull/18010) to fix the snpeff wrapper; closing this PR.

PR #1000, opened 2019-12-03
add code to illumina.py::RunInfo() to infer instrument model

tomkinsc:


tomkinsc:
Should we migrate this over to viral-core?

PR #1066, opened 2026-05-11
Fix CVE-2026-42246: upgrade Ruby net-imap gem to >=0.6.4

dpark01:
Fixes #1065 

## Changes

Applies the established gem-fixup pattern (remove old default gem, install patched version) to `net-imap` in three Dockerfiles:
- `docker/Dockerfile.assemble`
- `docker/Dockerfile.phylo`
- `docker/Dockerfile.mega`

Each file now removes the vulnerable `net-imap` 0.6.2 default gem and installs `>=0.6.4` inline in the same RUN layer (so the vulnerable files never appear in a committed layer).

## Vulnerability Background

**CVE-2026-42246**: STARTTLS-stripping vulnerability where a MITM can cause `Net::IMAP#starttls` to return successfully without TLS being negotiated (CVSS 4.0 score 7.6 HIGH).

The gem enters via the `mummer4 -> yaggo -> ruby` transitive chain — the same chain that introduced CVE-2026-33210 (json) and CVE-2026-41316 (erb).

Practical exploitability is near-zero (no IMAP connections are made in our pipelines), but the Rego policy correctly surfaces it because it has `AV:N + VC:H + VI:H` (network-accessible with high confidentiality and integrity impact).

## Testing

CI will:
- Build all three images on amd64 and arm64 (native runners, no emulation)
- Run unit tests (no functional change — this is a gem replacement)
- Re-scan images via `container-scan.yml` (CVE-2026-42246 should no longer appear)

---

*Authored by Claude Sonnet 4.5 via Claude Code. Copilot review feedback addressed in commit 74552556.*

### DIFF
--- a/.github/actions/setup-docker-build/action.yml
+++ b/.github/actions/setup-docker-build/action.yml
@@ -19,16 +19,16 @@ runs:
   steps:
     - name: Checkout repository
       if: ${{ inputs.checkout == 'true' }}
-      uses: actions/checkout@v4
+      uses: actions/checkout@v5
       with:
         fetch-depth: ${{ inputs.fetch-depth }}
         fetch-tags: ${{ inputs.fetch-tags }}
 
     - name: Set up Docker Buildx
-      uses: docker/setup-buildx-action@v3
+      uses: docker/setup-buildx-action@v4
 
     - name: Log in to GitHub Container Registry
-      uses: docker/login-action@v3
+      uses: docker/login-action@v4
       with:
         registry: ghcr.io
         username: ${{ github.actor }}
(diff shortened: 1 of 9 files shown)


dpark01:
Thanks @copilot-pull-request-reviewer for catching those issues! Fixed in 74552556:

1. **Depth issue**: Changed `-maxdepth 3` to `-maxdepth 4`. The gemspec is at `/opt/conda/lib/ruby/gems/4.0.0/specifications/net-imap-0.6.2.gemspec` which is 4 levels deep from the find starting point.

2. **Gemspec path issue**: Changed `specifications/default/` to `specifications/`. Unlike the typical default gem pattern, net-imap is installed as a regular gem in the specifications directory, not in the default/ subdirectory.

Verified these paths by inspecting the current main-assemble image.

PR #1072, opened 2026-05-19
Feature/centrifuger integration

carze:
Adds Centrifuger (https://github.com/mourisl/centrifuger) as a metagenomics classifier alongside the existing kraken2, kma, kb, and krona tooling. Exposes four CLI subcommands under `viral_ngs metagenomics`:

| Command | Purpose |
|---|---|
| `centrifuger` | Classify reads from an unaligned BAM |
| `centrifuger_build` | Build a centrifuger index from FASTAs or a file list |
| `centrifuger_quant` | Quant report from a per-read classification (centrifuger native / metaphlan / CAMI / kraken-report formats) |
| `centrifuger_kreport` | Kraken-style hierarchical report from a per-read classification |

## Changes

- **Conda dep**: `centrifuger>=1.0` added to `docker/requirements/classify.txt`. Bioconda has `linux-aarch64` builds, so no entry in `classify-x86.txt` is needed — image builds cleanly on both architectures.
- **Tool wrapper**: `src/viral_ngs/classify/centrifuger.py` — `Centrifuger(core.Tool)` with `build()`, `classify()`, `quant()`, `kreport()` methods. Follows the established `kraken2.py` pattern (Picard SamToFastq with illumina clipping, paired-vs-single detection by FASTQ size, empty-input short-circuits at both `classify()` and `kreport()`, `try/finally` temp cleanup).
- **Empty-input contract**: `classify()` silently skips empty BAMs (writes an empty classification TSV) and `kreport()` silently skips empty/header-only classifications (writes an empty report). This keeps a `classify → kreport` chain working unconditionally — without it, downstream consumers like the viral-pipelines `feature-centrifuger` WDL would fail on empty BAMs because `centrifuger-kreport` exits 255 on empty input. `classify()` also raises `FileNotFoundError` early when the BAM path doesn't exist (fail fast instead of letting samtools surface a less-clear error). `_classification_is_empty()` is header-aware so `is_count_table=True` mode correctly recognizes single-row count tables as non-empty.
- **CLI**: four new subcommands in `src/viral_ngs/metagenomics.py`, using snake_case flags (consistent with the more recently added classify commands — kb, kma_build, kraken2_build).
- **Tests**: 16 new tests across 3 files (11 unit, 4 CLI plumbing, 3 integration; superseding the original single combined classify+quant integration test).

## Testing

Verified in the classify container with centrifuger installed:

```
docker run --rm -v $(pwd):/opt/viral-ngs/source \
  quay.io/broadinstitute/viral-ngs:main-classify bash -c \
  "micromamba install -y -c bioconda 'centrifuger>=1.0' && \
   pip install -e /opt/viral-ngs/source --quiet && \
   pytest -rsxX -n auto \
     /opt/viral-ngs/source/tests/unit/classify/test_tools_centrifuger.py \
     /opt/viral-ngs/source/tests/unit/classify/test_integration_centrifuger.py \
     /opt/viral-ngs/source/tests/unit/classify/test_metagenomics.py"
```

Result: **36 passed, 1 skipped** (pre-existing krakenuniq skip, unrelated).

Test coverage:
- **Unit (mocked subprocess)** — 11 tests: build (arg construction + ref XOR validation), classify (paired, single, empty-BAM short-circuit, missing-BAM raises), quant, kreport (all flags + defaults + 3 short-circuit cases + count-table mode + positive control).
- **CLI plumbing** — 4 tests: each `parser_centrifuger*` parses CLI args and dispatches to the expected mocked `Centrifuger.*` call.
- **Integration (real binary, end-to-end via CLI)** — 3 tests against `TestMetagenomicsSimple` fixture: build → classify → quant + kreport. Cleanly skipped via `pytestmark.skipif` when `centrifuger` is not on PATH.

## Compatibility

- Multi-arch (x86_64 + ARM64) — verified bioconda has both builds.
- No changes to existing kraken2/kma/kb behavior or tests.
- No new top-level Python dependencies (`pyproject.toml` unchanged).

**Feature created utilizing Opus 4.7 and GPT5.5**

### DIFF
--- a/src/viral_ngs/classify/centrifuger.py
+++ b/src/viral_ngs/classify/centrifuger.py
@@ -1,6 +1,7 @@
 '''
 Centrifuger metagenomics classifier.
 '''
+import csv
 import logging
 import os
 import shutil
@@ -13,6 +14,17 @@ import viral_ngs.core.picard as picard
 import viral_ngs.core.samtools as samtools
 
 log = logging.getLogger(__name__)
+
+CLASSIFICATION_HEADER = [
+    'readID',
+    'seqID',
+    'taxID',
+    'score',
+    '2ndBestScore',
+    'hitLength',
+    'queryLength',
+    'numMatches',
+]
 
 
 class Centrifuger(core.Tool):
@@ -97,7 +109,7 @@ class Centrifuger(core.Tool):
 
         self.execute('centrifuger-build', options=opts)
 
-    def classify(self, in_bam, db, output, k=None, unclassified_prefix=None,
+    def classify(self, in_bam, db, output, k=1, unclassified_prefix=None,
                  classified_prefix=None, min_hitlen=None, hitk_factor=None,
                  merge_readpair=False, num_threads=None):
         '''Classify reads from an unaligned BAM file.'''
@@ -105,8 +117,8 @@ class Centrifuger(core.Tool):
             raise FileNotFoundError(in_bam)
         if samtools.SamtoolsTool().isEmpty(in_bam):
             log.warning('Input BAM is empty, skipping Centrifuger classification')
-            with open(output, 'wt'):
-                pass
+            with open(output, 'wt') as outf:
+                outf.write('\t'.join(CLASSIFICATION_HEADER) + '\n')
             return
 
         tmp_fastq1 = util_file.mkstempfname('.1.fastq')
@@ -172,6 +184,49 @@ class Centrifuger(core.Tool):
             opts['--output-format'] = output_format
 
         self.execute('centrifuger-quant', output=output, options=opts)
+
+    @staticmethod
+    def classification_to_kraken2(classification, output):
+        '''Convert Centrifuger classification TSV to Kraken2-style read TSV.
+
+        This assumes the Centrifuger classification was produced with k=1.
+        The output contains one classified Kraken2-style row per classified
+        Centrifuger row and omits unclassified reads.
+        '''
+        previous_classified_read_id = None
+
+        with util_file.open_or_gzopen(str(classification), 'rt') as inf, \
+                util_file.open_or_gzopen(str(output), 'wt') as outf:
+            header = inf.readline()
+            if not header:
+                return
+            writer = csv.writer(outf, delimiter='\t', lineterminator='\n')
+
+            for line_no, line in enumerate(inf, start=2):
+                if not line.strip():
+                    continue
+                cols = line.rstrip('\n').split('\t')
+                if len(cols) < 7:
+                    raise ValueError(
+                        'Malformed Centrifuger classification row '
+                        f'at line {line_no}: expected at least 7 columns'
+                    )
+
+                read_id = cols[0]
+                seq_id = cols[1]
+                tax_id = cols[2]
+                query_length = cols[6]
+
+                if tax_id == '0' or seq_id == 'unclassified':
+                    continue
+                if read_id == previous_classified_read_id:
+                    raise ValueError(
+                        'Consecutive duplicate classified readID in Centrifuger output: '
+                        f'{read_id}. Expected k=1 input.'
+                    )
+
+                previous_classified_read_id = read_id
+                writer.writerow(['C', read_id, tax_id, query_length, 'N/A'])
 
     def kreport(self, db, classification, output, no_lca=False,
                 show_zeros=False, is_count_table=False, min_score=None,
--- a/src/viral_ngs/classify/centrifuger.py
+++ b/src/viral_ngs/classify/centrifuger.py
@@ -1,0 +1,172 @@
+'''
+Centrifuger metagenomics classifier.
+'''
+import logging
+import os
+import shutil
+import subprocess
+
+import viral_ngs.core as core
+import viral_ngs.core.file as util_file
+import viral_ngs.core.misc as util_misc
+import viral_ngs.core.picard as picard
+import viral_ngs.core.samtools as samtools
+
+log = logging.getLogger(__name__)
+
+
+class Centrifuger(core.Tool):
+
+    def __init__(self, install_methods=None):
+        if not install_methods:
+            install_methods = [
+                core.PrexistingUnixCommand(
+                    shutil.which('centrifuger'),
+                    require_executability=False,
+                )
+            ]
+        super().__init__(install_methods=install_methods)
+
+    def version(self):
+        # Centrifuger uses `-v` (not `--version`); passing `--version`
+        # triggers a buffer overflow in the binary and aborts.
+        try:
+            result = subprocess.run(
+                ['centrifuger', '-v'],
+                capture_output=True,
+                text=True,
+                check=True,
+            )
+        except (subprocess.CalledProcessError, FileNotFoundError):
+            return 'unknown'
+        return (result.stdout or result.stderr).strip() or 'unknown'
+
+    @property
+    def libexec(self):
+        if not self.executable_path():
+            self.install_and_get_path()
+        return os.path.dirname(self.executable_path())
+
+    def execute(self, command, output=None, args=None, options=None):
+        '''Run a Centrifuger command.'''
+        args = args or []
+        options = options or {}
+
+        cmd = [command]
+        for key, value in options.items():
+            if value is None:
+                cmd.append(key)
+            elif isinstance(value, (list, tuple)):
+                for item in value:
+                    cmd.extend([key, str(item)])
+            else:
+                cmd.extend([key, str(value)])
+        cmd.extend(args)
+
+        log.debug('Calling %s: %s', command, ' '.join(cmd))
+        if output:
+            with open(output, 'wt') as outf:
+                subprocess.check_call(cmd, stdout=outf)
+        else:
+            subprocess.check_call(cmd)
+
+    def build(self, db_prefix, taxonomy_tree, name_table, ref_fastas=None,
+              ref_list=None, conversion_table=None, build_mem=None,
+              num_threads=None):
+        '''Build a Centrifuger index.'''
+        if bool(ref_fastas) == bool(ref_list):
+            raise ValueError('Specify exactly one of ref_fastas or ref_list')
+
+        opts = {
+            '--taxonomy-tree': taxonomy_tree,
+            '--name-table': name_table,
+            '-o': db_prefix,
+        }
+        if ref_fastas:
+            if isinstance(ref_fastas, str):
+                ref_fastas = [ref_fastas]
+            opts['-r'] = ref_fastas
+        else:
+            opts['-l'] = ref_list
+        if conversion_table:
+            opts['--conversion-table'] = conversion_table
+        if build_mem:
+            opts['--build-mem'] = build_mem
+        if num_threads:
+            opts['-t'] = util_misc.sanitize_thread_count(num_threads)
+
+        self.execute('centrifuger-build', options=opts)
+
+    def classify(self, in_bam, db, output, k=None, unclassified_prefix=None,
+                 classified_prefix=None, min_hitlen=None, hitk_factor=None,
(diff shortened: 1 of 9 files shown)


src/viral_ngs/classify/centrifuger.py
```diff
@@ -0,0 +1,230 @@
...
+    def classify(self, in_bam, db, output, k=None, unclassified_prefix=None,
+                 classified_prefix=None, min_hitlen=None, hitk_factor=None,
+                 merge_readpair=False, num_threads=None):
+        '''Classify reads from an unaligned BAM file.'''
+        if not os.path.isfile(in_bam):
+            raise FileNotFoundError(in_bam)
+        if samtools.SamtoolsTool().isEmpty(in_bam):
+            log.warning('Input BAM is empty, skipping Centrifuger classification')
+            with open(output, 'wt'):
+                pass
```
dpark01: According to [docs](https://github.com/mourisl/centrifuger#inputoutput), a semantically empty output is not an empty file

```suggestion
            with open(output, 'wt') as outf:
                row = ["readID", "seqID", "taxID", "score", "2ndBestScore", "hitLength", "queryLength", "numMatches"]
                outf.write('\t'.join(row) + '\n')
```

dpark01 (changes requested):
This is great thanks! I looked through it by hand (not by AI!) and only had two suggestions:
1. is the minor comment shown in-line
2. The larger ask is that it'd be great to add a helper function to the `Centrifuger` class that (because it's computationally cheap) maybe we'd invoke all the time in `main_centrifuger` to produce one more output file, perhaps something like `out_classified_reads_kstyle` or something like that. The helper function would do a transformation of [this tsv format](https://github.com/mourisl/centrifuger#inputoutput) to [that tsv format](https://github.com/DerrickWood/kraken2/wiki/Manual#standard-kraken-output-format) while using `.core.file.open_or_gzopen` to write the output (and our WDLs can ask for .tsv.gz outputs when invoking). This would be a basic like.. col 1 = "C"; col2 = input col1; col3 = col3; col4 = col7; col5 can be left blank or "N/A" (we don't really read this). This would allow our `filter_bam_to_taxa` command to work out of the box and for the outputs to be used interchangeably with other classifier tools in viral-ngs. To be honest, the col2 and col3 are the only ones we truly rely on downstream.
3. Sorry one more thing -- have we tested whether the existing `krona` command in `metagenomics.py --inputType kraken2` works seamlessly on the centrifuger_kreport output?

src/viral_ngs/classify/centrifuger.py
dpark01: In fact -- maybe also add an integration test that runs centrifuger on an empty.bam followed by running -kreport on the resulting output (and validating the output of that)?

carze:
> 2. The larger ask is that it'd be great to add a helper function to the `Centrifuger` class that (because it's computationally cheap) maybe we'd invoke all the time in `main_centrifuger` to produce one more output file, perhaps something like `out_classified_reads_kstyle` or something like that. The helper function would do a transformation of [this tsv format](https://github.com/mourisl/centrifuger#inputoutput) to [that tsv format](https://github.com/DerrickWood/kraken2/wiki/Manual#standard-kraken-output-format) while using `.core.file.open_or_gzopen` to write the output (and our WDLs can ask for .tsv.gz outputs when invoking). This would be a basic like.. col 1 = "C"; col2 = input col1; col3 = col3; col4 = col7; col5 can be left blank or "N/A" (we don't really read this). This would allow our `filter_bam_to_taxa` command to work out of the box and for the outputs to be used interchangeably with other classifier tools in viral-ngs. To be honest, the col2 and col3 are the only ones we truly rely on downstream.

Let me take a look at this one -- it might be easy to do this by back-porting from the k2report format (maybe a utility out there for it) otherwise shouldn't be too hard to tackle.

> 3. Sorry one more thing -- have we tested whether the existing `krona` command in `metagenomics.py --inputType kraken2` works seamlessly on the centrifuger_kreport output?

I'll try this to see if it works without any modifications.

dpark01:
> Let me take a look at this one -- it might be easy to do this by back-porting from the k2report format (maybe a utility out there for it) otherwise shouldn't be too hard to tackle.

Well one advantage to doing it ourselves (it's just rearranging tsv columns) vs calling a utility is that we would gain the multi-format optional compression option of the output file (it's one row per read, so millions of rows of tsv)

carze:
> > Let me take a look at this one -- it might be easy to do this by back-porting from the k2report format (maybe a utility out there for it) otherwise shouldn't be too hard to tackle.
> 
> Well one advantage to doing it ourselves (it's just rearranging tsv columns) vs calling a utility is that we would gain the multi-format optional compression option of the output file (it's one row per read, so millions of rows of tsv)

OK; so it looks like the only challenge here is that Centrifuger can emit multiple hits per-read, where K2 will LCA down to one classification per read. We can kinda come close by setting the `-k 1` parameter when invoking Centrifuger which will force it to output one row per read ID and use some logic to approximate the "correct" hit. 

We'd want to hardcode `-k` to 1 in this case as if we call this by default it would break if we had multiple hits per read. Alternatively we can have this as an optional output that when toggled on sets `-k 1`?

dpark01:
> > > Let me take a look at this one -- it might be easy to do this by back-porting from the k2report format (maybe a utility out there for it) otherwise shouldn't be too hard to tackle.
> > 
> > 
> > Well one advantage to doing it ourselves (it's just rearranging tsv columns) vs calling a utility is that we would gain the multi-format optional compression option of the output file (it's one row per read, so millions of rows of tsv)
> 
> OK; so it looks like the only challenge here is that Centrifuger can emit multiple hits per-read, where K2 will LCA down to one classification per read. We can kinda come close by setting the `-k 1` parameter when invoking Centrifuger which will force it to output one row per read ID and use some logic to approximate the "correct" hit.
> 
> We'd want to hardcode `-k` to 1 in this case as if we call this by default it would break if we had multiple hits per read. Alternatively we can have this as an optional output that when toggled on sets `-k 1`?

Ahh.... interesting. I wonder how the kreport script deals with this.

I mean.. well this is a bit more work but might be worth it if we don't trust the `-k 1` behavior:
1. slurp in all the rows into a sqlite or something (unless it's pre-sorted by read ID in which case we can do this in memory with a dequeue or something)
2. run all the rows for a given read ID (which hopefully includes all hits from both reads in a pair since they share the same read ID) through LCA (see `.classify.taxonomy` for our LCA-based functions -- this would require having the taxonomy db `nodes.dmp` available)
3. emit one row per read with the LCA taxon as the result

This is revealing that the output of centrifuger (by default) is semantically quite different from some of our other read classifier tools in that it's not giving just one answer...

dpark01:
Actually if `-k 1` was in fact just an LCA anyway, is there a reason we wouldn't always want (or default to want) `-k 1`?

carze:
> Actually if `-k 1` was in fact just an LCA anyway, is there a reason we wouldn't always want (or default to want) `-k 1`?

Yea that's where I am a little confused as `-k 1` uses the following function to reduce down to one taxID: https://github.com/mourisl/centrifuger/blob/aa1e0640187d477812eb8836f294a38b76ea95ab/Taxonomy.hpp#L763

Which kinda sounds like LCA to me? I think we could definitely set `k` equal to 1 as our default and allow the user to change it if possible but if its done that can disable the kraken2-output from the run?

dpark01:
Agentic analysis by Claude Opus 4.7 via Claude Code, comparing the centrifuger source at the commit pinned in this PR to current `main` in viral-ngs:

---

**TL;DR:** `k=1` in `ReduceTaxIds` is computing an LCA, and for the common case it returns the same node our `coverage_lca` would with `lca_percent=100`. But the two algorithms differ in *how* they identify "common" and can diverge on edge cases.

### What each does

**centrifuger `ReduceTaxIds(..., k)` (bottom-up, rank-bucketed)** — [Taxonomy.hpp#L763](https://github.com/mourisl/centrifuger/blob/aa1e0640187d477812eb8836f294a38b76ea95ab/Taxonomy.hpp#L763)

- For each input taxon, walks parent pointers to root, dropping each ancestor into a bucket keyed by *rank-number* (a compressed integer derived from `_taxRankNum[]`, where sub/super-variants fold into their canonical rank).
- Nodes whose rank is `RANK_UNKNOWN` (i.e. NCBI "no rank") are **skipped** — they never enter a bucket.
- Picks the lowest rank-number whose bucket has `≤ k` distinct entries; with `k=1` that's the deepest rank where every lineage has collapsed to a single ancestor → LCA at the lowest canonical rank.
- Falls back to root if nothing converges.

**viral-ngs `coverage_lca(query_ids, parents, lca_percent=100)`** — [taxonomy.py#L85](https://github.com/broadinstitute/viral-ngs/blob/main/src/viral_ngs/classify/taxonomy.py#L85)

- For each query, builds the full root→leaf path (every node, regardless of rank).
- Descends top-down level by level; at each path-depth level takes the single most-common node; keeps going while its count ≥ `lca_needed` (= 100% of inputs when `lca_percent=100`).
- Returns the last node that satisfied the threshold → LCA at whatever tree node the lineages happen to share deepest.

### Where they agree

When the deepest common ancestor sits at a canonical NCBI rank (species/genus/family/…), both return the same taxon. That's the vast majority of NCBI lineages, so in practice you'd expect matching outputs for `k=1` vs. `lca_percent=100`.

### Where they can disagree

1. **"no rank" deepest ancestor.** NCBI is full of unranked clades (e.g. `cellular organisms`, virus group nodes, "unclassified X"). If two inputs' deepest shared node has `rank=no rank`, `coverage_lca` returns that node; centrifuger skips it and the result lands on its nearest canonical ancestor above (or, if rank buckets above also disagree, climbs further). Centrifuger's answer will be **less specific** in that case.
2. **Generalization.** `k>1` in centrifuger has no analog in `coverage_lca` — it returns a *set* of up-to-k promoted ancestors. Conversely, `lca_percent<100` (majority-based LCA) has no analog in centrifuger.
3. **Traversal model.** Bottom-up rank-bucketing vs. top-down path descent. They reach the same LCA on well-formed lineages but the rank-bucket fill trick (each empty intermediate rank gets stuffed with the current node `t`) is centrifuger's way of tolerating lineages of unequal depth — `coverage_lca` instead tolerates that implicitly by letting shorter paths drop out of the count, which is why `lca_percent=100` still produces the right answer when one input is an internal node above another.

Same idea, same answer in the common case, but not interchangeable as primitives — particularly around no-rank nodes and the `k>1` / `lca_percent<100` knobs.

dpark01:
So.. either way. Set `-k 1` is super fast and easy. Longer term though, I do think a lot of the useful taxonomy lies in unranked nodes -- I mean most interesting viruses live at nodes that aren't strictly `species`. Maybe we just do quick and easy now and note it as an enhancement for later.

carze:
Maybe before we go ahead and make this change I should run centrifuger with `-k 1` on some samples we previously analyzed and compare the results? I'm going to guess in most cases here we won't see much of a change but might be a good idea to do so just in case.

dpark01:
> Maybe before we go ahead and make this change I should run centrifuger with `-k 1` on some samples we previously analyzed and compare the results? I'm going to guess in most cases here we won't see much of a change but might be a good idea to do so just in case.

I predict that the difference is that most of the classification will basically look similar but at coarser levels of taxonomy. Like they'll all bubble up to the `species` level, and failing that, the `genus` level, etc... that might matter more for some bugs than others. Like all enterovirus A subtypes would get rolled up into `Enterovirus alphacoxsackie`. More problematic, all polio viruses would get rolled up into entero C (along with all the other non-polios).

dpark01 (approved):

dpark01:
Ok this will need to wait for #1077 to merge to fix the above CI issues. Then add a new CLI entry point in `metagenomics.py` for the native -> kraken style read tsv conversion.

PR #1073, opened 2026-05-20
Feature/genomad integration

carze:
## Summary

Adds [geNomad](https://portal.nersc.gov/genomad/) (Camargo et al., 2023) as a viral and plasmid identification tool in the classify subpackage. geNomad uses a machine-learning + marker-gene hybrid approach to classify sequences as viral, plasmid, or chromosomal — complementing the existing read-based classifiers (kraken2, kma, kb) with a sequence-based identifier.

Exposes one new CLI subcommand under `viral_ngs metagenomics`:

| Command | Purpose |
|---|---|
| `genomad` | Run geNomad end-to-end on a FASTA, producing viral/plasmid classifications |

### Tooling
- **Conda dep**: `genomad>=1.11.0` added to `docker/requirements/classify.txt`. Bioconda has `linux-aarch64` builds, so no `classify-x86.txt` entry needed.
- **`tensorflow<2.19` pin** (the big one) — see [Dependency fix](#dependency-fix) below.
- **Tool wrapper**: `src/viral_ngs/classify/genomad.py` — `Genomad(core.Tool)` with `end_to_end()`, `version()`, and `execute()` methods. Follows the kraken2/centrifuger wrapper pattern (`PrexistingUnixCommand` install method, early input validation, empty-FASTA short-circuit).
- **Input validation**: `end_to_end()` raises `FileNotFoundError` on missing FASTA, `ValueError` on missing/non-directory database path, and silently skips invocation on empty/whitespace-only FASTA inputs.
- **CLI**: snake_case args — positional `in_fasta`, `database`, `out_dir`, plus `--threads` from `cmd.common_args`.

### Dependency fix
The conda-forge `tensorflow=2.19.1` build ships `libtensorflow_cc.so.2` with `PT_GNU_STACK` marked executable. glibc on hardened kernels (including CI runners) refuses to load such libraries with:

```
ImportError: libtensorflow_cc.so.2: cannot enable executable stack
as shared object requires: Invalid argument
```

This breaks every `import tensorflow` — and therefore every `import genomad` — on the classify image. Reproduced on `quay.io/broadinstitute/viral-ngs:main-classify` after `micromamba install -c conda-forge tensorflow=2.19`.

Pinning `tensorflow<2.19` makes micromamba resolve to `2.18.0` on both `linux-64` and `linux-aarch64` (the most recent build without the broken flag). The pin is documented inline in `classify.txt` with the failure mode, affected kernels, and revisit condition (upstream conda-forge packaging bug).

## Testing

Verified in the classify container with geNomad + the tensorflow pin applied:

```
docker run --rm -v $(pwd):/opt/viral-ngs/source \
  quay.io/broadinstitute/viral-ngs:main-classify bash -c \
  "micromamba install -y -c bioconda -c conda-forge 'genomad>=1.11.0' 'tensorflow<2.19' && \
   pip install -e /opt/viral-ngs/source --quiet && \
   pytest -rsxX -n auto \
     /opt/viral-ngs/source/tests/unit/classify/test_genomad.py \
     /opt/viral-ngs/source/tests/unit/classify/test_integration_genomad.py"
```

Result: **10 passed**.

Test coverage:
- **Unit (mocked subprocess)** — 8 tests covering: CLI arg construction and ordering, output dir creation, missing-FASTA `FileNotFoundError`, invalid-DB `ValueError`, empty-FASTA short-circuit, `version()` output parsing, direct `main_genomad` dispatch, and end-to-end `parser_genomad → args.func_main → Genomad.end_to_end` plumbing.
- **Integration / smoke** — 2 tests gated by `pytest.mark.skipif(not shutil.which('genomad'))`. They run `genomad --version` and `genomad end-to-end --help` to exercise the full native + Python import chain (tensorflow, libtensorflow_cc, pyrodigal-gv, numpy) without needing the multi-GB geNomad database. These specifically catch the class of regression that motivated the tensorflow pin.

**Skip behavior verified**: with the `genomad` binary hidden from PATH, both smoke tests skip cleanly with `"geNomad executable is not installed"` instead of failing.

## Compatibility

- Multi-arch (x86_64 + ARM64) — both `genomad>=1.11.0` and `tensorflow<2.19` (resolves to 2.18.0) are available for `linux-aarch64` via conda-forge/bioconda.
- No changes to existing classify tool behavior.
- No new top-level Python dependencies (`pyproject.toml` unchanged).

**Feature implemented using Opus 4.7 and GPT5.5**


src/viral_ngs/classify/genomad.py
```diff
@@ -0,0 +1,118 @@
...
+            return True
+
+        # For small files, check content
+        if file_size <= 1024:
+            with file.open_or_gzopen(fasta_file, 'rt') as f:
+                content = f.read()
+                if content.strip() == '':
+                    return True
+
+        # For files >1024 bytes, assume non-empty for performance
```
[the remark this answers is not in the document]
dpark01: Regarding this Copilot complaint -- I'm not concerned about its unusual edge case that it's worried about. On a different note though, this is the third instance of otherwise already-duplicated code (from core.samtools and core.picard)... the original sin was committed earlier (probably by me) but maybe this third time let's make the world a better place and pull out into core.file or something as a shared helper function for all three of them to call?

src/viral_ngs/classify/genomad.py
```diff
@@ -0,0 +1,118 @@
...
+        if not os.path.isdir(db_path):
+            raise ValueError(f"Database path does not exist or is not a directory: {db_path}")
+
+        # Create output directory
+        file.mkdir_p(out_dir)
+
+        # Check for empty input FASTA
+        if self._is_fasta_empty(in_fasta):
+            log.warning("Input FASTA is empty, skipping geNomad")
+            return
```
dpark01, line 137: So, are we skipping geNomad because geNomad crashes on empty input? If so, is an empty output directory the semantically correct representation of empty output? I think it is, because according to https://github.com/apcamargo/genomad#understanding-the-outputs I think they're saying it emits twelve files per input sequence in the fasta so... zero files would be correct?

docker/Dockerfile.classify
dpark01: This might be belt and suspenders, since `from viral_ngs import metagenomics` kind of already does this, but it might be good to explicitly check that importing genomad works here again since that was what didn't work with the broken tensorflow package.

```suggestion
    python -c "from viral_ngs.classify import blast, bmtagger, genomad, kraken2, krona, kma, kmc, last; print('Classify tools OK')"
```

src/viral_ngs/classify/genomad.py
```diff
@@ -0,0 +1,118 @@
...
+        if file_size <= 1024:
+            with file.open_or_gzopen(fasta_file, 'rt') as f:
+                content = f.read()
+                if content.strip() == '':
+                    return True
+
+        # For files >1024 bytes, assume non-empty for performance
+        return False
+
+    def end_to_end(self, in_fasta, db_path, out_dir, num_threads=None):
```
dpark01: Are there no real knobs to expose here? I can't seem to access the real documentation (website down) but the example invocations on github talk about `--cleanup` and `--splits`

src/viral_ngs/metagenomics.py
```diff
@@ -1369,6 +1370,21 @@ def kb_build(ref_fasta, index, workflow='standard', kmer_len=31, protein=False,
...
+    parser.add_argument('out_dir', help='Output directory for geNomad results.')
+    cmd.common_args(parser, (('threads', None), ('loglevel', None), ('version', None), ('tmp_dir', None)))
+    cmd.attach_main(parser, main_genomad, split_args=True)
+    return parser
+def main_genomad(in_fasta, database, out_dir, threads=None):
+    '''
+        Classify viral and plasmid sequences using geNomad
+    '''
+    genomad_tool = genomad.Genomad()
+    genomad_tool.end_to_end(in_fasta, database, out_dir, num_threads=threads)
```
dpark01: So according to the github for genomad, it looks like it will automatically handle fasta and compressed (gz, bz2, or xz) fastas -- though not fasta.zst nor fastq nor bam. Is there a world in which we'd want our wrapper here to do fastq or bam -> fasta conversions for genomad? Or out of scope for this PR?

docker/requirements/classify.txt
```diff
@@ -2,8 +2,15 @@
 # Note: x86-only packages (bmtagger, kallisto, kb-python) are in classify-x86.txt
 anndata>=0.10
 blast>=2.15.0
+genomad>=1.11.0
+# Pin <2.19: conda-forge tensorflow 2.19.1 ships libtensorflow_cc.so.2
+# with PT_GNU_STACK marked executable, which glibc refuses to load on
+# hardened kernels (ImportError: "cannot enable executable stack as
+# shared object requires"). Affects any genomad import on the classify
+# image. Revisit when the conda-forge packaging bug is fixed upstream.
+tensorflow<2.19
```
dpark01, line 11: I suppose there's a world in which we could just `chmod -x` in the Dockerfile itself post-install, but if there's no pressing need for a newer version of tensorflow, this works just fine

src/viral_ngs/metagenomics.py
carze: geNomad typically works on contigs so I wouldn't expect to see fastqs or BAM files as an input here. I guess technically you could expect geNomad to work on long reads as well (could be contig-like) so we may see fastq or BAMs from that perspective but this pipeline isn't being advertised to have out of the box support for long reads. 

I think we maybe note this and hold off until a later date?

docker/requirements/classify.txt
carze: Dropping the tensorflow version down to 2.18 does impact other dependencies in the Dockerfile but running the testing suite seems to indicate that this is not an issue. This is the only thing that caused me to pause when pinning to a version below 2.19.

src/viral_ngs/classify/genomad.py
carze: Good point, because end-to-end usually runs with no parameters I didn't expose anything but to support smaller compute resources it does make sense to expose many parameters as well as for anyone who wants to tweak detection sensitivity and other related parameters. Adding this in now.

src/viral_ngs/classify/genomad.py
carze: Let me double check the behavior here but I do think that upon no output being generated no files are generated.

carze: Passing an empty FASTA file causes an error and one file is written, an annotate.log file as the first step geNomad attempts is to identify viral genes. 

So I suppose we could allow this to go through and it would just throw an error and generate this log file.

dpark01: Nah...
 - my read of the docs is that the directory should have one set of files per input sequence (where the prefix of the filenames is derived from the fasta header id of the sequence)
 - but if you're saying that passing an empty file causes a log file to be generated, that violates the contract that the output directory should really have zero files in it (like what would that log file be named? all answers are wrong)

PR #1078, opened 2026-05-22
Feature/virnucpro integration

carze:
Adds VirNucPro-adjacent post-processing helpers for converting VirNucPro chunk/contig outputs into human-readable TSV outputs that can be consumed downstream without running the GPU-heavy VirNucPro classifier inside the standard classify image. Exposes two CLI subcommands under `viral_ngs metagenomics`:

| Command | Purpose |
|---|---|
| `virnucpro_contigs` | Collapse VirNucPro chunk-level `highestscore.tsv` output into per-contig classifications |
| `virnucpro_label_reads_by_contig` | Label reads with the VirNucPro classification of their best-mapping contig using augmented minimap2 PAF |

## Changes

- **Conda dep**: `duckdb>=1.1.0,<2.0` added to `docker/requirements/classify.txt`. The `<2.0` cap avoids silent SQL behavior drift in the DuckDB-backed read-by-contig classifier.
- **Post-processing helpers**: `src/viral_ngs/classify/virnucpro.py` adds `classify_contigs()` and `classify_reads_by_contig()` helpers. This intentionally does **not** wrap the main GPU-accelerated VirNucPro classifier, which should run in its own Docker image.
- **Contig classification**: `classify_contigs()` summarizes VirNucPro chunk-level highest-score rows into sorted per-contig calls with confidence tiers and supporting metrics.
- **Read classification**: `classify_reads_by_contig()` joins augmented PAF alignments to per-contig classifications with DuckDB, selects the best primary alignment per read, flags reads that map across distinct calls as `Multi-mapped`, and writes TSV output.
- **Compressed output**: read-level outputs are written through `open_or_gzopen`, including `.gz`, `.zst`, `.lz4`, and `.bz2` support.
- **Memory control**: DuckDB memory limits can be passed explicitly via `--duckdb-memory-limit`; otherwise the helper auto-detects the cgroup memory limit and uses ~75% as a cap.
- **CLI**: two new subcommands in `src/viral_ngs/metagenomics.py`, with both camelCase and kebab-case option aliases where useful for WDL compatibility/readability.
- **Docs build cleanup**: `cmd.attach_main()` now cleans command docstring indentation before attaching parser descriptions, preventing Sphinx/docutils rendering issues from indented function docstrings.
- **Utility fix**: text-mode `.zst` writing in `open_or_gzopen()` now uses the zstd stream writer correctly and flushes the text wrapper.
- **Tests**: adds unit coverage for contig classification, read-by-contig labeling, compressed output, DuckDB memory-limit handling, CLI dispatch, docstring cleanup, and `.zst` text writing.

## Design Notes

This PR deliberately keeps the VirNucPro classifier itself out of `metagenomics.py`. The main VirNucPro classification task is GPU-accelerated and should run in a dedicated Docker image. The code here is limited to CPU-friendly post-processing utilities that fit the existing `viral-ngs` classify image design.

An earlier SIGTERM temp-dir cleanup approach was reverted. The final implementation relies on normal `TemporaryDirectory` cleanup and avoids extra signal-handling complexity.

## Testing

Local preflight:

```text
py_compile on changed Python files
git diff --check main...HEAD
```

Focused Docker test against the current `main-classify` image:

```text
53 passed, 1 skipped, 4 failed
```

The 4 failures are expected in the current `main-classify` image because that image does not yet include the new `duckdb` dependency. CI should validate the DuckDB-backed tests against the branch-built classify image after `docker/requirements/classify.txt` is applied.

Test coverage:

- **VirNucPro helpers** — contig call generation, missing-column validation, unmatched-ID validation, read-level labeling, fractional identity/coverage scale detection, empty PAF output, compressed `.zst` output, missing classification columns.
- **DuckDB memory behavior** — cgroup v2 detection, cgroup v1 fallback, unlimited/malformed/zero limits, explicit caller override, empty-string opt-out.
- **CLI plumbing** — `virnucpro_contigs` direct wrapper delegation and `virnucpro_label_reads_by_contig` parser → `args.func_main` dispatch.
- **Core utilities** — docstring dedent behavior in `cmd.attach_main()` and text-mode `.zst` writing through `open_or_gzopen()`.

## Compatibility

- Keeps the GPU VirNucPro classifier outside the standard classify command module.
- Adds one classify-image dependency: `duckdb>=1.1.0,<2.0`.
- No changes to existing Kraken2, Centrifuger, KMA, KB, Krona, or geNomad behavior.
- Outputs remain TSV-first and human-readable, with optional compression inferred from output extension.

**Feature created utilizing GPT-5.5**


dpark01:
Quick question before I look through this -- is the omission of the GPU path because of concern of size bloat or dependency graph complications in the resulting image?

carze:
Both but mainly size. CUDA is quite heavy and I already had to shave a lot off of the docker container I created for VirNucPro.

dpark01:
Maybe, like our BEAST/BEAGLE one, we make a standalone docker for CUDA-enabled virnucpro sometime...

carze:
Yea I actually made one and modeled it after the BEAST/BEAGLE image: https://github.com/broadinstitute/virnucpro-cuda

This should be called by the viral-pipelines side of things.

dpark01:
Ok some big picture questions on this PR. This PR introduces two new CLI commands in `metagenomics.py` One of them runs VNP on contigs. The other "labels reads". But there's some unclearness on how it does that. It seems to require
 1. the output of VNP on contigs (that part is easy to understand)
 2. "augmented PAF output" (not sure how this is done... can we expose a new CLI command that does this step for us / invokes the right arguments etc?) -- also .. must it be augmented PAF or can it just be BAM (because then it could be more modular / we could pick any aligner of our preference at the time)
 3. the output is not described/specified... is it a kraken-style per-read classification tsv? Because then we could use it in all sorts of things downstream.

docker/requirements/classify.txt
```diff
@@ -2,6 +2,12 @@
 # Note: x86-only packages (bmtagger, kallisto, kb-python) are in classify-x86.txt
 anndata>=0.10
 blast>=2.15.0
+# Cap below 2.0 so a future major-version release does not silently
+# change SQL behavior in the VirNucPro read-by-contig classifier
+# (which uses EXCLUDE, window functions, and read_csv with the
+# explicit columns= arg). Revisit the pin when DuckDB 2.0 ships and
+# the SQL paths have been validated against it.
+duckdb>=1.1.0,<2.0
```
dpark01, line 10: I could be okay with leaving this here, because the only dependency on it so far is here... but part of me feels like this could be promoted to `core.txt` as it is very general. But maybe we wait until the second use case to emerge... then we "LCA" the dependency up to the right docker image.

src/viral_ngs/classify/virnucpro.py
```diff
@@ -0,0 +1,701 @@
...
+    if missing:
+        raise ValueError("Missing required columns: {}".format(", ".join(missing)))
+
+    ids = df[id_col].astype(str)
+    df["ID"] = ids.str.replace(r"_chunk_\d+$", "", regex=True)
+    df["_group"] = ids.str.extract(id_pattern, expand=False)
+
+    n_unmatched = int(df["_group"].isna().sum())
+    if n_unmatched == len(df):
+        raise ValueError("No valid IDs extracted. Check id_pattern.")
```
dpark01, line 178: So... if no valid IDs extract... if the VNP tool emits no hits.. if the VNP tool was presented with empty input, these would all `raise ValueError`. Shouldn't we instead silently write a (semantically) empty output file?

src/viral_ngs/classify/__init__.py
```diff
@@ -11,3 +11,4 @@
 from . import krona
 from . import last
 from . import taxonomy
+from . import virnucpro
```
dpark01, line 14: perhaps we can also add a build-time validation of import success to the end of the (currently untouched) `Dockerfile.classify` file where it already does a build-time import test for a few other things??

src/viral_ngs/classify/virnucpro.py
```diff
@@ -0,0 +1,701 @@
...
+    if group["confidence"].sum() > 0:
+        weighted_delta = (
+            group["delta"] * group["confidence"]
+        ).sum() / group["confidence"].sum()
+    else:
+        weighted_delta = group["delta"].mean()
+
+    confident_viral = (group["max_score_1"] > 0.8) & (group["max_score_0"] < 0.3)
+    confident_nonviral = (group["max_score_0"] > 0.8) & (group["max_score_1"] < 0.3)
+    ambiguous = (group["max_score_1"] > 0.7) & (group["max_score_0"] > 0.7)
```
dpark01: Can we expose all these hardcoded cutoffs as defaulted/argparse-parameterized knobs that can be turned by the caller?

carze:
> Ok some big picture questions on this PR. This PR introduces two new CLI commands in `metagenomics.py` One of them runs VNP on contigs. The other "labels reads". But there's some unclearness on how it does that. It seems to require

Yea thats my fault, I think I kinda made a PR and a half. Will explain below

> 1. the output of VNP on contigs (that part is easy to understand)

Yup this is the default behavior from VNP, produce scores on a piece of sequence, typically contigs.

> 2. "augmented PAF output" (not sure how this is done... can we expose a new CLI command that does this step for us / invokes the right arguments etc?) -- also .. must it be augmented PAF or can it just be BAM (because then it could be more modular / we could pick any aligner of our preference at the time)

The augmented PAF is something that slipped by me when I was copying over my work (or re-building it) from the monolithic branch. It was added as a BamToPAF utility function in `tasks_read_util.wdl`:

 ```wdl
   paftools.js sam2paf ... > sample.paf

   # Add BLAST-like identity and query coverage
   id = $10/$11
   qcov = ($4-$3)/$2
   ...
   print standard_12_paf_fields, tags, id, qcov
   gzip sample.paf
```

This is somewhat of the hackiness we are trying to get away from in this refactor, I went with PAF because it was a tabular format already for us out of minimap2 that we could use for the next PR (hence the PR and a half). The next PR is to add a utility component that takes a minimap2 alignment of reads to each contig that VNP scored and to classify the reads that compose that contig as viral or non-viral. I got a little to trigger happy with this PR and added one piece of that already hence the confusion. We may want to pull it out and have that as a separate PR.

> 3. the output is not described/specified... is it a kraken-style per-read classification tsv? Because then we could use it in all sorts of things downstream.

The output is in a custom VirNucPro format that looks like the following:

highest_scores.tsv
```
Modified_ID     Is_Virus        max_score_0     max_score_1
NODE_1_length_15270_cov_1345.121011_chunk_5     True    0.0095721846446394      0.9904278516769408
NODE_1_length_15270_cov_1345.121011_chunk_6     True    0.1221813708543777      0.8778186440467834
NODE_1_length_15270_cov_1345.121011_chunk_7     True    0.342418372631073       0.657581627368927
NODE_1_length_15270_cov_1345.121011_chunk_10    True    0.0613824948668479      0.9386175274848938
NODE_1_length_15270_cov_1345.121011_chunk_12    True    0.2075118273496627      0.7924882173538208
NODE_1_length_15270_cov_1345.121011_chunk_13    True    0.1459545046091079      0.8540454506874084
NODE_1_length_15270_cov_1345.121011_chunk_17    True    0.1032852232456207      0.8967147469520569
NODE_1_length_15270_cov_1345.121011_chunk_18    False   0.6600467562675476      0.4333395957946777
NODE_1_length_15270_cov_1345.121011_chunk_19    False   0.7990233302116394      0.2735124230384826
```

scores.tsv
```
Sequence_ID     Prediction      score1  score2
NODE_1_length_15270_cov_1345.121011_chunk_5F1   virus   0.009572184644639492    0.9904278516769409
NODE_1_length_15270_cov_1345.121011_chunk_6F2   virus   0.12218137085437775     0.8778186440467834
NODE_1_length_15270_cov_1345.121011_chunk_7F3   virus   0.342418372631073       0.657581627368927
NODE_1_length_15270_cov_1345.121011_chunk_10F2  virus   0.06138249486684799     0.9386175274848938
NODE_1_length_15270_cov_1345.121011_chunk_12F1  virus   0.20751182734966278     0.7924882173538208
NODE_1_length_15270_cov_1345.121011_chunk_13F2  virus   0.14595450460910797     0.8540454506874084
NODE_1_length_15270_cov_1345.121011_chunk_17F3  virus   0.10328522324562073     0.8967147469520569
NODE_1_length_15270_cov_1345.121011_chunk_18F1  others  0.5666604042053223      0.43333959579467773
NODE_1_length_15270_cov_1345.121011_chunk_18F2  others  0.6600467562675476      0.33995321393013
```

Because VNP will fragment contigs and score each individual fragment, the highest score output picks the highest scoring fragment and reports that one. This is why we have two output files from it. I end up using the `scores.tsv` file downstream as I have my own heuristic to call a full contig viral based off the fragments and their scores.

File: tests/unit/classify/test_virnucpro.py [hunk 5]
```diff
@@ -0,0 +1,319 @@
...
+    highestscore_tsv.write_text("Modified_ID\tmax_score_0\nNODE_1_chunk_1\t0.1\n")
+
+    with pytest.raises(ValueError, match="Missing required columns"):
+        virnucpro.classify_contigs(
+            str(highestscore_tsv),
+            str(tmp_path / "contigs.tsv"),
+        )
+
+
+def test_classify_contigs_rejects_unmatched_ids(tmp_path):
```
dpark01, line 169: can we also add test coverage for empty input to `classify_contigs` (where the highestscore.tsv is header-only, no data)?

src/viral_ngs/classify/virnucpro.py
carze: Yes, I left this in here when I was meaning to either set them as constant and/or also have them exposed. Will do that for sure.

src/viral_ngs/classify/virnucpro.py
carze: Let me check what the behavior is here, I'm guessing that empty hits isn't actually a scenario where we'd want to say that the task failed.

dpark01:
> The augmented PAF is something that slipped by me when I was copying over my work (or re-building it) from the monolithic branch. [.....] I went with PAF because it was a tabular format already for us out of minimap2 that we could use for the next PR (hence the PR and a half). The next PR is to add a utility component that takes a minimap2 alignment of reads to each contig that VNP scored and to classify the reads that compose that contig as viral or non-viral. I got a little to trigger happy with this PR and added one piece of that already hence the confusion. We may want to pull it out and have that as a separate PR.

Sure -- though ... if we make a utility component that takes `<VNP per-contig outputs> + <aligned reads to contigs> = <reports per read>`, why would we hardcode minimap2 and custom PAF specifically -- shouldn't we just take any aligned BAM as input for the second input and go from there?

> > 3. the output is not described/specified... is it a kraken-style per-read classification tsv? Because then we could use it in all sorts of things downstream.
> 
> The output is in a custom VirNucPro format that looks like the following:
> [highest_scores.tsv] [scores.tsv]
> 
> Because VNP will fragment contigs and score each individual fragment, the highest score output picks the highest scoring fragment and reports that one. This is why we have two output files from it. I end up using the `scores.tsv` file downstream as I have my own heuristic to call a full contig viral based off the fragments and their scores.

Sorry, that's the *input* to `main_virnucpro_label_reads_by_contig` -- what's the output tsv from that, that output format isn't VNP's that's just a format we made up ourselves right?

carze:
> Sorry, that's the _input_ to `main_virnucpro_label_reads_by_contig` -- what's the output tsv from that, that output format isn't VNP's that's just a format we made up ourselves right?

Ah sorry, mis-read your initial comment -- the output of this component will be a PAF-like alignment file but with the VNP scores attached to it. So TSV/tabular in format that captures the viral-ness of the subset of reads that formed contigs. Basically we generate a bunch of TSV files here and at the end I parse them all and squash them into one big parquet file.

Which actually had me debating if we should just output parquet directly from each of these components but it also takes away the human-readble-ness of outputs from tools that could be run outside of this one end-to-end pipeline.

src/viral_ngs/classify/virnucpro.py
dpark01: Yeah generally not... but in this case, we're causing the behavior (not the tool)

dpark01:
> Ah sorry, mis-read your initial comment -- the output of this component will be a PAF-like file but with the VNP scores attached to it. So TSV/tabular in format. Basically we generate a bunch of TSV files here and at the end I parse them all and squash them into one big parquet file.
> 
> Which actually had me debating if we should just output parquet directly from each of these components but it also takes away the human-readble-ness of outputs from tools that could be run outside of this one end-to-end pipeline.

Gotcha... I kind of like the idea of just having downstream processes take compressed tsv -> parquet later for various reasons.

Ok in this case, what information is VNP ultimately telling us: `is-virus` boolean, and some confidence scores on whether the boolean is right? Are there any other things we ought to be storing in the results?

Because we invented the tsv format for this output, can you just document the format in the docstring of the CLI command itself?

carze:
> Because we invented the tsv format for this output, can you just document the format in the docstring of the CLI command itself?

This is actually the output that VNP writes itself. The two scores files in which you get some sort of score and in the highest-scores file it does actually call a boolean (vs. the scores file it tags as viral or other). Yes a bit confusing in general but I just left the outputs as is -- I will document them.

dpark01:
> > Because we invented the tsv format for this output, can you just document the format in the docstring of the CLI command itself?
> 
> This is actually the output that VNP writes itself. The two scores files in which you get some sort of score and in the highest-scores file it does actually call a boolean (vs. the scores file it tags as viral or other). Yes a bit confusing in general but I just left the outputs as is -- I will document them.

Oh I see -- you're only just lifting over the outputs of VNP on contigs and changing the IDs (from contig IDs to read IDs) and replicating it a zillion times (for each read)? If that's the case, maybe just have it say so and maybe even point to the vnp docs describing that format.

carze:
> > > Because we invented the tsv format for this output, can you just document the format in the docstring of the CLI command itself?
> > 
> > 
> > This is actually the output that VNP writes itself. The two scores files in which you get some sort of score and in the highest-scores file it does actually call a boolean (vs. the scores file it tags as viral or other). Yes a bit confusing in general but I just left the outputs as is -- I will document them.
> 
> Oh I see -- you're only just lifting over the outputs of VNP on contigs and changing the IDs (from contig IDs to read IDs) and replicating it a zillion times (for each read)? If that's the case, maybe just have it say so and maybe even point to the vnp docs describing that format.

Exactly. Which is why I need the alignment step beforehand since we don't keep track of the reads that were used to create an individual contig. The VNP site itself doesn't do a good job capturing what its two outputs are so I can add information it as well as what the output from the classify reads step is as well.

src/viral_ngs/classify/virnucpro.py
carze: One thing here, I was not checking for empty header -> fail gracefully but this particular if statement is trying to understand what the sequence ID's look like for these assemblies (in this case we default to spades naming) in order to grab the root ID for the sequence. 

So raising an error here is more correct in the sense that we already know we have sequences but if the component can't pull out the root ID for the assembled sequence it can't look at all the fragments as a block to make its decision if the contig is viral or not. Not sure if that makes any sense in the way I described it but this to me is the correct behavior.

dpark01: Oh I see you're saying this code path only executes when the input is non-empty? Then I suppose there's nothing to do here especially if we add that other test coverage

carze:
I went ahead and moved our PAF-entrypoint to the classify-reads-by-contig future component to use a BAM file so this should require much less hacking upstream. Think this is about ready to go pending tests + manual review on your end.

dpark01:
Ok I wanted to get my head around the us-specific logic we were adding post-tool-call here and had Claude summarize the following (see below).

1. If this is an accurate depiction, this feels like it has just enough complexity & author choices/decisions baked into it that it seems worth some sort of additional documentation somewhere (more than just docstrings and argparse help strings).. I wonder if it could/ought to go in just some companion markdown file in the code repo near where the class sits... or in the readthedocs stuff under `docs/`... open to whatever, there isn't much precedent here, but it seems worth capturing in text somewhere.
2. Q1: was the SQL-based collapse by query_name intentional that it reduced to one hit per paired read?
3. Q2: currently a read with one alignment to Viral and another missing (aka Unclassified) becomes Multi-mapped, that doesn't seem right? Shuld we filter out Unclassified before counting distinct calls?
4. Q3: should `mapped_well` be used in any SQL filters or no?

---

# VirNucPro post-processing

This document describes the post-processing logic that `viral_ngs.classify.virnucpro` layers on top of the raw VirNucPro classifier output. It is purely descriptive: it explains what the code does today so that reviewers can evaluate the choices independently. It does **not** prescribe what the logic should be.

> **Scope.** The VirNucPro classifier itself (a GPU/LLM-based viral-vs-non-viral predictor) lives in a separate Docker image and is not part of this repository. This module only handles CPU-friendly post-processing of VirNucPro outputs.

---

## Pipeline overview

```
    contigs.fasta                                reads.fastq
         │                                            │
         ▼                                            ▼
   ┌─────────────┐                            ┌──────────────┐
   │  VirNucPro  │  (separate Docker image)   │   Aligner    │  (any aligner; minimap2, BWA, …
   │ (GPU/LLM)   │                            │              │   anything that emits NM tag)
   └─────────────┘                            └──────────────┘
         │                                            │
         │  highest_scores.tsv                        │  aligned.bam
         │  (one row per chunk)                       │  (reads aligned to contigs)
         ▼                                            │
   ┌─────────────────────────────┐                    │
   │   classify_contigs()        │  ◄────── Stage 1   │
   │   (this module)             │                    │
   └─────────────────────────────┘                    │
         │                                            │
         │  contigs.tsv                               │
         │  (one row per contig:                      │
         │   call + tier + metrics)                   │
         │                                            │
         └─────────────┬──────────────────────────────┘
                       ▼
              ┌──────────────────────────────────┐
              │   classify_reads_by_contig()     │  ◄────── Stage 2
              │   (this module)                  │
              └──────────────────────────────────┘
                       │
                       ▼
                 reads_classified.tsv
                 (one row per query_name from the BAM)
```

Two stages, two CLI subcommands:

| CLI subcommand                       | Helper                          | Stage                          |
|--------------------------------------|---------------------------------|--------------------------------|
| `virnucpro_contigs`                  | `classify_contigs`              | Chunk → contig roll-up         |
| `virnucpro_label_reads_by_contig`    | `classify_reads_by_contig`      | Contig → read label propagation|

---

## Background: what VirNucPro produces

VirNucPro chops each input contig into fixed-length non-overlapping chunks (the published benchmarking work uses 300 bp and 500 bp), runs a six-frame translation, scores each frame with an LLM, and emits two TSVs. We consume `highest_scores.tsv`, which has one row per chunk:

| column         | meaning                                                       |
|----------------|---------------------------------------------------------------|
| `Modified_ID`  | `{contig_id}_chunk_{N}` — N is sparse (chunks failing internal VirNucPro filters are omitted) |
| `Is_Virus`     | boolean; empirically equal to `max_score_1 > max_score_0`     |
| `max_score_0`  | non-viral class score (probability-like, 0–1)                 |
| `max_score_1`  | viral class score (probability-like, 0–1)                     |

`Is_Virus` is not consumed by this module — the two score columns carry strictly more information.

---

## Stage 1: chunk → contig (`classify_contigs`)

Takes a VirNucPro `highest_scores.tsv` and produces a per-contig TSV with a categorical `call`, a `tier` confidence label, and supporting metrics.

### Grouping chunks back into contigs

VirNucPro flattens the contig structure into chunk IDs like `NODE_1_length_15270_cov_X_chunk_5`. To re-aggregate, the helper does two things to the `Modified_ID` column:

1. **Contig label** — strip the trailing `_chunk_\d+` suffix. Used as the human-readable `ID` in the output.
5. **Group key** — match against `id_pattern` (default `(NODE_\d+)`) and use the matched group as the grouping key for `pandas.DataFrame.groupby`.

These two derivations are independent. With the default SPAdes-style pattern, the group key is the `NODE_N` prefix and the contig label is the full pre-chunk string (`NODE_N_length_X_cov_Y`). Rows whose ID does not match `id_pattern` are dropped with a warning. If **no** rows match, `ValueError` is raised. If the input file is empty or the table is header-only, an empty (header-only) output TSV is written and the helper returns without error.

### Per-chunk derived quantities

For each chunk in a contig group:

```
delta      = max_score_1 - max_score_0          # signed; positive = viral, negative = non-viral
confidence = sqrt(|delta|)                       # confidence weight for the per-contig average
```

Three orthogonal categorical labels are also computed at the chunk level, used later for tier assignment:

```
confident_viral    = (max_score_1 > min_confident_score) AND (max_score_0 < max_opposing_score)
confident_nonviral = (max_score_0 > min_confident_score) AND (max_score_1 < max_opposing_score)
ambiguous          = (max_score_1 > min_ambiguous_score) AND (max_score_0 > min_ambiguous_score)
```

A chunk can be **none** of the three (e.g. `max_score_1 = 0.6, max_score_0 = 0.4`), exactly one, or `ambiguous` plus possibly one of the confident flags (depending on parameter values). The default parameters (`min_confident_score=0.8`, `max_opposing_score=0.3`, `min_ambiguous_score=0.7`) make these flags effectively mutually exclusive.

### Per-contig aggregated quantities

```
weighted_delta = Σ(delta_i * confidence_i) / Σ confidence_i
                 (falls back to mean(delta) if all confidences are zero)

n_chunks                = total chunks in the group
n_confident_viral       = count of chunks where confident_viral is true
n_confident_nonviral    = count of chunks where confident_nonviral is true
n_ambiguous             = count of chunks where ambiguous is true
n_effective             = n_chunks - n_ambiguous
viral_proportion        = n_confident_viral    / n_effective    (0 if n_effective == 0)
nonviral_proportion     = n_confident_nonviral / n_effective    (0 if n_effective == 0)
```

### Call decision

A single threshold splits the three call categories:

```
if   weighted_delta >  min_weighted_delta : call = "Viral"
elif weighted_delta < -min_weighted_delta : call = "Non-viral"
else                                       : call = "Ambiguous"
```

So `call` is fully determined by `weighted_delta` and `min_weighted_delta` (default 0.3). The other parameters do not move contigs between call categories — they only affect `tier`.

### Tier decision

```
Ambiguous call:
    tier = "review"

Viral call:
    if n_confident_viral >= 1 AND viral_proportion >= min_viral_proportion:
        tier = "high_confidence"  if weighted_delta >  high_confidence_delta else "moderate_confidence"
    else:
        tier = "low_confidence"

Non-viral call:  (mirror of Viral; uses nonviral_proportion and the negative delta)
    if n_confident_nonviral >= 1 AND nonviral_proportion >= min_nonviral_proportion:
        tier = "high_confidence"  if weighted_delta < -high_confidence_delta else "moderate_confidence"
    else:
        tier = "low_confidence"
```

Finally, a **short-contig demotion** is applied:

```
if n_chunks < min_chunks:
    "high_confidence" or "moderate_confidence"  →  "low_confidence"
    "low_confidence"                            →  "review"
    # "review" is unchanged
```

### Parameters

| Parameter (helper kwarg / CLI flag)          | Default | Affects | Meaning |
|---------------------------------------------|---------|---------|---------|
| `min_weighted_delta` / `--min-weighted-delta` | 0.3     | **call** + tier | Min `|weighted_delta|` for a Viral/Non-viral call. Below: Ambiguous. |
| `high_confidence_delta` / `--high-confidence-delta` | 0.6 | tier | Split between `moderate_confidence` and `high_confidence`. |
| `min_confident_score` / `--min-confident-score` | 0.8  | tier | Winning-class score for `confident_viral`/`confident_nonviral` flags. |
| `max_opposing_score` / `--max-opposing-score` | 0.3   | tier | Max opposing-class score for the same. |
| `min_ambiguous_score` / `--min-ambiguous-score` | 0.7 | tier | Min score in **both** classes for the `ambiguous` flag (and `n_effective` denominator). |
| `min_viral_prop` / `--min-viral-prop`        | 0.1     | tier | Min `viral_proportion` for non-`low_confidence` Viral tiers. |
| `min_nonviral_prop` / `--min-nonviral-prop`  | 0.1     | tier | Mirror for Non-viral. |
| `min_chunks` / `--min-chunks`                | 5       | tier | Below this, tier is demoted one notch. |
| `id_col` / `--id-col`                        | `"Modified_ID"` | grouping | Source column for chunk IDs. |
| `id_pattern` / `--id-pattern`                | `(NODE_\d+)` | grouping | Regex group used to roll chunks back into contigs. |

### Output schema (`classify_contigs`)

| column                 | type   | notes |
|------------------------|--------|-------|
| `ID`                   | string | chunk-suffix-stripped contig label (e.g. `NODE_1_length_15270_cov_X`) |
| `call`                 | string | one of `Viral`, `Non-viral`, `Ambiguous` |
| `tier`                 | string | one of `high_confidence`, `moderate_confidence`, `low_confidence`, `review` |
| `weighted_delta`       | float  | rounded to 3 decimals |
| `n_chunks`             | int    | |
| `n_confident_viral`    | int    | |
| `n_confident_nonviral` | int    | |
| `n_ambiguous`          | int    | |
| `viral_proportion`     | float  | rounded to 3 decimals |
| `nonviral_proportion`  | float  | rounded to 3 decimals |

Rows are sorted by `ID` with a natural-sort key (so `NODE_2` precedes `NODE_10`).

---

## Stage 2: contig → read (`classify_reads_by_contig`)

Takes (a) an aligned BAM of reads-to-contigs, and (b) the per-contig TSV from Stage 1, and emits a per-read TSV with each read's call/tier propagated from its best-mapping contig.

### Step A: BAM filtering and per-alignment augmentation

`_prepare_augmented_bam_file` iterates the BAM with `pysam.AlignmentFile` and emits one row per **retained** alignment to a normalized intermediate TSV. Rows are filtered out **before** they reach DuckDB:

- `record.is_unmapped` (flag `0x4`)              → dropped
- `record.is_secondary` (flag `0x100`)           → dropped (counted as `n_secondary`)
- `record.is_supplementary` (flag `0x800`)       → dropped (counted as `n_secondary`)

Everything else (i.e. **primary mapped alignments**) is emitted. The flow does not inspect the paired-end flags (`0x1`, `0x40`, `0x80`); R1 and R2 of a pair both pass the filter if both are primary mapped, and they appear as two separate rows that share `query_name`.

For each retained alignment, the helper validates and computes:

| field                  | derivation |
|------------------------|------------|
| `source_order`         | 1-based counter assigned in BAM iteration order (final tiebreaker for best-alignment selection) |
| `query_name`           | `record.query_name` |
| `query_length`         | `record.infer_read_length()` → `infer_query_length(always=True)` → `record.query_length` (first non-None) |
| `query_start`/`_end`   | `record.query_alignment_start` / `record.query_alignment_end` (excludes soft clips) |
| `strand`               | `'-'` if `record.is_reverse` else `'+'` |
| `target_name`          | reference name from `bam.get_reference_name(record.reference_id)` |
| `target_length`        | `bam.get_reference_length(target_name)` |
| `target_start`/`_end`  | `record.reference_start` / `record.reference_end` |
| `alignment_block_length` | sum of CIGAR op lengths for `M`, `I`, `D`, `=`, `X` |
| `num_matches`          | `alignment_block_length − NM` |
| `pct_identity`         | `100 × num_matches / alignment_block_length` (BLAST-like; indels in denominator) |
| `pct_query_cov`        | `100 × (query_end − query_start) / query_length` |
| `mapping_quality`      | `record.mapping_quality` |

Validation errors that raise `ValueError`:

- Missing `NM` tag on any retained primary alignment.
- `query_length` cannot be inferred or is zero.
- `reference_end` cannot be inferred.
- `alignment_block_length == 0`.
- `NM > alignment_block_length`.

### Step B: DuckDB pipeline

The normalized alignment TSV and a normalized copy of the contig classifications are loaded into an in-memory DuckDB connection. DuckDB memory is capped from the cgroup limit (`~75%` of the container's memory) unless overridden by `--duckdb-memory-limit`; passing an empty string opts out of any limit.

A `mapped_well` flag is computed but **not used to filter rows**:

```sql
mapped_well = (mapping_quality >= min_mapq
               AND pct_identity   >= min_identity
               AND pct_query_cov  >= min_query_cov)
```

The retained-primary alignments are LEFT-JOINed to the contig classifications. Contigs absent from the classifications table get a synthesized call:

```sql
COALESCE(c.call, 'Unclassified')   AS call
COALESCE(c.tier, '')               AS tier
COALESCE(c.<other_metrics>, 0)     AS <other_metrics>
```

This `merged` table has one row per retained primary alignment, with the contig's call attached (or `Unclassified`).

Two derived tables are then computed in parallel from `merged`:

**`call_counts`** — number of distinct call categories per `query_name`, across **all** rows in `merged`:

```sql
SELECT query_name, COUNT(DISTINCT call) AS n_distinct_calls
FROM merged
GROUP BY query_name
```

Notes on what this counts:
- It groups by `query_name` only. R1 and R2 of a paired-end pair share `query_name`, so their alignments both contribute.
- It treats `'Unclassified'` as a value alongside `'Viral'`, `'Non-viral'`, `'Ambiguous'` — so a read whose alignments touch one classified contig and one unclassified contig has `n_distinct_calls == 2`.
- It does **not** filter by `mapped_well` — sub-threshold alignments contribute to the count just like above-threshold ones.

**`best`** — single winning alignment per `query_name`:

```sql
ROW_NUMBER() OVER (
    PARTITION BY query_name
    ORDER BY mapping_quality DESC, pct_identity DESC, source_order ASC
) AS rn
WHERE rn = 1
```

Notes:
- Partitioning by `query_name` collapses paired-end mates (and any other case where one `query_name` has multiple primary alignments) down to a single row.
- The losing alignment(s) are dropped from the output entirely — they are not represented as separate rows.
- `source_order` (BAM iteration index) is the final tiebreaker, so two alignments with identical `mapping_quality` and `pct_identity` are resolved by which appeared first in the input BAM. Reordering the BAM between alignment and classification can change which mate "wins" in true ties.

### Step C: result composition

The final result joins `best` to `call_counts` and applies a `Multi-mapped` override:

```sql
call = CASE WHEN cc.n_distinct_calls > 1 THEN 'Multi-mapped' ELSE b.call END
tier = CASE WHEN cc.n_distinct_calls > 1 THEN 'review'       ELSE b.tier END
```

So the per-read `call` can take five values:

| value         | source |
|---------------|--------|
| `Viral`       | propagated from the winning contig's Stage-1 call |
| `Non-viral`   | propagated from the winning contig's Stage-1 call |
| `Ambiguous`   | propagated from the winning contig's Stage-1 call |
| `Unclassified`| synthesized when the winning contig is missing from the classifications table |
| `Multi-mapped`| synthesized when `n_distinct_calls > 1` for the read's `query_name` (overrides the four above) |

### Output schema (`classify_reads_by_contig`)

One row per `query_name` that had at least one retained primary alignment. Unmapped reads, secondary alignments, and supplementary alignments produce no output rows.

| column                 | source |
|------------------------|--------|
| `read_id`              | `query_name` of the winning alignment |
| `read_length`          | `query_length` of the winning alignment |
| `contig_id`            | `target_name` of the winning alignment |
| `contig_length`        | `target_length` of the winning alignment |
| `strand`               | from the winning alignment |
| `mapping_quality`      | from the winning alignment |
| `pct_identity`         | from the winning alignment |
| `pct_query_cov`        | from the winning alignment |
| `mapped_well`          | from the winning alignment (boolean) |
| `call`                 | one of the five values above |
| `tier`                 | from Stage 1, or `'review'` when overridden by Multi-mapped, or `''` for Unclassified |
| `weighted_delta`       | from Stage 1 (0 for Unclassified) |
| `n_chunks`             | from Stage 1 (0 for Unclassified) |
| `n_confident_viral`    | from Stage 1 (0 for Unclassified) |
| `n_confident_nonviral` | from Stage 1 (0 for Unclassified) |
| `n_ambiguous`          | from Stage 1 (0 for Unclassified) |
| `viral_proportion`     | from Stage 1 (0 for Unclassified) |
| `nonviral_proportion`  | from Stage 1 (0 for Unclassified) |

Output row order is whatever DuckDB's `COPY` emits (no explicit `ORDER BY`). The output extension governs compression: `.gz`, `.zst`, `.lz4`, `.bz2` are detected automatically; anything else is written uncompressed.

If the BAM yields zero retained primary alignments (empty BAM or all reads unmapped/secondary/supplementary), an empty header-only output TSV is written and the DuckDB pipeline is skipped.

### Parameters

| Parameter (CLI flag)                | Default     | Meaning |
|------------------------------------|-------------|---------|
| `--min-mapq`                       | 5           | Threshold for the `mapped_well` flag. Sub-threshold alignments are **not** dropped. |
| `--min-identity`                   | 90.0        | Percent identity threshold for `mapped_well`. Percent units (90, not 0.9); a fractional value 0 < x < 1 raises `ValueError`. |
| `--min-query-cov`                  | 80.0        | Percent query coverage threshold for `mapped_well`. Same percent-units rule. |
| `--duckdb-memory-limit`            | auto-detect | DuckDB memory cap (e.g. `"8GB"`). Default auto-detects ~75% of the cgroup limit. Empty string opts out of any cap. |
| `--work-dir`                       | system tmp  | Parent directory for the per-run scratch tree (used for the normalized TSVs and DuckDB spill). Distinct from the `--tmp_dir` flag provided by `cmd.common_args`. |

---

## Behavioral notes for reviewers

The following are factual statements about the implementation, not recommendations.

1. **Primary-alignment filter is at the pysam stage, not in SQL.** Secondary (`0x100`) and supplementary (`0x800`) alignments never reach DuckDB. The `ROW_NUMBER()` window therefore exists only to deduplicate `query_name` values whose primary alignments appear more than once in the BAM — i.e. paired-end mates or pathological multi-primary cases.

2. **Paired-end mates share `query_name` and are not distinguished.** Both the `PARTITION BY query_name` (winner selection) and the `GROUP BY query_name` (distinct-call count) treat R1 and R2 of a pair as alignments of the same logical read. There is no use of the BAM paired-end flags (`0x1`, `0x40`, `0x80`).

6. **`Unclassified` participates in the `Multi-mapped` count.** Because `COALESCE(c.call, 'Unclassified')` is applied before `COUNT(DISTINCT call)`, a `query_name` with one alignment to a classified contig and one to an unclassified contig has `n_distinct_calls == 2` and is labeled `Multi-mapped`.

7. **`mapped_well` is informational only.** It is computed and exposed as an output column, but no DuckDB query filters on it. Sub-threshold alignments contribute to both `best` selection (where high-quality alignments will still win on `mapping_quality DESC` ordering) and `call_counts` (where any call value they pull from the classifications table is counted as a distinct call).

8. **Output row ordering is not guaranteed.** No explicit `ORDER BY` on the terminal `COPY`.

9. **Default thresholds in percent units.** `min_identity` and `min_query_cov` validate against `0 < value < 1.0` and raise — so fractional thresholds cannot be silently passed. Zero and ≥1.0 are accepted.

carze:
> Ok I wanted to get my head around the us-specific logic we were adding post-tool-call here and had Claude summarize the following (see below).
> 
> 1. If this is an accurate depiction, this feels like it has just enough complexity & author choices/decisions baked into it that it seems worth some sort of additional documentation somewhere (more than just docstrings and argparse help strings).. I wonder if it could/ought to go in just some companion markdown file in the code repo near where the class sits... or in the readthedocs stuff under `docs/`... open to whatever, there isn't much precedent here, but it seems worth capturing in text somewhere.

Yea this seems like a good idea; perhaps in the `classify/` folder we can have a markdown with extra notes for each classifier broken up by classifier?

> 2. Q1: was the SQL-based collapse by query_name intentional that it reduced to one hit per paired read?

Yes; Kraken and Centrifuger collapse their reads down so we want to do the same with VNP and Kallisto to match that convention and have one entry per read-pair in our output table.

> 3. Q2: currently a read with one alignment to Viral and another missing (aka Unclassified) becomes Multi-mapped, that doesn't seem right? Shuld we filter out Unclassified before counting distinct calls?

So unclassified here could also mean that one of the other classifiers could not place it. Which are interesting scenarios for the non-reference based classifiers as this could be a novel hit. But I don't think multi-mapped is right here -- maybe something like putative viral would be better.

> 4. Q3: should `mapped_well` be used in any SQL filters or no?

This is actually used downstream in some of our analysis/figure generating scripts.

dpark01:
> Yea this seems like a good idea; perhaps in the `classify/` folder we can have a markdown with extra notes for each classifier broken up by classifier?

Yeah since we don't have much precedent, then that seems fine for now. I mean I could see putting it in `docs/` (but then it'd have to be restructuredtext) but maybe we just drop a markdown in there for now. The main thing is the logic and decisions *we* made (not the tool itself), so a lot of these terms and categories and scoring for VNP were decided by us. I think our tooling around kraken and others is not really that interesting to document (they're more mechanical and less logic flow). We could definitely use more thorough documentation in our assembly pipeline, as there are decisions and choices being made there that are of similar consequence to what we did here for VNP, but some other day / PR.

> > 3. Q2: currently a read with one alignment to Viral and another missing (aka Unclassified) becomes Multi-mapped, that doesn't seem right? Shuld we filter out Unclassified before counting distinct calls?
> 
> So unclassified here could also mean that one of the other classifiers could not place it. Which are interesting scenarios for the non-reference based classifiers as this could be a novel hit. But I don't think multi-mapped is right here -- maybe something like putative viral would be better.

Ok so then should we update this logic so that it doesn't produce multi-mapped if the read pair contains `{Viral, Unclassified}`? Honestly I think that should probably just be Viral.

carze:
> Ok so then should we update this logic so that it doesn't produce multi-mapped if the read pair contains `{Viral, Unclassified}`? Honestly I think that should probably just be Viral.

I keep jumping out of context here since unclassified and "multi-mapped" are utilized at multiple points in our pipeline.

For this context this would mean that a read mapped to both a contig that VirNucPro classified as viral and a contig that was marked as not viral by VNP. So here I do think that Unclassified could be somewhat justified but we also could try to look at specifics of the alignment itself to weight whether or not the viral alignment was stronger than the non-viral alignment.

dpark01:
> > Ok so then should we update this logic so that it doesn't produce multi-mapped if the read pair contains `{Viral, Unclassified}`? Honestly I think that should probably just be Viral.
> 
> I keep jumping out of context here since unclassified and "multi-mapped" are utilized at multiple points in our pipeline.
> 
> For this context this would mean that a read mapped to both a contig that VirNucPro classified as viral and a contig that was marked as not viral by VNP. So here I do think that Unclassified could be somewhat justified but we also could try to look at specifics of the alignment itself to weight whether or not the viral alignment was stronger than the non-viral alignment.

Wait, I thought `Unclassified` contig did not mean that "VNP classified as not viral" -- I thought it meant it wasn't sure what it was? If it was "VNP classified contig as not viral" then it should be `Non-viral` instead.

If a contig has some chunks viral and some unclassified -> that seems viral to me

carze:
> Wait, I thought `Unclassified` contig did not mean that "VNP classified as not viral" -- I thought it meant it wasn't sure what it was? If it was "VNP classified contig as not viral" then it should be `Non-viral` instead.
> 
> If a contig has some chunks viral and some unclassified -> that seems viral to me

Again confusing my context. So definitively if a read maps to a cluster that was not classified by VirNucPro it is tagged as `Unclassified` -- so our `Multi-mapped` tag is only added when a read simultaneously maps to both a contig VNP classifies as Viral and another contig that it didn't classify at all.

Sorry for all the confusion!

dpark01:
> > Wait, I thought `Unclassified` contig did not mean that "VNP classified as not viral" -- I thought it meant it wasn't sure what it was? If it was "VNP classified contig as not viral" then it should be `Non-viral` instead.
> > If a contig has some chunks viral and some unclassified -> that seems viral to me
> 
> Again confusing my context. So definitively if a read maps to a cluster that was not classified by VirNucPro it is tagged as `Unclassified` -- so our `Multi-mapped` tag is only added when a read simultaneously maps to both a contig VNP classifies as Viral and another contig that it didn't classify at all.
> 
> Sorry for all the confusion!

Right -- in that case though, shouldn't we just call it viral (not multi-mapped... what's the multi in this case)? All of the things the read pair mapped to were classified, if at all, as viral, and none of its mappings suggest otherwise.

carze:
> Right -- in that case though, shouldn't we just call it viral (not multi-mapped... what's the multi in this case)? All of the things the read pair mapped to were classified, if at all, as viral, and none of its mappings suggest otherwise.

I guess considering we are working with a tool that is not so battle-tested it might be better to err on the conservative side here since if we see two equivalent mappings, one to a viral contig and one that was not classified that would make me a little more hesitant to call the read viral.

dpark01:
> I guess considering we are working with a tool that is not so battle-tested it might be better to err on the conservative side here since if we see two equivalent mappings, one to a viral contig and one that was not classified that would make me a little more hesitant to call the read viral.

Just to clarify, when you say two mappings of a read, you mean two mappings of a read pair, right? (one read maps to one contig, its mate maps to another contig)? Because your pysam load already throws away all secondary, unmapped, non-primary alignments, but then coalesces on read ID, so the only time the `COUNT(DISTINCT call) AS n_distinct_calls` would be `> 1` is if one read and its mate are mapping to different contigs *and* those two contigs have different VNP calls?

carze:
> Just to clarify, when you say two mappings of a read, you mean two mappings of a read pair, right? (one read maps to one contig, its mate maps to another contig)? Because your pysam load already throws away all secondary, unmapped, non-primary alignments, but then coalesces on read ID, so the only time the `COUNT(DISTINCT call) AS n_distinct_calls` would be `> 1` is if one read and its mate are mapping to different contigs _and_ those two contigs have different VNP calls?

Yes this is correct -- if mates map to two different contigs and one of those contigs was not called by VNP this is where `Multi-mapped` is used. I would still be somewhat conservative here as I would think there would be a lot more confidence here if both mates mapped to the contig that was called `Viral`.

PR #1082, opened 2026-05-28
Feature/kallisto refactor

carze:
Adds a Kallisto/kb-python refactor that replaces the previous `kb` wrapper naming with `kallisto` and exposes TSV-first outputs for downstream WDL/Terra workflows. The refactor keeps `kb-python` as the execution backend, but moves the public viral-ngs interface to Kallisto-oriented command names and human-readable reports.

| Command | Purpose |
|---|---|
| `kallisto` | Run Kallisto/kb count and emit long-form `counts.tsv` plus internal H5AD |
| `kallisto_extract` | Run Kallisto/kb extract and emit read-level `summary.tsv` |
| `kallisto_build` | Build a Kallisto index from reference sequences |
| `kallisto_top_taxa` | Summarize top focal-taxon hits from long-form Kallisto `counts.tsv` |

## Changes

- **Component rename**: replaces the old `viral_ngs.classify.kb` wrapper with `viral_ngs.classify.kallisto`. This is intentionally a breaking refactor and does not preserve old `kb` command aliases.
- **CLI rename**: exposes Kallisto-oriented subcommands under `metagenomics`: `kallisto`, `kallisto_extract`, `kallisto_build`, and `kallisto_top_taxa`.
- **Counts TSV**: `kallisto` now writes long-form `counts.tsv` with `sample_id`, `db_hit_id`, and `count`, using the collapsed Kallisto/kb count matrix.
- **Read summary TSV**: `kallisto_extract` writes read-level `summary.tsv` with `SAMPLE_ID`, `READ_ID`, `DB_ID`, `TAXONOMY_LINEAGE`, `TAXONOMY_NAME`, and `SEQUENCE_LENGTH`.
- **Top taxa report**: `kallisto_top_taxa` consumes `counts.tsv` and writes a stable ranked TSV report for WDL/Terra scalar extraction.
- **Taxonomy map parsing**: supports CSV/TSV maps, headered and headerless files, duplicate ID columns, placeholder `.` values, and trailing strand/genome-type values such as `+ssRNA` or `-ssRNA`.
- **Sample IDs**: adds `--sample-id` support so count and extract outputs can carry workflow-level sample identifiers rather than relying only on input basenames.
- **Docker classify check**: adds `kallisto` to the classify Docker import verification.
- **Tests**: replaces kb-focused unit/integration coverage with Kallisto-focused tests and adds regression coverage for TSV generation, taxonomy parsing, CLI dispatch, top-taxa reporting, and extract summaries.

## Design Notes

This PR keeps `kb-python` as the backend implementation while renaming the viral-ngs-facing component to `kallisto`, matching the terminology used by downstream pipeline components.

The public output contract is TSV-first. H5AD is still produced because `kb extract` needs it for target selection, but downstream-facing outputs are now human-readable TSVs. The extract FASTQ outputs remain implementation detail rather than the primary exposed artifact.

Top-taxa reporting now operates on `counts.tsv` rather than a count tarball/H5AD. This keeps the reporting path aligned with the new public count output and gives WDL a stable TSV schema to cut scalar values into Terra table columns.

## Testing

Focused Docker tests against the current `main-classify` image with local source reinstalled:

```text
41 passed, 1 skipped
```

Kallisto integration tests:

```text
4 passed
```

Test coverage:

- **Kallisto wrapper** — count invocation, BAM-to-FASTQ handling, FASTQ input handling, empty-BAM behavior, sample ID propagation.
- **Counts output** — dense/sparse H5AD conversion to long-form `counts.tsv`, header-only empty output.
- **Extract output** — read hit TSV generation, summary TSV generation, FASTQ parsing validation, taxonomy annotation.
- **Top taxa** — focal-taxon filtering, no-taxonomy fallback, no-hit zero row, invalid count/schema validation.
- **Taxonomy maps** — headered maps, headerless maps, duplicate ID columns, ignored `.` placeholders, ignored trailing `+ssRNA`/`-ssRNA` values.
- **CLI plumbing** — parser dispatch for Kallisto top-taxa reporting and updated metagenomics command registration.

## Compatibility

- Breaking rename from `kb`/`kb_*` to `kallisto`/`kallisto_*`; no compatibility aliases are included.
- Kallisto/kb-python remains x86-only because of dependency availability.
- Existing non-Kallisto classifiers are unchanged.
- Public outputs move toward TSV-first, human-readable artifacts suitable for WDL/Terra consumption.

**Feature created utilizing GPT-5.5 and Opus 4.7**


src/viral_ngs/classify/kallisto.py
```diff
@@ -0,0 +1,769 @@
...
+        try:
+            if self._is_fastq(in_bam):
+                if not os.path.exists(in_bam):
+                    return
+
+                # Input is already FASTQ, use it directly (don't delete it later)
+                self.execute('kb count', out_dir, args=[in_bam], options=opts)
+            else:
+                if samtools.SamtoolsTool().isEmpty(in_bam):
+                    return
```
dpark01: Instead of `return`, can we invert the conditional and indent the rest of the code? Or if you must keep it this way (return if empty), maybe call `self._finalize_count_outputs(out_dir, sample_name)` before `return`? The problem is that this isn't producing proper (empty) output files when the input is empty.

src/viral_ngs/classify/kallisto.py
```diff
@@ -0,0 +1,769 @@
...
+    @staticmethod
+    def _parse_counts_tsv_count(raw_count, db_hit_id):
+        try:
+            count = float(raw_count)
+        except ValueError:
+            raise ValueError(f"Invalid count for {db_hit_id!r}: {raw_count!r}")
+        if count < 0:
+            raise ValueError(f"Negative count for {db_hit_id!r}: {raw_count!r}")
+        if not count.is_integer():
+            raise ValueError(f"Non-integer count for {db_hit_id!r}: {raw_count!r}")
```
dpark01, line 507: Why are floating point / non-integers rejected here? The upstream code (`_format_count`) that writes the file that's being read in here seems to be happy to write floats, so we can't glue these two steps together

src/viral_ngs/classify/kallisto.py
```diff
@@ -0,0 +1,769 @@
...
+        
+        # Find h5ad file in counts_unfiltered folder
+        h5ad_files = glob.glob(os.path.join(tmp_dir, "counts_unfiltered", "*.h5ad"))
+        log.debug(f"Found h5ad files in {count_tar}: {h5ad_files}")
+        if len(h5ad_files) == 0:
+            raise FileNotFoundError(f"No .h5ad file found in counts_unfiltered/ directory of {count_tar}")
+        if len(h5ad_files) > 1:
+            log.warning(f"Multiple .h5ad files found in {count_tar}, using first one: {h5ad_files[0]}")
+        
+        return h5ad_files[0]
```
dpark01: Is this now vestigal code?

src/viral_ngs/classify/kallisto.py
```diff
@@ -0,0 +1,769 @@
...
+    SUMMARY_TSV = 'summary.tsv'
+
+    def __init__(self, install_methods=None):
+        if not install_methods:
+            install_methods = []
+            install_methods.append(core.PrexistingUnixCommand(shutil.which('kb'), require_executability=False))
+        super(Kallisto, self).__init__(install_methods=install_methods)
+
+    def version(self):
+        return '1'
```
dpark01: does kallisto itself not have a `--version` or `-v` that we can call? Most other tools we ask the tool to tell us the version and return it here.

src/viral_ngs/classify/kallisto.py
dpark01: relatedly -- maybe we should have an integration test that tests gluing the two together?

tests/unit/classify/test_tools_kallisto.py
```diff
@@ -0,0 +1,456 @@
...
+
+        mock_process = mock_popen.return_value
+        mock_process.communicate.return_value = ('', '')
+        mock_process.returncode = 0
+
+        kallisto_tool.classify(kallisto_inputs['bam'], kallisto_inputs['index'], 'out_dir', kallisto_inputs['t2g'])
+
+        mock_popen.assert_not_called()
+        picard.execute.assert_not_called()
+        finalize_outputs.assert_not_called()
```
dpark01: So this seems to be testing behavior that's the opposite of what we want -- honestly we don't care if classify returns early or not or whether certain mocked things are called or not -- we should test that 1) it doesn't crash on empty input and 2) it produces a header-only tsv as output properly when run on empty input (which probably requires `finalize_outputs` to actually be called, but that's up to the implementing code)

src/viral_ngs/metagenomics.py
```diff
@@ -1390,220 +1387,94 @@ def parser_kb_extract(parser=argparse.ArgumentParser()):
...
+    parser.add_argument('--out_dir', dest='out_dir', help='Output directory (default: kallisto_out)', default='kallisto_out')
     parser.add_argument('--protein', action='store_true', help='True if sequence contains amino acids (default: False).')
     parser.add_argument('--targets', help='Comma-separated list of target sequences to extract from input sequences.', default=None)
-    parser.add_argument('--h5ad', help='Path to the output h5ad file. Can pull IDs to extract from this file.', default=None)
+    parser.add_argument('--h5ad', help='Path to a kallisto count h5ad file. Can pull IDs to extract from this file.', default=None)
     parser.add_argument('--threshold', type=int, help='Minimum read count threshold for a target to be extracted (only used when extractin IDs from h5ad; default: %(default)s)', default=1)
+    parser.add_argument('--sample-id', dest='sample_id', help='Sample identifier to write in summary.tsv. Defaults to the input basename.')
+    parser.add_argument('--id-to-tax-map', dest='id_to_tax_map', help='Optional ID to taxonomy mapping CSV/TSV file.')
```
dpark01, line 1398: The format of this custom input file needs far more documentation, the code that parses it is unclear and it sounds like it could be one of many possible structures and not clear what is supposed to make this file. Can we document what the format of the file is supposed to be and, if it really can be more than one thing, explain why (like it's sometimes written by this tool or that tool)?

dpark01 (changes requested):

src/viral_ngs/classify/kallisto.py
carze: Unfortunately no. The best you can do is a `-h`:

```
$ kb -h
usage: kb [-h] [--list] <CMD> ...

kb_python 0.29.1

positional arguments:
  <CMD>
    info      Display package and citation information
    compile   Compile `kallisto` and `bustools` binaries from source
    ref       Build a kallisto index and transcript-to-gene mapping
    count     Generate count matrices from a set of single-cell FASTQ files
    extract   Extract sequencing reads that were pseudoaligned to specific genes/transcripts (or extract all reads that were / were not pseudoaligned)

options:
  -h, --help  Show this help message and exit
  --list      Display list of supported single-cell technologies
  ```

File: src/viral_ngs/classify/kallisto.py (hunk 3 above)
Comment (carze) [reply src/viral_ngs/classify/kallisto.py]: Yes, will commit updated version with this function removed.

File: src/viral_ngs/classify/kallisto.py (hunk 4 above)
Comment (carze) [reply src/viral_ngs/classify/kallisto.py]: Actually `info` might do it but running it on my local install just throws an error.

Comment (carze) [reply src/viral_ngs/classify/kallisto.py]: I don't know if this is any better than the straight `-h` given the amount of info it spits out:

```
kb_python 0.30.2
kallisto: 0.52.0 (/data/projects/Evo2-synthetic/.pixi/envs/default/lib/python3.12/site-packages/kb_python/bins/linux/kallisto/kallisto)
bustools: 0.45.1 (/data/projects/Evo2-synthetic/.pixi/envs/default/lib/python3.12/site-packages/kb_python/bins/linux/bustools/bustools)
kb is a python package for rapidly pre-processing single-cell RNA-seq data. It
is a wrapper for the methods described in [2].

The goal of the wrapper is to simplify downloading and running of the kallisto
[1] and bustools [2] programs. It was inspired by Sten Linnarsson’s loompy
fromfq command (http://linnarssonlab.org/loompy/kallisto/index.html)

The kb program consists of three parts:

The `kb ref` command builds or downloads a species-specific index for
pseudoalignment of reads. This command must be run prior to `kb count`, and it
runs the `kallisto index` [1].

The `kb count` command runs the kallisto [1] and bustools [2] programs. It can
be used for pre-processing of data from a variety of single-cell RNA-seq
technologies, and for a number of different workflows (e.g. production of gene
count matrices, RNA velocity analyses, etc.). The output can be saved in a
variety of formats including mix and loom. Examples are provided below.

The `kb extract` command runs the kallisto [1] and bustools [2] programs. It
pulls out the raw reads from a fastq file that align to a specified gene or
transcript in the index.

Examples are available at: https://www.kallistobus.tools/tutorials

References
==========
[1] Bray, N. L., Pimentel, H., Melsted, P., & Pachter, L. (2016). Near-optimal
probabilistic RNA-seq quantification. Nature biotechnology, 34(5), 525.
[2] Melsted, P., Booeshaghi, A. S., Liu, L., Gao, F., Lu, L., Min, K. H., da
Veiga Beltrame, E., Hjorleifsson, K. E., Gehring, J., & Pachter, L. (2021).
Modular and efficient pre-processing of single-cell RNA-seq. Nature
Biotechnology.
```

I suppose we could just take the first line.

File: src/viral_ngs/metagenomics.py (hunk 6 above)
Comment (carze) [reply src/viral_ngs/metagenomics.py:1398]: I can add a section to our design document under the `classify` folder to capture some of this detail.

File: src/viral_ngs/classify/kallisto.py (hunk 4 above)
Comment (dpark01) [reply src/viral_ngs/classify/kallisto.py]: that's annoying.. how about subprocess.call'ing like a `conda list` or `conda freeze` or something for the version?

File: src/viral_ngs/classify/kallisto.py (hunk 2 above)
Comment (carze) [reply src/viral_ngs/classify/kallisto.py:507]: I think the coding agent was being somewhat pedantic as we would not expect to see a float value here (would indicate a partial hit?) as these are read counts and would be integers. Can leave it or change it honestly.

File: src/viral_ngs/classify/kallisto.py (hunk 1 above)
Comment (carze) [reply src/viral_ngs/classify/kallisto.py]: At this point if our SAM/BAM is empty we won't run kb/kallisto and have generated the output files needed by the `finalize_count_outputs` function. Instead we can add a helper function that creates dummy outputs.

File: KALLISTO_CODE_REVIEW.md [hunk 7]
```diff
@@ -0,0 +1,363 @@
+# Kallisto Implementation — Code Review Findings
```
dpark01: This file looks like a stray accidental commit -- delete?

src/viral_ngs/classify/kallisto.py
```diff
@@ -0,0 +1,738 @@
...
+        if 'sample' in adata.obs:
+            return [str(sample) for sample in adata.obs['sample'].tolist()]
+        return [str(obs_name) for obs_name in adata.obs_names.tolist()]
+
+    @staticmethod
+    def _format_count(count):
+        as_float = float(count)
+        if as_float.is_integer():
+            return str(int(as_float))
+        return str(as_float)
```
dpark01:
```suggestion
    @staticmethod
    def _format_count(count):
        return str(int(round(count)))
```

As discussed elsewhere, fractional counts are not semantically meaningful, so let's either clean up / enforce an integer output, or alternatively throw exceptions on fractional output.

src/viral_ngs/classify/kallisto.py
dpark01: Ok then switching this discussion to the writer method.

Also, should we add an integration test that validates that the two steps can be run one after the other on the output of the previous?

carze: Test added:`test_write_counts_tsv_from_h5ad_rejects_fractional_counts`

src/viral_ngs/classify/kallisto.py
carze: We are now raising an error as we want to capture if we are getting fractional values rather than automatically casting them and truncating the decimal off.

dpark01 (approved):

PR #1110, opened 2026-08-04
Add .trivyignore entry for CVE-2025-47273 (phantom setuptools from pip's vendored SBOM)

dpark01:
## Why

`scan-containers` has been failing on `setuptools 70.3.0 → 78.1.1` (HIGH, CVE-2025-47273) across every image tier, blocking the required `scan-containers (mega)` check. **Nothing is installed at that version.**

The source is `pip/_vendor/bom.cdx.json` — a CycloneDX SBOM that pip 26.2 ships describing its vendored dependencies, which declares `setuptools 70.3.0`. Trivy's **SBOM analyzer** reads that manifest and synthesizes a package entry from it:

```json
{
  "ID": "setuptools@70.3.0",
  "Version": "70.3.0",
  "AnalyzedBy": "sbom"
}
```

No `FilePath`. That is why the vulnerability had no `PkgPath` and why no amount of searching the filesystem for a `setuptools-70.3.0.dist-info` ever found anything.

It is a **false positive**. The flaw is in `setuptools/package_index.py` (`PackageIndex`, which fetches and extracts sdists). pip vendors only `pkg_resources`, a subset of the distribution:

```
$ ls /opt/conda/lib/python3.12/site-packages/pip/_vendor/
cachecontrol  certifi  distlib  distro  idna  msgpack  packaging
pkg_resources  platformdirs  pygments  pyproject_hooks  requests
resolvelib  rich  tomli  tomli_w  truststore  urllib3

$ find .../pip/_vendor -name "package_index*"
(no results — vulnerable module absent)

$ python -c "import setuptools; print(setuptools.__version__)"
82.0.1        # well above the 78.1.1 fix
```

### Why it appeared now

pip 26.0.1 — the version in the cached 2026-07-23 `baseimage` layer — ships no SBOM. `pip install --upgrade pip` now pulls 26.2, which does. So the finding appeared the moment those layers were rebuilt, on an otherwise unchanged commit.

## Fix

A `.trivyignore` entry, in the same shape as the existing jackson-databind exceptions (vulnerable code path is architecturally absent, with the reasoning written out).

**Why not delete pip's SBOM.** That would silence it, and I had it implemented that way first. But the SBOM is accurate supply-chain metadata that also covers pip's *other* vendored packages — `urllib3`, `requests`, `certifi`, `idna` — which are real code present in the image. Deleting it to suppress one mis-mapped CVE would blind the scanner to genuine future findings in that tree. One documented exception is the better trade.

Worth knowing: pip's vendored SBOM already produces a second phantom, `GHSA-6v7p-g79w-8964` against `msgpack 1.1.2`, which the Rego policy happens to filter. So this class of finding will recur; the treadmill is real but slow, and per-CVE entries keep the reasoning auditable.

## Verification

Against the `baseimage` built by this PR, using CI's exact gating filter (HIGH/CRITICAL, fixable only):

```
without ignorefile:              CVE-2025-47273 setuptools 70.3.0
                                 GHSA-6v7p-g79w-8964 msgpack 1.1.2
with .trivyignore + rego policy: no findings
```

## Two corrections to earlier versions of this PR

I got this wrong twice before landing on the SBOM. Recording both so nobody re-runs them:

1. **Unpinned setuptools in the isolated FreeBayes env.** Wrong — that env solves to setuptools **83.0.0**, confirmed by a dry-run solve *and* by extracting the built layer.
2. **`pip/_vendor/vendor.txt`.** It does contain `setuptools==70.3.0`, so it looked conclusive. Also wrong: I removed it, and the resulting build shows the file gone from the layer while Trivy still reports the CVE against that same layer. A plausible-looking artifact is not the same as the one the scanner reads.

The Dockerfile and `install-conda-deps.sh` changes from those attempts are reverted; the net diff is `.trivyignore` only.

## Not included

`CVE-2026-59890` (MEDIUM) affects the *real* setuptools 82.0.1, fixed in 83.0.0. It is below the gating threshold so it isn't blocking, but bumping `setuptools>=80.10.1` → `>=83.0.0` in `docker/requirements/baseimage.txt` would clear it. Left out to keep this PR to one concern — happy to add it here or file it separately.

Related: #1109 (the auto-fix bot's `cryptography` bump) clears the other finding that was failing this gate.


dpark01:
## Scope widened — the first commit was incomplete

The scan on #1109 rebuilt every tier (bumping a requirements file invalidates the conda layer), and that exposed more of this than the assemble-only view showed: **CVE-2025-47273 now appears in `baseimage`, `core`, `classify` and `phylo` as well.**

Same root cause, second location. The main conda env has its own copy of the manifest:

```
$ pip --version
pip 26.0.1 from /opt/conda/lib/python3.12/site-packages/pip (python 3.12)
$ grep -n setuptools /opt/conda/lib/python3.12/site-packages/pip/_vendor/vendor.txt
15:setuptools==70.3.0
$ python -c "import setuptools; print(setuptools.__version__)"
79.0.1
```

It went unflagged until now only because the published `baseimage` layer was being served from registry cache and never rescanned after a real rebuild. So my note in the original description — that the base env's pip upgrade records a different setuptools — was wrong. Every env with pip carries this manifest.

pip can't just be deleted from the main env: `Dockerfile.baseimage` installs firecloud with it and every tier runs `pip install .`. So remove only the manifest, which pip does not need at runtime.

Three placements, each verified in a container:

1. **`install-conda-deps.sh`** — one change covers every tier. Written as a `find` sweep so it also catches a repopulated `pkgs` cache.
2. **`Dockerfile.baseimage`** — required because `pip install --upgrade pip` runs *after* that script and restores the file. Confirmed directly:
   ```
   rm -f .../pip/_vendor/vendor.txt   ->  gone
   pip install --upgrade pip wheel    ->  .../pip/_vendor/vendor.txt   (restored)
   ```
3. **`Dockerfile.mega`** — it creates its own `/opt/freebayes`, so it needs the same pip removal as assemble. Missed in the first commit.

pip remains fully functional without the file:

```
pip 26.2 from /opt/conda/lib/python3.12/site-packages/pip (python 3.12)
pip install six            -> works
import pkg_resources, setuptools -> setuptools 79.0.1
```

## Status of the gate

`scan-containers` on #1109 confirms the other half is fixed — **CVE-2026-69247 (`cryptography`) is gone** from both baseimage and mega there. `setuptools` is now the only finding failing that check, on either PR.

dpark01:
## Migrated to Trivy's YAML ignore format — every entry is now scoped

Following up on the review point: a plain-text `.trivyignore` line matches a CVE ID **globally**, suppressing it anywhere in the image including copies nobody assessed. All three existing jackson-databind entries already documented that as an accepted compromise in prose:

> *"this entry will NOT suppress those findings — Trivy matches `.trivyignore` by CVE ID globally, but we accept this because no other JAR in our images bundles this old version."*

The YAML format removes that compromise rather than describing it.

| Entry | Scope | Why that key |
|---|---|---|
| CVE-2020-25649, CVE-2026-54512, CVE-2026-54513 (jackson-databind) | `paths: opt/conda/share/snpeff-*/snpEff.jar` | These findings carry a real `PkgPath`. The same CVE in picard / gatk / fgbio / fastqc jars is now **reported**, not suppressed. |
| CVE-2025-47273 (setuptools) | `purls: pkg:pypi/setuptools@70.3.0` | `paths` cannot work — the finding is synthesized from pip's embedded CycloneDX SBOM (`"AnalyzedBy": "sbom"`) and has **no** `PkgPath` at all. |

The glob on the jar path is deliberate: that file has moved between `snpeff-5.1-0`, `5.2-3` and `5.4.0c-0`, so a literal path would silently stop matching on the next version bump — and a silently-dead ignore is worse than no ignore.

### Verified, not taken from the docs

Built two throwaway images to test each mechanism in isolation:

```
paths, matching glob        -> SUPPRESSED
paths, wrong path           -> still reported
paths, on the SBOM finding  -> still reported   (no path to match — this is why purls)
purls @70.3.0               -> SUPPRESSED
purls @1.2.3                -> still reported   (version scoping discriminates)
```

The one that matters most, using this exact file against an image with a **genuinely vulnerable** `setuptools==70.0.0` installed:

```
CVE-2025-47273  70.0.0  PkgPath: .../setuptools-70.0.0.dist-info/METADATA   <- still reported
```

The old bare-CVE-ID entry would have hidden that. This is exactly the resurface-by-other-means case.

### Also updated

- The four `trivyignores:` inputs across `docker.yml` and `container-scan.yml`.
- The auto-fix **path-allowlist regex** — it whitelisted `.trivyignore`, so without this the bot would be blocked from editing the renamed file.
- Both Claude prompts: new entries must now be scoped, with guidance on which key applies when (`paths` for a finding with a `PkgPath`, `purls` for SBOM-derived ones).
- `SKILL.md`, including the `AnalyzedBy: sbom` tell — that field is what identifies a phantom, and is what I should have looked at first.

### Deliberately not included

`expired_at`, which makes Trivy stop honouring an entry after a given date. It would enforce the quarterly review cadence the file header asks for, but it turns a required security gate red on a timer, possibly unattended. Happy to add it if you want that trade — it seemed like a call to make explicitly rather than slip in.

### Gate status

The setuptools half is confirmed fixed: the latest scan on this PR shows it gone from every tier. The only finding still failing `scan-containers` here is **CVE-2026-69247 `cryptography`**, which #1109 fixes and which is not in this branch — `baseimage` builds from a cached layer here because this PR touches no Dockerfile input.

So the two PRs each block the other: #1109 is red only on setuptools (fixed here), #1110 is red only on cryptography (fixed there).

dpark01:
## Merging with an admin bypass of `scan-containers (mega)`

Recording the rationale, as with #1108.

Required-check status at merge:

```
create-manifest-assemble       pass
create-manifest-baseimage      pass
create-manifest-classify       pass
create-manifest-core           pass
create-manifest-mega           pass
create-manifest-phylo          pass
all-tests-pass                 pass
scan-containers (mega)         fail   <-- bypassed
scan-secrets (amd64)           pass
scan-secrets (arm64)           pass
```

`scan-containers (mega)` is red on exactly one finding, and it is not the one this PR addresses:

```
baseimage: CVE-2026-69247 cryptography 49.0.0
mega:      CVE-2026-69247 cryptography 49.0.0
```

That is #1107, fixed by #1109, which is not in this branch. This PR touches no Dockerfile input, so `baseimage` builds from a cached layer and keeps 49.0.0.

**What this PR was supposed to fix is confirmed fixed.** `CVE-2025-47273` no longer appears in any tier — the PURL-scoped entry works in CI, not just locally.

Also confirmed from the `scan-containers (baseimage)` log, since the YAML format was the one thing that passed locally but ran through a different code path in CI (I tested with the raw `trivy` binary; CI goes through the action wrapper):

```
INPUT_TRIVYIGNORES: .trivyignore.yaml
Using YAML ignorefile '.trivyignore.yaml':
```

The pinned action SHA branches on `*.yml|*.yaml`, sets `TRIVY_IGNOREFILE` directly, and rejects mixing formats or passing more than one YAML file. All four call sites pass exactly one, and the old `.trivyignore` is deleted, so neither error case applies.

### Why the bypass is safe

This diff changes `.trivyignore.yaml`, two workflow files and `SKILL.md`. No `docker/requirements/*.txt`, no Dockerfile, no runtime code — it cannot change the contents of any image. The failing finding is equally present on `main` right now and is not made worse by merging.

### Why not just fix cryptography here

It would be circular. #1109 is the auto-fix bot's own PR for that CVE, and it is red on *only* the setuptools finding that this PR fixes. Each PR blocks the other. Merging this one first means #1109 rebuilds against a `main` that has the ignore entry and should go green on its own — so the bot's PR passes required checks with **no bypass**, which is the more meaningful validation of that pipeline.

### Next

1. Update #1109 against the new `main` and let its CI re-run.
2. If green, merge it normally. That closes #1107 and returns `scan-containers (mega)` to enforcing without exception.

